PR #1, opened 2016-09-13
Updates the the line comment to hash instead of double-slash

Fitzsimmons:
This tells vscode that the single line comment is prefixed with a `#` instead of a `//`. The double-slash comment does appear to work, but:

a) it does not show up in the [official documentation](https://www.terraform.io/docs/configuration/syntax.html).
b) it is not syntax highlighted, but `#` comments are.

This is useful for automated block commenting and uncommenting in vscode with `ctrl+/`, which will use the wrong kind of comments until this is merged.

### DIFF
--- a/terraform.configuration.json
+++ b/terraform.configuration.json
@@ -1,7 +1,7 @@
 {
 	"comments": {
 		// symbol used for single line comment. Remove this entry if your language does not support line comments
-		"lineComment": "//",
+		"lineComment": "#",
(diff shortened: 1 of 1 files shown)


boyvinall:
I just made the same fix in my local copy.  Be great to have this on the vscode marketplace.

josh-padnick:
Ha, I made the same change, too! Would be nice to get this merged.

mauve:
thanks I will merge now and make a new release

boyvinall:
Awesome, thanks @mauve

PR #2, opened 2016-09-19
Recognize "data" resources.

josh-padnick:
Previously, the terraform plugin for vscode only recognized resources starting with the keyword `resource`. But in Terraform 0.7.x, a new resource type `data` was introduced. This is a naive implementation to support that.

Note that while I'm using this locally, I have not tested this thoroughly, so this PR is more of an invitation to complete this work. There may be unique syntax for the `data` resources that this admittedly simple change doesn't account for.

### DIFF
--- a/syntaxes/terraform.tmLanguage
+++ b/syntaxes/terraform.tmLanguage
@@ -98,7 +98,7 @@
 				</dict>
 			</dict>
 			<key>match</key>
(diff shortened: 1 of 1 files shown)


boyvinall:
I just made exactly the same change to mine!  Thanks for doing the PR @josh-padnick.

@mauve  - would be great to get this merged please, thanks.

mauve:
thanks, I merged and published 0.0.3 to the marketplace

Issue #23, opened 2017-05-19
Highlight newish "terraform" keyword


alexsomesan:
A `terraform` HCL block was introduced recently, which so far is only used to pin the templates to a specific terraform version.

The `terraform` word doesn't get highlighted as a keyword, like `resource` or `variable` would.

Documented here: https://www.terraform.io/docs/configuration/terraform.html#example

PR #24, opened 2017-05-19
Highlight terraform keyword

alexsomesan:
Is this the correct approach to fix https://github.com/mauve/vscode-terraform/issues/23 ?

### DIFF
--- a/syntaxes/terraform.json
+++ b/syntaxes/terraform.json
@@ -68,6 +68,15 @@
     },
     {
       "captures": {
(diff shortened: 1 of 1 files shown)


mauve:
No idea if this is the correct way, can you also add a short terraform file in `/templates` which highlights the new syntax?

haad:
This PR can be closed I have added same code in #30 + examples. Thanks @alexsomesan @mauve

PR #32, opened 2017-09-18
Add support for local variable type

mspaulding06:
Terraform 0.10.4 introduced local variables with [documentation here](https://www.terraform.io/docs/configuration/locals.html).

This adds syntax support for keywords `local` and `locals`.


mauve:
thanks for you contribution, I will do a new release later today or as soon as possible.

Issue #34, opened 2017-10-05
Validate returns "Validation failed, more information in the output tab" but output tab is empty


zxaos:
Using terraform 0.10.4 and vscode-terraform 0.0.18 with vscode user settings:

```
    "terraform.lintPath": "/usr/local/bin/tflint",
    "terraform.path": "/usr/local/Cellar/tfenv/0.5.2/versions/0.10.4",
    "terraform.formatOnSave": true
```

mauve:
`terraform.path` should be the path to the executable. I will have a look wrt the empty output tab.

zxaos:
Oops! Fixed. But even adjusting `terraform.path` to the correct path, the behaviour is the same.

mauve:
You might also need to set `terraform.templateDirectory`:

```js
  // A relative path to where your templates are stored relative to the workspace root.
  "terraform.templateDirectory": "templates",
```

`terraform validate` only works when given a directory containing `.tf` files. It does not work if given directory of directory for example. This is a restriction I will work around in the plugin (I will open a separate issue).

However when validation fails for me I actually get output in the output tab (did you select *Terraform*?):

![screen shot 2017-10-06 at 9 27 07 am](https://user-images.githubusercontent.com/795800/31268188-75b41a84-aa7b-11e7-80c6-3b723a43ef25.png)

zxaos:
Got it now - adding the `templateDirectory` made it start working! Thanks for the help.

PR #37, opened 2017-10-11
Show output channel on validation fail

jackric:
Make the output easy to find for users. I had a frustrating time finding the validation output (like in #34). This change highlights the output channel containing the validation warnings.

The previous behavior requires the user to find the 'Terraform' entry in the output pane's dropdown. This wasn't obvious to me.


mauve:
thanks for your contribution I will make a new release soonishly.

PR #55, opened 2018-03-11
Fix typo in README.md

chroju:
Hello. I have fixed typo in README.md because the image about auto-completion is not displayed.

This tool is very helpful for my Terraform work. thx! :)


mauve:
thanks for the fix! will do a release in a few minutes

PR #66, opened 2018-04-10
[Discussion] [Live Share] Restricting language services to local files

lostintangent:
In preparation for [Visual Studio Live Share](https://aka.ms/vsls) adding support for "guests" to receive remote language services for Terraform, this PR simply updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files. This way, when someone has the Terraform extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with).

*Note: As an example, the TypeScript/JavaScript language services that come in-box with VS Code [already have this scheme restriction](https://github.com/Microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/fileSchemes.ts#L12), and so this PR replicates that behavior.*

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,6 +9,11 @@ import { CompletionProvider } from './providers';
 export let errorDiagnosticCollection = vscode.languages.createDiagnosticCollection("terraform-error");
 export let outputChannel = vscode.window.createOutputChannel("Terraform");
 
+const documentSelector: vscode.DocumentSelector = [
+    { language: "terraform", scheme: "file" },
+    { language: "terraform", scheme: "untitled" }
+];
+
 export function activate(ctx: vscode.ExtensionContext) {
     ctx.subscriptions.push(errorDiagnosticCollection);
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -9,22 +9,29 @@ import { CompletionProvider } from './providers';
...
+    { language: "terraform", scheme: "untitled" }
+];
+
 export function activate(ctx: vscode.ExtensionContext) {
     ctx.subscriptions.push(errorDiagnosticCollection);
 
     let formattingProvider = new FormattingEditProvider;
 
     // we need to manually handle save events otherwise format on autosave does not work
     ctx.subscriptions.push(vscode.workspace.onDidSaveTextDocument((doc) => formattingProvider.onSave(doc)));
```
mauve: do we need some magic in the `onDidSaveTextDocument()` handler aswell?

src/extension.ts
```diff
@@ -9,22 +9,29 @@ import { CompletionProvider } from './providers';
...
 
     ctx.subscriptions.push(vscode.commands.registerCommand('terraform.validate', () => { validateCommand(); }));
     ctx.subscriptions.push(vscode.commands.registerCommand('terraform.lint', () => { lintCommand(); }));
 
-    ctx.subscriptions.push(vscode.languages.registerCompletionItemProvider("terraform", new CompletionProvider, '.', '"'));
+    ctx.subscriptions.push(vscode.languages.registerCompletionItemProvider(documentSelector, new CompletionProvider, '.', '"'));
 
     // index operations
-    initializeIndex(ctx);
+    if (vscode.workspace.rootPath) {
```
mauve: just to double-check: this means that for the live-share guest `initializeIndex()` is not called? is there a way to get those features to work for a guest?

lostintangent: The Live Share guest would automatically get all of the features that `initializeIndex()` is registering (e.g. go to definition, rename, document symbols). We could have updated the calls to `vscode.languages.register*` to take the same `DocumentSelector` I'm using here, but `initializeIndex` also does other things that would be driven by the host (e.g. listening for config changes, scanning for Terraform documents), and so it seemed simpler to just prevent calling it entirely for guests.

src/extension.ts
lostintangent: @mauve Ah good point! I'll move this subscription into the same conditional statement as the call to `initializeIndex`.

mauve:
thanks for your contribution, I have merged the PR; will try to do a release later this week.

PR #78, opened 2018-05-05
Fix autocomplete of variables and outputs

ratza:
Looks like after some of the latest changes the autocompletion seems to be completely broken for me. I've seen autocomplete working somewhat well on a machine I have no access to at the moment, but I'm not sure what version it was running (not latest for sure).

I tried to test v1.0.5 (VSCode v1.23.0) that currently is the latest stable and it was still failing to me, so I dug in to see what was going on.

This MR addresses lookup of variables and outputs in the Index.

I still seem to experience issues with the rest of the autocomplete (resources, etc)


mauve:
Thanks. Looks good. I will see if I can add some automated tests for the completion.

PR #100, opened 2018-06-16
Return uniq interpolationCompletions

### LINKED ISSUE #98: Auto-completion duplicates
On a similar setup to the Issue #96 I get duplicates on the Auto-completion
image

heldersepu:
Testing to see if the PR fires the UnitTest
If all OK this fixes #98

### DIFF
--- a/src/autocompletion/completion-provider.ts
+++ b/src/autocompletion/completion-provider.ts
@@ -118,7 +118,7 @@ export class CompletionProvider implements vscode.CompletionItemProvider {
       if (token) {
         let interpolationCompletions = this.interpolationCompletions(document, position);
         if (interpolationCompletions.length !== 0)
-          return interpolationCompletions;
+          return _.uniq(interpolationCompletions);
       }
     }
 


src/autocompletion/completion-provider.ts
```diff
@@ -118,7 +118,7 @@ export class CompletionProvider implements vscode.CompletionItemProvider {
       if (token) {
         let interpolationCompletions = this.interpolationCompletions(document, position);
         if (interpolationCompletions.length !== 0)
-          return interpolationCompletions;
```
mauve: unsure if this does what you want. If we really want this to be a unique list, rather than just ensure that the results have more information (e.g. show some sort of information where the resource is defined in case of duplicates), I think the making it unique part should be part of `interpolationCompletions` (e.g here: https://github.com/mauve/vscode-terraform/pull/100/commits/764bd8d4117ea943f3497e12816d7832786149a0#diff-c8cbd71ada55b23b86d5b1e2aac1255eL101)

Uniqueness of `Segment`s (which is the things returned by `(File)Index::query()` can be ensured using `Segment::id()`. 

In any case I would prefer if we could also add a test for this in https://github.com/mauve/vscode-terraform/blob/master/test/autocompletion/completion-provider.test.ts or in any of the test files in https://github.com/mauve/vscode-terraform/tree/master/test/index depending on where the uniqueness is guaranteed.

mauve:
First of all, thanks for your contribution :) it always glads me when ppl make pull requests. I am currently a bit unsure how to tackle this problem (I need to think about the *uniqueness-per-folder* thing).

In the meantime I have updated the instruction in `BUILDING.md` so that you should be able to build and run everything locally.

heldersepu:
Since #98 is now closed this makes no sense to keep it open...
But the idea still valid = all items in the completions should be unique

mauve:
feel free reopen this PR but please we need tests and uniqueness should be established using `Segment::id()`.

heldersepu:
yes If I can get my environment up and running I surely will

PR #120, opened 2018-08-10
Fixed broken multiline strings

### LINKED ISSUE #109: Multi-line string interpolation breaks syntax highlighting
Copying the example from the [terraform EKS getting started page](https://www.terraform.io/docs/providers/aws/guides/eks-getting-started.html) breaks syntax highlighting.

screen shot 2018-07-22 at 3 36 16 pm

squidfunk:
This PR fixes #109 - encountered it today while switching from Atom to VS Code and thought I give it a try. I'm not an expert in the syntax so let's improve it together.

### DIFF
--- a/syntaxes/terraform.json
+++ b/syntaxes/terraform.json
@@ -166,7 +166,7 @@
               "name": "entity.tag.embedded.start.terraform"
             }
           },
-          "end": "\\}",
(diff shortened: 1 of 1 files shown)


mauve:
thanks for your contribution :) thanks for using my plugin

PR #126, opened 2018-09-06
Allow brackets auto-closing in double quote strings

aliusmiles:
@mauve This should enable brackets autoclosing  to work with a new `Auto Closing Brackets` settings set to `languageDefined`

Ref: https://github.com/Microsoft/vscode/pull/52634


mauve:
thanks for your contribution, merged!

PR #200, opened 2019-07-17
Use LSP Terraform language server

lawrencegripper:
Current status: Beyond a hack. Language server working. 

# What's in scope?

This PR adds code to download and wire up the language server. I'm not building the language server and bugs with it sit outside of the scope of this PR. They should be raised here https://github.com/juliosueiras/terraform-lsp 

> Note: Be mindful that the language server is being worked on for free - do report bugs but you have no right to demand anything is fixed, if you want it done.. cut the code and submit a PR or start a bug bounty to support the awesome work of @juliosueiras

# How it works

The new code is mainly focused around getting the LSP and storing it somewhere. Currently when you start up this branch you'll be prompted to install the LSP and then given a picker to select a version. 

Once installed the language server will kick in and you'll get TF 0.12 goodness. 

> Warning: The language server is still experimental by the looks of things so is likely to be a bit unstable. I've only tested on a smallish `terraform` folder. 

# Todo

- [x] Fix up config to allow enable/disable LSP (blocked see `What needs doing that I can't do`)
- [x] Add feature to use existing language server from existing path on disk
- [x] Add command to update the LSP to another release
- [x] Re-enable existing Index based commands by registering dummy handler when LSP enabled.
- [x] Add command to toggle between `indexer` and `languageServer` 
- [x] Check for updates to `languageServer` on startup and prompt to install
- [x] Notification on startup if running TF 0.12 that `languageServer` can be enabled
- [x] Add integration tests (check `languageServer` install and basic operation with known good version)
-  ~~Merged https://github.com/juliosueiras/terraform-lsp/pull/11 shutdown fix~~ De-scoped (can ship out of band with LSP update)
-  ~~Fix up subfolder provider discovery issue https://github.com/juliosueiras/terraform-lsp/issues/12~~ De-scoped (can ship out of band with LSP update)
- ~~Investigate if grammer needs updating for HCL2~~ De-scoped

# What needs doing that I can't do?

1. I'm on Linux, would be great if someone could pull down and try this branch on Windows and Mac. 
2. @juliosueiras could you cut a new release from the `master` of the LSP. Seems to be much more stable than current release `v0.0.3`
3. Polish and pointers on the typescript... been an age since I've worked with it and I've never written a VSCode extension before. 
4. @mauve Can you give me some pointers on how you'd like this to work? Current thinking is to have a setting like `ExperimentalLanguageServerEnabled` which would bring start this code. What's the best way to toggle between this and the existing indexer? Ideally I'd have a way to enforce the use of either one or the other but can't wrap my head about the `vscode` extensions settings stuff. Any pointers would be ace. If this works nicely are you happy to release a version with the LSP as a toggable option off by default?


lawrencegripper:
Grr shouldn't have committed so late I've included the binary and appear to have mashed up the readme too. Going to bed will fix as and when I have time.

lawrencegripper:
@mauve the build looks like it fails due to a missing environment var, guessing it not available as it's a secret and this PR comes from outside the repo.

chucknelson:
@lawrencegripper - I'll try to test this out on Windows later today.

lawrencegripper:
Thanks. To temper excitement I haven't seen great stability from this atm but with the feature to download and use newer LSP servers this could be shipped and newer versions of the LSP resovle things. May also be something I've done wrong in wiring things up. Need more time to look at it.

mauve:
@lawrencegripper thanks for taking a look at this, looking forward to see progress. I will be able to review during the weekend.

As for approach I think it might make sense to be very aggressive that means kill of features in order to just get the ball rolling.

We can do a preview release with metrics enabled to get crash-reports.

With regard to the build failure. In the gulp files there is an if somewhere which detects if the build is running in CI and then requires the key. You can remove that entire logic, the key doesn’t exist anymore.

I will try to use the weekend to also migrate to azure DevOps for building that should simplify preview releases (automatic pushing)

Again: I am very grateful that you are working on this.

lawrencegripper:
Thanks @mauve no rush on looking at stuff this weekend I'd like a bit more time for tweaking and testing my end. If you get a moment for cursory read of the code and thoughts I wouldn't say no but don't go super detailed. 

Know the pain of having a pet project grow and the pressure that can bring, really happy to help out I've been using this for an age so about time I gave something back... just hope I can get it working nicely.

chucknelson:
@lawrencegripper - Quick attempts on Windows at work have hit some issues, but want to try at home later tonight.

Right now the issues are:
* Even after `npm install`ing and double-checking dependency versions, I'm getting some typescript-related compile errors like:
  ```
  .../node_modules/@octokit/rest/index.d.ts(5,1): error TS1084: Invalid 'reference' directive syntax.
  .../node_modules/vscode-languageclient/lib/client.d.ts(1,399): error TS2305: Module ''vscode'' has no 
  exported member 'DefinitionLink'.
  ...
  ```
* The LSP download fails - might be related to the compile errors above, but it's a strange "connection refused" type error for now:
  ```
  Activating extension `mauve.terraform` failed:  connect ECONNREFUSED 127.0.0.1:80
  console.pub.js:42
  Here is the error stack:  Error: connect ECONNREFUSED 127.0.0.1:80
  	at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1113:14)
  ```

lawrencegripper:
Amazing, thanks @chucknelson I'll get these fixed rolled into the PR when I have some time. 

The download issue is odd. Looks like it's trying to download from `localhost` where I'd expect a url like `https://github.com/juliosueiras/terraform-lsp/releases/download/v0.0.3/terraform-lsp_0.0.3_windows_amd64.tar.gz`. Are you running behind a proxy or have any other network bits that might be getting in the way? Most likely I've messed something up my end. I'll add some more logging in next push.

lawrencegripper:
Thanks @chucknelson pushed latest build with fixes for the TS errors, your help was spot on!

In the latest push I've added the following config so you can toggle between `indexing` and the `languageServer` as the more I play the more I think they'll need to co-exist for a bit. 

```json
"terraform.indexing": {
        "enabled": false,
        "liveIndexing": true,
        "delay": 500,
        "exclude": [
            ".terraform/**/*",
            "**/.terraform/**/*"
        ]
    },
    "terraform.languageServer": {
        "enabled": true,
        "args": []
    }
```

juliosueiras:
Thanks for creating the PR =) , also I added the args option to the binary for disable and changing location for logs

Will mostly focus on covering as much ground for all types(backends,provisionsers,functions,etc)

juliosueiras:
Also yeah , will add a auto releasing to the CI/CD

smiller171:
Once there's a preview release ready I'm happy to help test on MacOS if needed.

lawrencegripper:
@smiller171 I've build a Visx which should mean you can have a go at testing now. It's still early days but any feedback you can give would be great. 

[terraform-1.3.12.zip](https://github.com/mauve/vscode-terraform/files/3412148/terraform-1.3.12.zip)

mauve:
@lawrencegripper I have switched master over to Azure DevOps which builds and runs test on linux and windows (mac doesnt work yet because of the docker dependency), that way you should be able to test windows by just adding automated tests

mauve:
@lawrencegripper hmm weird `npm run integration-test` works locally but fails on Azure DevOps (linux), I need to go AFK for a few minutes so if you have any clue how this stuff works I just wanted to let you know that it actually works locally :)

smiller171:
> @smiller171 I've build a Visx which should mean you can have a go at testing now. It's still early days but any feedback you can give would be great.
> 
> [terraform-1.3.12.zip](https://github.com/mauve/vscode-terraform/files/3412148/terraform-1.3.12.zip)

I got:

```
Unable to install extension 'mauve.terraform' as it is not compatible with VS Code '1.36.1'.
````

**Comment by mauve** [comment]
@lawrencegripper this might be of interest, and might make it possible to simplify the downloading stuff (i.e. do not implement it just wait for this to be implemented in vscode):

https://github.com/microsoft/vscode/wiki/Roadmap#extension-authors
  Enable extensions to install additional platform specific components at extension installation time.

File: src/languageclient.ts [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+
+        let downloadUrl = "";
+        release.data.assets.forEach(asset => {
+            if (!asset.name.endsWith('.tar.gz')) {
+                return;
+            }
+
+            if (asset.name.includes(platform)) {
+                downloadUrl = asset.browser_download_url;
+            }
```
chucknelson: This currently doesn't work for Windows - the node process platform is `win32`, but the LSP release artifact only has `windows` in the name.

package.json
```diff
@@ -257,16 +254,21 @@
     "source-map-support": "^0.5.6",
     "ts-node": "^7.0.0",
     "tslint": "^5.10.0",
-    "typescript": "^2.9.1",
+    "typescript": "^3.0.0",
+    "vscode": "^1.1.32",
```
chucknelson: `vscode` needs to be removed since we moved to `vscode-test` + `@types/vscode` - causes build errors otherwise.

See: https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode

mauve: good catch I think rebasing on master should fix this (as it is correct on master)

package.json
```diff
@@ -257,16 +254,21 @@
...
     "applicationinsights": "^1.0.5",
     "graphlib-dot": "^0.6.2",
     "lodash": "^4.17.11",
     "minimatch": "^3.0.4",
+    "tar": "^4.4.10",
+    "typed-rest-client": "^1.5.0",
     "uuid": "^3.3.2",
     "viz.js": "^1.8.2",
+    "vscode-languageclient": "^5.2.1",
     "vscode-uri": "^1.0.5"
```
chucknelson: For now we may want to have this capped at `1.0.6` - anything beyond that, including the `2.x` releases, causes tests to break. Or maybe there is a simple test tweak we can make...

```suggestion
    "vscode-uri": "^1.0.5 <=1.0.6"
```

chucknelson:
@smiller171 
> I got:
> 
> ```
> Unable to install extension 'mauve.terraform' as it is not compatible with VS Code '1.36.1'.
> ```

@lawrencegripper, @mauve - Just FYI that we need to update the [`engines` version in `package.json`](https://github.com/mauve/vscode-terraform/blob/b62eb9db30b313e2c6186067a57fa5452b258471/package.json#L10). 

I think it was accidentally set to `1.36.0` only in the latest changes on `master`, but should probably be changed to `^1.36.0` to specify a minimum, not an absolute version reference.

package.json
mauve: master now uses `"vscode-uri": "^1.0.8"` and the tests pass but I haven't tested the plugin more than just running units and integs, I had to fix some tests to make it pass though

package.json
lawrencegripper: Thanks, I've merged in master now.

src/languageclient.ts
lawrencegripper: Ah man feel silly, should have spotted that one. I've added a fix and a runtime check to report and error if this occurs for a platform or we introduce a regression.

package.json
lawrencegripper: Cool, I've merged master into this work so hopefully all set now.

lawrencegripper:
> https://github.com/microsoft/vscode/wiki/Roadmap#extension-authors
Enable extensions to install additional platform specific components at extension installation time.

@mauve this looks interesting, I think for the time being I'll keep the custom approach as I can see a use case for updating the LSP out-of-band of the extension. For example, a new version ships users can do `CTRL-SHIFT-P` -> `Update Terraform Language Server` and pull it down ... if things don't work they can switch back etc. Sound reasonable? I've added this command in the latest commit. 

For others to play here is a new build based off these changes:
[terraform-1.3.12.zip](https://github.com/mauve/vscode-terraform/files/3415013/terraform-1.3.12.zip)

lawrencegripper:
I've added back in the existing commands using the `indexer` and now when using the LSP these return a warning that they'll not function. This gets the build back to green and put me in a good place to start adding some tests around the LSP!

src/languageclient.ts
lawrencegripper: Hopefully not all fixed :)

lawrencegripper:
@smiller171 and @chucknelson both of your issues should now be resolve in latest visx ... let me know if they aren't.

lawrencegripper:
@mauve I've updated the build to roll the minor release number forward each time and publish an artifact (zip file) out of the build. This means people can pull down the outputted `visx` from each successful build for testing and it'll have a unique version. Is that alright? Hopefully makes it easier when reviewing PR's too as can pull down and run without full build etc.

azure-pipelines.yml
```diff
@@ -31,6 +31,10 @@ steps:
 - bash: |
     npm install --no-save
     npm install -g vsce
+    # Configure git otherwise `npm version` will fail
+    git config --global user.email "adobuildagent@notanemail.com"
+    git config --global user.name "adobuildagent"
+    npm version 1.4.$(Build.BuildId)
```
mauve: looks good, would be awesome if the 4 would come from a variable as well but we can improve that later

mauve:
@lawrencegripper great idea publishing an artifact automatically, I have been thinking about automatically pushing a terraform-preview package to the marketplace aswell but have not yet figured out how to do that

juliosueiras:
=) added auto github release on tagged commit, so the release flow on the lsp side will be easier,

question: should I include 32-bit build? and also what would be the best tagging convention(like for test release, using -testing? to skip for auto update from vscode-terraform side)

src/languageclient.ts
```diff
@@ -0,0 +1,202 @@
...
+                // Expected
+            }
+            // Todo:
+            // Shutdown method on LSP not implemented :(
+            //    `await langClient.stop();` fails
+            //
+            // We need the process to be stopped so we can overwrite the binary with the new version.
+            // This is a hack for the time beind based on internal property of the languageClient object.
+            const PID = langClient["_serverProcess"].pid;
+            process.kill(PID, 9)
```
lawrencegripper: To remove the need for this nasty hack I've created a PR on the LSP to add a `shutdown` command so `langClient.stop()` will work.

lawrencegripper:
@juliosueiras I'm happy with 64bit builds as I think 32 is an edge case these days, happy to add them in future if that proves to be wrong - sound like a good plan?

For releases would you be able to tag any `testing` releases as `pre-release` in Github when pushed? That way I'll skip pre-release version when prompting to update. 

I've also created a small PR to help with an issue around shutting down cleanly, hope that's ok.

lawrencegripper:
@mauve I've updated the `todo` list in the PR description with the remaining stuff I think is required to get this ready for a preview release. Does this feel about right to you?

mauve:
@lawrencegripper the todo list looks fine for me 👍 great work so far

By introducing a second `templates`-folder with a `.vscode/settings.json` in it you should be able to run the tests a second time with the language server enabled, that then just requires doubling these lines (https://github.com/mauve/vscode-terraform/blob/master/test/run-test.ts#L12-L23) so that the tests are invoked a second time but with the LSP enabled. In order for the tests to actually succeed we might need some way to filter out incompatible tests (for the time being at least).

Unsure if we really need to support both usecases though, the LSP approach is definitely the way forward and as more and more users migrate to 0.12 there won't be that much to save in the old approach. That means that we could just start stripping out the old features and focussing more on the LSP approach, that might suck a little for the 0.11 users but at the same time give us a quicker way forward.

lawrencegripper:
Nice, I'll give that a try and see how I get on. 

@mauve I agree LSP looks like the way forward but it's not a tough job to support both in parallel for the time being. 

The `vscode-go` extension is are doing something similar and as a user I've found myself hopping back and forward quite a bit between the LSP and original behavior. I think it's likely most people have some older `terraform` around which is going to play nicer with the existing behavior and then they can switch when doing `0.12` work. 

The other plus side is we can release this early and not worry about breaking existing `tf 0.11` users while also giving optinos for `tf 0.12` users to have early preview of the LSP. 

Then hopefully at some point, maybe month or weeks, once we've got stats and are happy we can fully cut over. Sound good?

mauve:
Sounds perfect, thanks for your work.

smiller171:
Sadly I was getting a bunch of incorrect errors from then language server, then the language server crashed 5 times so VSCode didn't bring it back

lawrencegripper:
@smiller171 no need for sadly... this is how the it goes, test, improve, test some more. 

What release tag of the `languageServer` where you using? If it was `0.0.3` are you alright to run the `Terraform: Install Language Server command` and retry with `0.0.5`. 

For the errors you saw where they on terraform in subfolders/modules? Any more detail you can share would be ace. If there were in subfolders it might be related to this issue https://github.com/juliosueiras/terraform-lsp/issues/12

smiller171:
@lawrencegripper I'm using `0.0.5` already. I think it's actually This issue: https://github.com/juliosueiras/terraform-lsp/issues/6 though I'm on AWS `2.16.0`

smiller171:
Here's a log and a screenshot. This is from a file in the root of my terraform project. https://gist.github.com/smiller171/e99a298df574031e8d9218bea9a4a439

Screen Shot 2019-07-22 at 10 45 12 AM

smiller171:
It looks like flagging locals is a known issue that hasn't been implemented yet, probably unrelated to the crash

smiller171:
I just noticed this line in the log:
```
reflect.Value.call(0x1a8dcc0, 0x1c20698, 0x13, 0x1bd8646, 0x4, 0xc000a78660, 0x2, 0x2, 0xc0004b4400, 0xc0000c1dc8, ...)
	/home/travis/.gimme/versions/go1.11.12.linux.amd64/src/reflect/value.go:447 +0x454
```
Shouldn't that be looking for a `darwin.amd64` package?

juliosueiras:
@smiller171 I will look into the issue on my side

also I will put in context references for the error checking today/tomorrow and push a new update for it

src/extension.ts
```diff
@@ -57,44 +62,62 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+                // start to build the index
+                if (watcher) {
+                    await watcher.crawl();
+                }
+            }
+    } else {
+        const IndexerNotEnabledCommandHandler = () => {
+            vscode.window.showErrorMessage('Cannot perform action currently using the Terraform Language Server not Indexer.');
+        };
+        Command.dynamicRegister(PlanCommand.CommandName, IndexerNotEnabledCommandHandler);
```
mauve: It makes sense to disable these commands (for now) and bring them back later. Currently the only reason why the `plan` command needs the `IndexAdapter` is to figure out which terraform version is required by the group (`PlanCommand.getExecutable` calls `Runner.getRequiredExecutable`).

I wonder how this can easiest be solved going forward? (@juliosueiras might wanna chime in) It would be nice to be able to know something about the templates from within the vscode-terraform plugin, I am unsure where to draw the line though.

It would be nice if we don't introduce too many dependencies between the plugin and the language-server as that makes deployment and releasing more complicated (in order to simplify things in the future we should introduce a version-check).

Also, I do not know a lot about the language-server-protocol, but I guess we can introduce our own calls so that we can communicate with the server outside of what vscode does itself?

juliosueiras: @mauve just wanted to make sure for this part " I guess we can introduce our own calls so that we can communicate with the server outside of what vscode does itself?", can you give a example of that call? since I am unsure what you mean by that

lawrencegripper: @mauve the LSP has a well defined interface, outside of that it's up to the extension to do the work. https://github.com/Microsoft/language-server-protocol

I'll add an issue to investigate re-instating `plan`, `validate` and other commands that can be picked up as a bit of followup work, worried that if I take on too much in this PR it won't get finished. Sound like a good plan?

mauve: @juliosuerias i was thinking how to implement the plan command. I guess there are two options:

1. Add the command directly to the LSP (if possible)

2. Implement the plan command in the plugin and Expose a custom interaction between the plugin and the LSP which tells us which terraform version is required by a set of templates

This feature is a bit of a corner case, but I implemented it because I have different versions of terraform installed because of conflicting requirements for different templates but this allowed me to have them in the same vscode workspace.

juliosueiras: @mauve for the lsp side, I can look into implementing either [Execute Command](https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand) or [Code action](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction),
I am also thinking of maybe allowing plan on selected resource/data source? what you think?

smiller171: @mauve I've found tfenv to be the best tool for handling multiple versions of Terraform

mauve: @smiller171 `tfenv` thanks for the tip, it looks fine but has a bash dependency and therefore only works on windows using git-bash or wsl (that feels like a show-stopper for us). (also it seems it cheats when picking the version).

mauve: @juliosueiras aha so that means you can implement the plan command directly (as is) in the LSP. I don't know what the best way forward is to be honest moving as much as possible into the LSP or having a split.

The benefit of writing features directly in the plugin is that it is relatively easy to debug and develop and that I can stacktraces and exceptions reported to my Azure AppInsights.

The benefit of writing features in the LSP is that is where all the data is.

It might be possible to do a split but quite honestly it seems like having everything in one place is probably better.

For now though I think we should wait with `plan` and friends and focus on doing one release with the LSP to get data. If we can initialize the LSP with the AppInsights key you could use the AppInsights-Go SDK to track exceptions in the LSP. I can pay for the metrics and give you guys access to my dashboard with stats and crash data (beware that I have currently disabled the AppInsights stuff but can enable again when we do the first release with the LSP).

package.json
```diff
@@ -207,6 +210,34 @@
...
           "description": "Indexes all terraform sources to get live syntax errors, rename support, go to symbol, and much more..."
         },
+        "terraform.languageServer": {
+          "type": "object",
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "description": "Enables or disables the indexing",
+              "default": false
+            },
```
chucknelson: It seems like it might be more clear for users if we expose these config options as standard controls in extension settings versus having them manually edit `settings.json` and figure out what the available options are.

Is there a particular reason we have them all hidden in an `object` type? Could we maybe [expose the obvious ones](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema), like `enabled`, via normal namespacing (i.e., `terraform.languageServer.enabled`)? This would apply to `terraform.indexing...` as well.

lawrencegripper: I'm sticking with the existing settings infrastructure for now as I think there is already enough churn in this PR without refactoring the settings stuff. My recommendation is that this could be picked up in a different PR if @mauve agrees once back from hols.

lawrencegripper:
@mauve Have a good hols! I'm pretty happy with the state of things now. Have a basic set of passing integration tests:

image

I've also added the last two features I see as required for basic first experimental release: 

1. The extension now shows a dialog on startup if a newer version of the `lsp` is availble
2. If the user has `terraform 0.12` on their `$path` the extension will show a dialog suggesting they try out the `lsp`

lawrencegripper:
@chucknelson To handle the case of uninitialized `.tf` folders (ie. don't contain and can't contain a `.terraform` folder). I've added in a step during install which adds common `providers` into the `./lspbin` folder so they'll get discovered. 

This is essentially a workaround for this issue: https://github.com/juliosueiras/terraform-lsp/issues/12 and had, hopefully, a big impact on the usability of the LSP while that is being fixed. 

Can you give it a test for me?

lawrencegripper:
@mauve Hope it's ok, I've made a minor tweak to the `gulp` so that it keeps a copy of the `hcl-hils.js` committed in the repo. This then allows the build (hopefully) to run cross plat, felt like a fair trade off to me. Let me know if you'd like me to roll back.

mauve:
Yeah I was about to checkin hcl-hil.js exactly for that reason but didn't have any time before I left for vacation.

lawrencegripper:
Cool, was hoping for things to be ready to merge when you got back but looks like the integration tests have some stability problems now I've added more platforms to run on. 

Will work through them but I'm now off on holiday for a week so will be deplayed until I'm back. Happy for you/anyone to work on it while I'm away or wait for me to be back ... whichever works. 

In terms of usability I've been running this locally for my TF work over the last few days and while the LSP does crash, fixed by reloading the vscode windows, I'm pretty happy with the usability improvement. Autocomplete working pretty well for most things.

chucknelson:
@lawrencegripper - About to start testing the common provider change, but the [latest successful build artifact](https://mauvezero.visualstudio.com/vscode-terraform/_build/results?buildId=57&view=results) is...300+ MB vs ~19 MB for prior versions. Is that expected?

lawrencegripper:
Wow, no I wasn't expecting that..... ooooooh! I think where I've run the integration tests in the build the providers have all been pulled down and included in the `visx`. One moment let me try and fix that up.

lawrencegripper:
@chucknelson fixed. Good spot, thanks!

chucknelson:
@lawrencegripper - Tested the latest build artifact (`v1.4.61`), and the provider plugin downloads worked as expected. Base on LSP output it seems to be invoking them and succeeding. As you probably already expected, current caveats are:
* It can take quite a while to download all of the providers (network is a little slow today, so probably made it worse)
* Providers are only latest versions as-of download and don't respect versions specified in a specific terraform config

Using the `0.0.5` version of the LSP, I'm getting a lot of crashes per the output for things like hover, but I may test that out more and report to @juliosueiras if it's something I can pinpoint.

lawrencegripper:
Thanks for trying it out @chucknelson, yeah those limitations are known ones that I don't have time to address I'm afraid - likely best to think of the default providers as a temporary stand-in until something more permanent can be done.

src/extension.ts
```diff
@@ -143,6 +147,17 @@ export async function activate(ctx: vscode.ExtensionContext) {
     }
 }
 
+function informUserAboutLspIfTf12() {
+    try {
+        const versionResponse = cp.execSync("terraform --version");
```
mauve: probably better to use `Runner.getRequiredExecutable(VersionRequirement.parse(">= 0.9.1"))` that way you will check for all configured terraforms (as you can specify multiple executables via config)`

src/languageclient.ts
```diff
@@ -0,0 +1,266 @@
...
+                    vscode.window.showErrorMessage(`Failed starting language server. Error: ${e}`);
+                    reject(e);
+                }
+            });
+        });
+
+    }
+
+    private async installCommonProviders(serverLocation: string) {
+        let terminal = vscode.window.createTerminal("Terraform: Install providers");
```
mauve: do you think it makes sense to create a `Terraform` terminal and reuse all over the place?

src/languageclient.ts
```diff
@@ -0,0 +1,266 @@
...
+     * @param  {String} startPath    Path relative to this file or other file which requires this files
+     * @param  {String} filter       Extension name, e.g: '.html'
+     * @return {Array}               Result files with path string in an array
+     */
+    private findFilesInDir(startPath: string, filter: string) {
+
+        let results = [];
+
+        if (!fs.existsSync(startPath)) {
+            console.log("no dir ", startPath);
```
mauve: for debugging purposes it might be better to use the output tab which is used by the rest of the plugin.

src/languageclient.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Find all files recursively in specific folder with specific extension, e.g:
+     * findFilesInDir('./project/src', '.html') ==> ['./project/src/a.html','./project/src/build/index.html']
+     * @param  {String} startPath    Path relative to this file or other file which requires this files
+     * @param  {String} filter       Extension name, e.g: '.html'
+     * @return {Array}               Result files with path string in an array
+     */
+    private findFilesInDir(startPath: string, filter: string) {
```
mauve: isn't there some `glob()` function we can use from somewhere else? I thought we even require it already (maybe only for building and maybe only transitively)?

test/completion-provider.test.ts
```diff
@@ -1,18 +1,18 @@
 import * as assert from 'assert';
 import * as vscode from 'vscode';
 
-async function executeProvider(uri: vscode.Uri, position: vscode.Position): Promise<vscode.CompletionList> {
+export async function executeProvider(uri: vscode.Uri, position: vscode.Position): Promise<vscode.CompletionList> {
```
mauve: I would prefer (but not super-important) that these files move to a separate file if they are to be reused

src/commands/installLanguageServer.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const serverLocation = getConfiguration().languageServer.pathToBinary || Path.join(this.ctx.extensionPath, "lspbin");
+
+    // Create dir for lsp binary if it doesn't exist
+    if (!fs.existsSync(serverLocation)) {
+      fs.mkdirSync(serverLocation);
+    }
+    const downloadedZipFile = Path.join(serverLocation, 'terraform-lsp.tar.gz');
+    // Download language server
+    const octokit = new Octokit();
```
mauve: the integs are failing because of rate-limiting on macos (probably because the linux started earlier): we probably want to add some simple retry and back-off logic here to ensure that this works consistently

mauve:
@lawrencegripper I added some more comments, in general it looks really good. I want to take the time again to thank you for your effort (as well as @juliosueiras of course) a few more changes and we will be able to do a release.

@juliosueiras I want to reiterate my offer to pay for AppInsights (by means of giving you access to the telemetry and crash logs in my accounts) if you integrate the SDK in the LSP, this might be a really good thing to do before the first release as we can expect somewhere around 80k monthly actives. (also thanks for kicking of the LSP which will allow us to bring TF 0.12 support to vscode-terraform)

lawrencegripper:
@mauve thanks for taking a look over these. I’ll have a look at fixing them up after my holiday. 

Regarding the github API rate limits causing issues with the integration test, the limit is per hour based on IP and with the build agents being shared this isn’t ideal as another persons build could have exhausted the limit and then we’d have to wait an hour. I think best approach is to mock this call out, agree?

azure-pipelines.yml
```diff
@@ -42,11 +46,25 @@ steps:
...
 - bash: |
-    npm run integration-test
+    npm run compile && npm run integration-test
   displayName: Running integration tests
   env:
     DISPLAY: ':99.0'
 
 - bash: |
+    # clean out providers and lsp 
+    rm -r ./lspbin
```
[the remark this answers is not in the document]
lawrencegripper: Ah sorry I misunderstood the earlier comment, didn't realize there was a specific `.vscodeignore` file. Added `lspbin` to it and remove this workaround.

src/commands/index.ts
```diff
@@ -3,8 +3,10 @@ import { IndexAdapter } from "../index/index-adapter";
 import { Command, CommandType } from "./command";
 
 export class IndexCommand extends Command {
+  public static readonly CommandName = "index-document";
```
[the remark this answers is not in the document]
lawrencegripper: I went with this approach as I needed a way to get the command name without registering the command with vscode.

If I did `new IndexCommand().command()` this would register the command into `vscode` as the constructor does this work. When wiring up the placeholder commands I need a way to get the command name without wiring up the command. 

For example:

`Command.dynamicRegister(IndexCommand.CommandName, IndexerNotEnabledCommandHandler);`

src/commands/installLanguageServer.ts
lawrencegripper: So looking at the API restrictions they're pretty broad. 

```
For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.
```
So it feels like backing off for an hour might not be a good option!

One option is to require users to login to github. I don't think this is a good UX so ruled that out. 

The second option I've gone with is to limit the number of calls I'm making and be much more efficient. I've

1. Halved the number of calls calls when doing at `Install`.
2. Made `Update` only check once a day (think this will have a big impact when running integration tests as this may have been firing for each test). 

Hopefully this get this sorted. Will test today and see if the rate limit issue re-occurs with these changes.

src/languageclient.ts
lawrencegripper: Cool, out of practice with `node`, I've moved this over to use `glob` from `npm`. :)

src/extension.ts
lawrencegripper: Ooh interesting. I took a look and I'm a bit worried that if they don't have `tf 0.12` then this would thow an `indexOutOfRange` error due to the the `[0]` at the end of the line. Is that right?

```
 getRequiredExecutable(requirement: VersionRequirement): TerraformExecutable {
    const supported = this.executables.filter(e => e.version && requirement.isFulfilledBy(e.version));
    return supported.sort((left, right) => TerraformVersion.compare(left.version, right.version))[0];
  }
```

azure-pipelines.yml
```diff
@@ -6,10 +6,10 @@ strategy:
     linux:
       imageName: 'ubuntu-16.04'
     # The build needs docker so skipping macOS for now
-    # mac:
-    #  imageName: 'macOS-10.14'
+    mac:
+     imageName: 'macOS-10.14'
     # Docker fails on Windows, temp. disabling for now, it is
-    # time to just bundle hcl-hil.js directly
+    # # time to just bundle hcl-hil.js directly
```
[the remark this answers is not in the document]
lawrencegripper: Currently still seeing issues with Windows integration tests. Will take a look and hopefully uncomment shortly.

azure-pipelines.yml
```diff
@@ -42,11 +46,25 @@ steps:
   displayName: Running unit tests
 
 - bash: |
-    npm run integration-test
+    npm run compile && npm run integration-test
```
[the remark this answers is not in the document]
lawrencegripper: Was a bit of a late night change but I was some odd behavior in the build build without this present. It looked like some of the test code wasn't being compiled but probably just my tired brain. Will take a look now.

src/languageclient.ts
```diff
@@ -0,0 +1,266 @@
...
+            // This is a hack for the time beind based on internal property of the languageClient object.
+            const PID = langClient["_serverProcess"].pid;
+            process.kill(PID, 9)
+        } catch {
+            // May occur until PR merged
+        }
+
+        this.isRunning = false;
+    }
+    public async start(prompt: boolean = true) {
```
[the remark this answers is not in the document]
lawrencegripper: Fixed

src/configuration.ts
```diff
@@ -7,6 +7,14 @@ export interface TerraformIndexConfiguration {
   exclude: string[];
 }
 
+export interface TerraformLanguageServerConfiguration {
+  enabled: boolean;
+  args: boolean;
+  pathToBinary: string;
+  installCommonProviders: boolean;
+}
+
```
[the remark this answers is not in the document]
lawrencegripper: Sorry fixed :)

src/commands/installLanguageServer.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // What releases are available?
+    const availableReleses = await octokit.repos.listReleases({
+      owner: 'juliosueiras',
+      repo: 'terraform-lsp'
+    })
```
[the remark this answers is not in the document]
lawrencegripper: I've add a semi-colon to the tslint rules and a couple of helpers `npm run tslint` and `npm run tslint-fix` and pushed those changes.

src/languageclient.ts
lawrencegripper: Agreed, as I've moved to glob now these aren't needed but will use `Logger` if I have any other output to display.

azure-pipelines.yml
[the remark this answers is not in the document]
lawrencegripper: This was catching a valid issue with the provider install stuff. Now fixed :+1:

src/languageclient.ts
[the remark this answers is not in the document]
lawrencegripper: Really odd, latest changes in my branch aren't showing in this PR?!

https://github.com/lawrencegripper/vscode-terraform/blob/960a509aced9239647403daf50c0e35b36ed2d8b/src/languageclient.ts#L60

test/codelense.test.ts
```diff
@@ -57,6 +57,10 @@ resource "null_resource" "reference_test" {
     assert(successful, "forced indexing not successful");
 
     let lenses = await vscode.commands.executeCommand('vscode.executeCodeLensProvider', doc.uri, 10) as vscode.CodeLens[];
+    if (!lenses) {
+      // Retry this seems a bit flakey
+      lenses = await vscode.commands.executeCommand('vscode.executeCodeLensProvider', doc.uri, 10) as vscode.CodeLens[];
+    }
```
lawrencegripper: flakeytest

I've seen this test be a bit flakey when testing locally and in the build. While not ideal a retry seems to handle this case (sometimes `lenses` is returned as `undefined`) 

Build example: https://mauvezero.visualstudio.com/vscode-terraform/_build/results?buildId=69

src/languageclient.ts
lawrencegripper: I like the idea, the problem would be that I wait on the process exiting to know the install of the providers is complete so I kinda need it to be single use.

bitfield:
I can successfully install and run the language server, but hovering over any item in the code window causes a crash:

```
2019/08/09 16:11:57 Server started
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x14d70ac]

goroutine 43 [running]:
github.com/hashicorp/terraform/lang.(*Scope).evalContext(0xc0000ab0e0, 0xc000438100, 0x1, 0x1, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0)
	/home/travis/gopath/pkg/mod/github.com/hashicorp/terraform@v0.12.5/lang/eval.go:184 +0x11c
github.com/hashicorp/terraform/lang.(*Scope).EvalContext(0xc0000ab0e0, 0xc000438100, 0x1, 0x1, 0x1, 0x0, 0x0, 0x0)
	/home/travis/gopath/pkg/mod/github.com/hashicorp/terraform@v0.12.5/lang/eval.go:160 +0x5f
github.com/hashicorp/terraform/lang.(*Scope).EvalExpr(0xc0000ab0e0, 0x1da8e00, 0xc00019c1e0, 0x1da9ac0, 0x2542040, 0x1a82940, 0xc00020c000, 0x1bda9e3, 0x6, 0xc0003a2160, ...)
	/home/travis/gopath/pkg/mod/github.com/hashicorp/terraform@v0.12.5/lang/eval.go:87 +0xa0
main.TextDocumentHover(0x1da87c0, 0xc000090780, 0xc00003e120, 0x56, 0x7, 0x8, 0x0, 0x0, 0x0, 0x0, ...)
	/home/travis/gopath/src/github.com/juliosueiras/terraform-lsp/main.go:549 +0x28c
reflect.Value.call(0x1a8dcc0, 0x1c20698, 0x13, 0x1bd8646, 0x4, 0xc0000908d0, 0x2, 0x2, 0x2525040, 0xc0001b7dc8, ...)
	/home/travis/.gimme/versions/go1.11.12.linux.amd64/src/reflect/value.go:447 +0x454
reflect.Value.Call(0x1a8dcc0, 0x1c20698, 0x13, 0xc0000908d0, 0x2, 0x2, 0x1a313c0, 0xc0000908d0, 0x1)
	/home/travis/.gimme/versions/go1.11.12.linux.amd64/src/reflect/value.go:308 +0xa4
reflect.Value.Call-fm(0xc0000908d0, 0x2, 0x2, 0x1, 0xc00000a660, 0x1)
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/handler/handler.go:211 +0x60
github.com/creachadair/jrpc2/handler.newHandler.func7(0x1da87c0, 0xc000090780, 0xc000043ac0, 0xc000043b30, 0xc000060ee8, 0x10, 0xc000060ee0)
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/handler/handler.go:222 +0x22b
github.com/creachadair/jrpc2/handler.Func.Handle(0xc00043a5e0, 0x1da87c0, 0xc000090780, 0xc000043ac0, 0x0, 0x0, 0x1da87c0, 0xc000090780)
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/handler/handler.go:23 +0x44
github.com/creachadair/jrpc2.(*Server).invoke(0xc0004ac000, 0x1da8700, 0xc000432c00, 0x1d9d820, 0xc00043a5e0, 0xc000043ac0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/server.go:294 +0x1b0
github.com/creachadair/jrpc2.(*Server).dispatch.func1(0xc0004ea520, 0xc0004ac000, 0xc0002f2510)
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/server.go:185 +0x101
created by github.com/creachadair/jrpc2.(*Server).dispatch
	/home/travis/gopath/pkg/mod/github.com/creachadair/jrpc2@v0.1.2/server.go:183 +0x13a
[Info  - 4:12:14 PM] Connection to server got closed. Server will restart.
[Error - 4:12:14 PM] Request textDocument/hover failed.
Error: Connection got disposed.
	at Object.dispose (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/main.js:876:25)
	at Object.dispose (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-languageclient/lib/client.js:71:35)
	at LanguageClient.handleConnectionClosed (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-languageclient/lib/client.js:2153:42)
	at LanguageClient.handleConnectionClosed (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-languageclient/lib/main.js:151:15)
	at closeHandler (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-languageclient/lib/client.js:2140:18)
	at CallbackList.invoke (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/events.js:62:39)
	at Emitter.fire (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/events.js:120:36)
	at closeHandler (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/main.js:226:26)
	at CallbackList.invoke (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/events.js:62:39)
	at Emitter.fire (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/events.js:120:36)
	at StreamMessageReader.fireClose (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/messageReader.js:111:27)
	at Socket.listen.readable.on (/Users/john/.vscode/extensions/mauve.terraform-1.4.50/node_modules/vscode-jsonrpc/lib/messageReader.js:151:46)
	at Socket.emit (events.js:187:15)
	at Pipe.Socket._destroy._handle.close (net.js:606:12)
2019/08/09 16:12:14 Server started
```

lawrencegripper:
@bitfield thanks for testing, this is one to raise in the `terraform-lsp` repository as fixing up bugs in that code are out of scope for this PR.

> This PR adds code to download and wire up the language server. I'm not building the language server and bugs with it sit outside of the scope of this PR. They should be raised here https://github.com/juliosueiras/terraform-lsp

@mauve How we looking on this now? Happy to roll out as an experiment?

tomasaschan:
I'm really stoked that this seems to be so close to an actual release - good work, everyone involved!

I've skimmed through this thread for instructions on how to install this branch in my VS Code (if it works I'll be super happy - if not, you'll have another Windows tester reporting errors), but can't figure out what I need to do. Are there instructions somewhere on how to obtain the latest version of the extension per this branch, and install them locally? Or is the actual release of the new extension version so close that I should just wait it out?

tam116:
@tomasaschan You can find the latest at https://mauvezero.visualstudio.com/vscode-terraform/_build?definitionId=5 -> click the latest build -> click Summary -> download the build artifact (visxPackage) -> unzip -> From Vscode open the command palette (ctrl+shirt+p) -> search for install from visx and choose the visx file you unzipped earlier.

tomasaschan:
Hm. I immediately ran into some issues; should I post them here, or as new issues?

src/languageclient.ts
```diff
@@ -0,0 +1,281 @@
...
+provider "null" {}
+provider "external" {}
+provider "template" {}
+provider "archive" {}
+                        `);
+        }
+        const defaultProvidersPluginsPath = Path.join(serverLocation, ".terraform", "plugins");
+        if (!fs.existsSync(defaultProvidersPluginsPath)) {
+            terminal.show();
+            terminal.sendText(`cd ${serverLocation}`);
```
tomasaschan: The server location url needs to be quoted here; otherwise it can fail for invalid paths (e.g. it fails on Windows if the user name contains a space).

tomasaschan:
Found another thing (after manually doing what the problematic installation script tried to do):

When I hover anything, I get:
```
2019/08/13 18:23:18 Server started
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x58 pc=0x8d6493]
```
[full log](https://gist.github.com/tomasaschan/bceb3e0117e92d62239eca60ba6dd9cd)

lawrencegripper:
@tomasaschan Thanks this is reported here: https://github.com/mauve/vscode-terraform/pull/200#issuecomment-519957698 and as mentioned isn't in scope this this as it's an issue with the LSP. Would you be alright to raise an issue here: https://github.com/juliosueiras/terraform-lsp

tomasaschan:
@lawrencegripper Ah, thanks - didn't read that closely enough to realize that it was the same issue. @bitfield did you file an issue or should I?

When I get back to work tomorrow (UTC+2 here) I can test the latest build and verify that installation works with the path fix.

lawrencegripper:
@tomasaschan think I lost a comment saying thanks for spotting the windows issue with longer paths. Good catch :)

tomasaschan:
Tried again with build 1.4.75 now, and installation seems to work as expected!

(The problem was not long paths on Windows, but paths with spaces in them, which would not have worked on any platform. But that's why OSS is nice - if you find something that's not working, you can help fix it 😄)

jrmannby:
Does anyone have a clue when this will get merged?

mauve:
tonight

lawrencegripper:
Exciting!

Just to temper excitement. The language server work is still ongoing so this is experimental at this stage and likely will have stability issues. Once merged best to follow that work here: https://github.com/juliosueiras/terraform-lsp

juliosueiras:
will work through the current issue and implement a few new features and
fixes in these few days, just being busy lately with work

tomasaschan:
@juliosueiras Ping me if you need help with testing on Windows. I have a fairly large workspace I'm working on for my day job, and I already have the extension set up to use the language server, so it's very little extra work to install a new LS version and try stuff out.

thefirstofthe300:
I'd love to help out here by testing things. I was looking to download the build artifacts from the Azure pipeline build but I may not have access as I am unable to see any builds. Is there a build (Windows or Linux) that I can download to help out testing?

Or better yet, are these changes going to be merged and release soon?

charlex:
Like @thefirstofthe300, I too am not seeing any builds in the Azure pipeline. The only VSIX builds I am able to download come from the above comments, but I don't think these are up to date.

Could someone post the latest ZIP to allow folks to download?

mauve:
Thanks everybody!

Merged to master and released as 1.4.0 :) 🎉

Feel free to post GIFs below.

charlex:
You are the best! Thank you!

thefirstofthe300:
Sweet!

bitfield:
Great! Do you have a build we can download?

tomasaschan:
@bitfield Version 1.4 is available through the regular installation channels now (although I had to uninstall my VSIX-installed version to see it). Just uninstall the extension, find it in the extension marketplace and install, and you'll have the latest!

bitfield:
Thank you! Just one minor problem; when I install the new version it prompts me in *every* open window to install the language server. I have approximately 20 open windows. Having installed the language server in one of them, all the other windows are now in some intermediate state of (1) prompting me to install the server, (2) prompting me for a version, (3) complaining that I must choose a version to continue, or (4) downloading the language server _again_.

This seems like one of those things that should only happen in the current window.

PR #225, opened 2019-10-02
Add `hcl` as lanugage id in markdown

stuartleeks:
The Terraform provider docs use `hcl` as the language identifier for code blocks in markdown rather than `terraform`. This PR enables the embedded codeblock support for `hcl`

With this PR, both code blocks below get language support (not just the second one):

    ```hcl
        resource "test" "test {
        }
    ```
    ```terraform
        resource "test" "test {
        }
    ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,8 +136,15 @@
         "path": "./syntaxes/terraform.json"
       },
       {
-        "scopeName": "text.markdown.terraform.codeblock",
+        "scopeName": "text.markdown.terraform.codeblock-terraform",
         "path": "./syntaxes/terraform.codeblock.json",
+        "injectTo": [
+          "text.html.markdown"
+        ]
+      },
+      {
+        "scopeName": "text.markdown.terraform.codeblock-hcl",
+        "path": "./syntaxes/terraform.codeblock-hcl.json",
         "injectTo": [
           "text.html.markdown"
         ]
--- a/syntaxes/terraform.codeblock-hcl.json
+++ b/syntaxes/terraform.codeblock-hcl.json
@@ -1,22 +1,22 @@
 {
 	"$schema": "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json",
-	"name": "sltest",
+	"name": "sltest-hcl",
 	"fileTypes": [],
 	"injectionSelector": "L:markup.fenced_code.block.markdown",
 	"patterns": [
 		{
-			"include": "#terraform-code-block-terraform"
+			"include": "#terraform-code-block-hcl"
 		}
 	],
 	"repository": {
-		"terraform-code-block-terraform": {
-			"begin": "(?<=[`~])terraform(\\s+[^`~]*)?$",
+		"terraform-code-block-hcl": {
+			"begin": "(?<=[`~])hcl(\\s+[^`~]*)?$",
 			"end": "(^|\\G)(?=\\s*[`~]{3,}\\s*$)",
 			"patterns": [
 				{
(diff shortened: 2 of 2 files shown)


stuartleeks:
@mauve - I'm not sure why the build is failing based on the changes I've made. Any ideas?

mauve:
Weird. The constants.json file is supposed to be checked in (and empty 😂) I don't think there are any tests for your additions so lgtm.

Thanks for your contribution and thanks for using my plugin, sorry that I cannot dedicate so much time to it anymore.

paultyng:
For now we've decided not to support generic HCL in this extension. I did notice we did remove general `terraform` injection to markdown though, so I've opened #403 to add that back.

stuartleeks:
> For now we've decided not to support generic HCL in this extension
@paultyng  - would you be open to sharing the reasons for that? I'm interested to understand :-)

paultyng:
Sure happy to. I'm not opposed to revisiting, the main thought process here is that [semantic highlighting is moving to the language server](https://github.com/microsoft/language-server-protocol/issues/18). 

When we can use the native Go HCL implementation in a language server to perform the highlighting, I think it makes a lot more sense to revisit this functionality. We are hoping longer term that our work in the language server and with schema on top of HCL allows for more generic implementations of many different types of HCL based language servers (for other HashiCorp tools or for anyone else who needed/wanted one).

Is there a specific `hcl` language file you think this extension should support? I read this as just general HCL support, but perhaps there is something more specific you want this tool to handle? I typically use `terraform` in my Markdown codeblocks, but is there a reason people may want to use `hcl` instead for `terraform` code?

paultyng:
Sorry I see you are referencing the provider docs here specifically. I should do a quick audit of those to see if this is still as prevalent.

stuartleeks:
Thanks! If terraform codeblocks should be marked as terraform rather than hcl then it would be good to ensure the provider docs are updated 😀

paultyng:
Yeah, interestingly I have been writing some code generation for docs that uses `terraform` and I realize it doesn't highlight properly in the registry, so this has pointed out a number of issues I need to look in to next week 😅

https://registry.terraform.io/providers/paultyng/unifi/latest/docs/resources/wlan

radeksimko:
I understand it may be daunting and take time, but I think it would be best to get those markdown blocks updated to `terraform` and make sure that `terraform` as syntax has a solid support on GitHub and elsewhere, as opposed to keep treating generic HCL the same as Terraform, because they are not the same.

radeksimko:
Hi @stuartleeks 
Sorry for the long delay in reviewing your PR.

The grammar which is currently in use by the extension is largely Terraform-specific, which is why I'd be very hesitant to highlighting just any "HCL" code with it, as this PR proposes.

We do have plans to publish a more generic grammar as per https://github.com/hashicorp/vscode-terraform/issues/638 and it is likely that this grammar would replace the existing one, but even then such grammar would - in the context of this (Terraform) extension - would still be used to highlight **Terraform** code, not HCL.

With that in mind I would be open to a PR of a smaller scope which highlights snippets in markdown of `terraform` language.
On a separate note there is a number of other PRs and issues concerning syntax and before tackling any of these we will need some test coverage, as outlined in https://github.com/hashicorp/vscode-terraform/issues/735, to be provide some stability and ensure that any changes to the grammar don't break things.

As this PR has now diverged significantly from `main` I'm going to close it. Also I'd like us to first introduce the tests for the existing (currently untested) grammar, before adding or modifying any more grammar. Once that is done (i.e. #735 is closed) we'd certainly welcome that smaller-scope PR with relevant tests attached.

Thank you for your contribution and understanding.

PR #299, opened 2020-04-03
support strings containing escaped doublequotes and hashmarks

### LINKED ISSUE #194: Syntax highlighting: # and \" inside strings
1. `#` inside `request_mapping_template = "## dummy\n{}"` is rendered as a beginning of a comment. They should not be highlighted.
2. Escaped double quotes `\"` are rendered as starting/ending a string value. They should be ignored. 

image

dwarburt:
resolves #194

### DIFF
--- a/syntaxes/terraform.json
+++ b/syntaxes/terraform.json
@@ -274,7 +274,11 @@
           "include": "#string_interpolation"
         },
         {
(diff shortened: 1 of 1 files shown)


dwarburt:
This is not needed anymore, current `master` branch handles this correctly.

PR #312, opened 2020-05-01
Remove duplicate commands

rf5860:
Several of the commands are duplicated (https://github.com/mauve/vscode-terraform/issues/285)


aeschright:
Hi @rf5860, thanks for spotting that! We're addressing the commands listed with #325 so I'm going to close this one.

PR #334, opened 2020-06-04
Separate toggle command for language server into enable and disable

### LINKED ISSUE #307: Command Palette entries named "Enable/Disable" are confusing
# Issue

Currently, the command pallete has the function "Terraform: Enable/Disable Language Server".  This does not follow the standard VSCode convention of having separate commands to Enable and Disable a feature.  This leads to confusion where people are not aware of the existing state of a feature and must toggle on and off to experiment.

# Resolution

Create separate commands to Enable and Disable the terraform language server.

aeschright:
Fixes #307


src/extension.ts
```diff
@@ -47,15 +47,14 @@ export function activate(context: vscode.ExtensionContext) {
...
-				useLs = true;
-				installThenStart(context, config);
-			}
-			config.update("languageServer.external", useLs, vscode.ConfigurationTarget.Global);
+			installThenStart(context, config);
+			config.update("languageServer.external", true, vscode.ConfigurationTarget.Global);
+		}),
+		vscode.commands.registerCommand('terraform.disableLanguageServer', () => {
+			stopLsClient();
+			config.update("languageServer.external", false, vscode.ConfigurationTarget.Global);
```
paultyng, line 57: Probably not a big deal, but should the config update here be first, in case stopping causes an error or something?

aeschright: Updated

Issue #335, opened 2020-06-05
Add easier way to toggle between terraform-ls / terraform-lsp but keeping auto-updating


paultyng:
While the feature sets are diverge more significantly, it may be useful to let users easily opt between language servers (but retaining installation and auto-updating). This should still default to the `terraform-ls` option.

paultyng:
I think we probably need a few UI options and some advanced options for this:

## UI Options:

* Language Server: `hashicorp/terraform-ls` (Default), `juliosueiras/terraform-lsp`, `Manual (set path in advanced)`, `None`

## Advanced Options

* Path to binary (for `Manual` only)
* Args (for any ls)

Manual would require you to specify a path for it to work, otherwise would have a warning or error.

For `terraform-lsp` it should use the GitHub release auto updating functionality if possible.

paultyng:
We need to make sure for the binary path, we are considering the scoping, it probably needs to just be `machine`, I have issues with the current scoping when I toggle between Windows/WSL.

jpogran:
We now include terraform-ls in the extension VSIX, so we do not update terraform-ls out of band.

Scoping of machine related settings has been adjusted in https://github.com/hashicorp/vscode-terraform/pull/1164, so that part is taken care of now.

VS Code separates extension settings by leading identifier, so `terraform.languageServer.pathTobinary` is technically a different extension setting than `terraform-ls.rootModules`. This is an somewhat unwritten convention that is upheld by every extension in existence. We've standardized all the extension names in https://github.com/hashicorp/vscode-terraform/pull/1156, so any language server planning on integrating with this extension should use the `terraform.languageServer*` settings provided here or request an additional setting added for their extension.

This ticket is complete

PR #342, opened 2020-06-05
Modify configuration, remove unused settings

paultyng:
@aeschright this may be better to rework as part of #335


jpogran:
Closing this as the extension settings moved on abit since this was introduced. We're reviewing the current settings and scope as part of ticket #791

PR #349, opened 2020-06-08
Enable syntax highlighting for tfvars

### LINKED ISSUE #348: Add syntax highlighting for .tfvars files
`.tfvars` files highlighting support should be able to reuse the HCL hightlighting.

aeschright:
Fixes #348


aeschright:
> Am I right that this is just enabling syntax highlighting, but will _not_ send tfvars files to the language server yet? If so, then 👍 :shipit:

Yep, that's correct.

PR #360, opened 2020-06-10
Create nightly (or every push) builds of VSIX

paultyng:


aeschright:
Looks like your README commit got thrown in here too.

PR #388, opened 2020-06-13
Automatically publish extension to OpenVSX (GitHub workflow)

jankeromnes:
Hello! 👋 Many thanks for the cool Terraform extension!

Fixes https://github.com/hashicorp/vscode-terraform/issues/379. This proof-of-concept GitHub workflow would:
- Run on every push to `master`
- Publish the Terraform VS Code Extension to https://open-vsx.org (using the `ovsx` CLI)

It requires:
- [ ] Setting up an `OVSX_PAT` GitHub secret with value from https://open-vsx.org/user-settings/tokens

Notes:
- ~~⚠️ Untested~~
- This workflow can probably be extended with a similar job to automate the publishing to VS Code Marketplace

### DIFF
--- a/.github/workflows/publish-extension.yml
+++ b/.github/workflows/publish-extension.yml
@@ -1,0 +1,20 @@
+name: Publish Terraform VS Code Extension
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish_to_openvsx:
+    name: publish to open-vsx.org
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '12'
+    - run: npm install
+    - run: npx ovsx publish
+      env:
+        OVSX_PAT: ${{ secrets.OVSX_PAT }}


jankeromnes:
Update: The GitHub workflow [seems to work as expected](https://github.com/jankeromnes/vscode-terraform/runs/768085250) on my fork, right up to the `npx ovsx publish` call.

(That call then fails with the error "A personal access token must be given", because I haven't set up the `OVSX_PAT` GitHub secret on my fork, because I don't want to accidentally publish your extension under my account.)

paultyng:
@jankeromnes thanks for this, should we just publish actual releases, or every push/commit?

jankeromnes:
@paultyng You're very welcome. The proposed GitHub Action publishes on every push to `master`, but it will do nothing if the version in `package.json` is already available on https://open-vsx.org (i.e. the GitHub Action will only publish once, whenever the `package.json` version is bumped).

If you have a dedicated mechanism for publishing actual releases, it may be better to run the `npx ovsx publish` there, instead of always publishing from `master` (especially if you do `-dev` versions, release candidates, etc. that you wouldn't want regular users to install).

I'm happy to adjust the PR if you point me to the official publishing script, but any way works for OpenVSX, so you can choose which option you prefer.

jankeromnes:
(Friendly ping @paultyng @appilon @radeksimko 🙂)

jpogran:
Apologies @jankeromnes for the radio silence. I see you've gone ahead and setup the automatic publishing through the generic OpenVSX account so that the extension is available in the OpenVSX gallery.

We've recently added automatic publsihing for the VS Code Marketplace in https://github.com/hashicorp/vscode-terraform/pull/631, so we'll add this action to that workflow. You would need to rebase your PR and move these changes to that file, but likely since there has been so much time since you contributed this PR I can understand if you don't have time to do that. If so, I'll handle that when the account for the OpenVSX gallery is created and setup. 

I'm going to mark this as draft until then and add it to our planning board.

filiptronicek:
@jpogran dropping in to the conversation to say that this still needs some time to work as expected, because OpenVSX doesn't yet support the target platform feature, which this extension uses. After the feature lands (https://github.com/eclipse/openvsx/pull/406), I can take over this Pull Request to get it working (if @jankeromnes lets me 😛). 

This is one of the biggest extensions on OpenVSX and is currently failing to publish (there it is still at https://github.com/hashicorp/vscode-terraform/releases/tag/v2.18.0), so I would loooove to see this merged soon 🤞🧡.

jankeromnes:
Hi @filiptronicek, many thanks for the offer! 🙏 I've added you as a collaborator on my fork, which should allow you to update this Pull Request. Please let me know if I can help with anything else.

jankeromnes:
Closing in favour of https://github.com/hashicorp/vscode-terraform/pull/1064

PR #414, opened 2020-06-23
Check shasum of language server download

aeschright:


src/languageServerInstaller.ts
```diff
@@ -90,27 +93,33 @@ export class LanguageServerInstaller {
...
-				// No matching build found
-				return Promise.reject();
-			}
+			// No matching build found
+			return Promise.reject();
+		}
+		try {
+			this.removeOldBinary(installDir, platform);
+		} catch {
+			// ignore
```
appilon, line 122: no need to log?

src/languageServerInstaller.ts
```diff
@@ -120,17 +129,36 @@ export class LanguageServerInstaller {
...
 			}, (progress, token) => {
 				token.onCancellationRequested(() => {
+					// TODO clean up partial downloads
 					return reject();
 				});
 
-				progress.report({ increment: 10 });
+				progress.report({ increment: 30 });
+
 				return new Promise<void>((resolve, reject) => {
```
appilon, line 138: Promises chain from in the inside, you can remove some of the "promisification" here and rewrite it as
```
return this.download(...)
    .then(() => this.verify())
    .then(() => this.unpack())
    .then(() => fs.unlinkSync(destination)
```

You don't need to "catch and reject", promises.then that return a promise will resolve and be forwarded to the next `.then()` on the chain.

Javascript is very subjective which is one of the reasons why I switched to Go :p these are just some suggestions that can help reduce some LOC.

src/languageServerInstaller.ts
```diff
@@ -168,14 +204,41 @@ export class LanguageServerInstaller {
...
+				response.on('data', (data) => {
+					shasumResponse += data;
+				});
+				response.on('end', () => {
+					const shasumLine = shasumResponse.split(`\n`).find(line => line.includes(buildName));
+					if (!shasumLine) {
+						return reject(`Install error: no matching SHA sum for ${buildName}`);
+					}
+					const shasum = shasumLine.split(" ")[0];
+					if (hash.digest('hex') !== shasum) {
```
appilon, line 227: practically speaking this likely always works but I believe `hash.digest()` should not be called until `pkgStream.on('end')` fires. This can be worked into a follow up PR though without going too far down the rabbit hole, but a little hashing promise helper could help

```
function hash(file) {
	return new Promise((resolve, reject) => {
		let hash = crypto.createHash('sha256');
		fs.createReadStream(file)
			.on('data', data => hash.update(data))
			.on('end', () => resolve(hash.digest('hex')))
			.on('error', reject);
	});
}
```
Then you could avoid the "outer promise" and have `verify` be written as
```
function verify(...) {
    return hash(pkg)
        .then(() => /* http request stuff, where you can then safely call hash.digest */)
}
```

src/languageServerInstaller.ts
appilon: it's important to not in my syntax the fat arrow func "one liner" means the invoked method has an implicit `return` in front of it, for multiline logic make sure to include the return
```
return this.download(...)
    .then(() => {
        // do stuff
        // do more stuff
        return this.verify();
    });

File: src/languageServerInstaller.ts (hunk 1 above)
Comment (aeschright) [reply src/languageServerInstaller.ts:122]: It's not an error we need to watch, it'll hit that if there's no previous binary in place (like with a new install).

File: src/languageServerInstaller.ts (hunk 2 above)
Comment (aeschright) [reply src/languageServerInstaller.ts:138]: Thanks, I expanded it out because we were having that weird problem where it wouldn't finish the install, so I wanted to make sure I could see how everything was being caught. I'd prefer to condense this.

File: src/languageServerInstaller.ts [hunk 4]
```diff
@@ -61,18 +64,18 @@ export class LanguageServerInstaller {
 		});
 	}
 
-	checkCurrent(identifier: string) {
+	checkCurrent(userAgent: string) {
 		const releasesUrl = "https://releases.hashicorp.com/terraform-ls/index.json";
```
radeksimko, line 68: Should this be `indexUrl` and reuse the `releasesUrl` from the top?

src/languageServerInstaller.ts
```diff
@@ -168,14 +204,41 @@ export class LanguageServerInstaller {
...
+			https.get(`${releasesUrl}/${release.version}/${release.shasums}`, (response) => {
+				response.on('data', (data) => {
+					shasumResponse += data;
+				});
+				response.on('end', () => {
+					const shasumLine = shasumResponse.split(`\n`).find(line => line.includes(buildName));
+					if (!shasumLine) {
+						return reject(`Install error: no matching SHA sum for ${buildName}`);
+					}
+					const shasum = shasumLine.split(" ")[0];
```
radeksimko, line 226: I'm assuming this is fine, given that you only need the first part after splitting, but FYI the checksum files have 2 spaces between the sum and filename, at least when released by GoReleaser: https://github.com/goreleaser/goreleaser/blob/fd7dbf65922663f912e6561e94aecc7d2a83c3c0/internal/pipe/checksums/checksums.go#L87

aeschright: Fixed it to avoid confusion later.

src/languageServerInstaller.ts
aeschright: This is a good idea that I'll come back to when we add the signature verification.

PR #420, opened 2020-06-26
Readme: Correct supported list of extensions

radeksimko:
Looks like this was just a relic from the older version of the extension. Generally `hcl` represents generic HCL configuration, not something Terraform would ever load.

That said it's something we may consider supporting in the future either directly or via a separate extension and/or separate language server.

For now _not_ supporting HCL in a _Terraform_ extension is intentional.

Relatedly if any other Terraform extension supports HCL and treats it the same way as Terraform's HCL, it's _probably_ inaccurate. Many extensions aren't currently aware of this distinction and that's something we also need to communicate going forward, as per https://github.com/hashicorp/terraform-ls/issues/50#issuecomment-630053301

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
 
 - Manages installation and updates of the [Terraform Language Server (terraform-ls)](https://github.com/hashicorp/terraform-ls), exposing its features:
   - Initialized provider completion (resource names, data source names, attribute names)
-- Includes syntax highlighting for `.tf` and `.tfvars` files (and `.hcl`) -- including all syntax changes new to Terraform 0.12
+- Includes syntax highlighting for `.tf` and `.tfvars` files -- including all syntax changes new to Terraform 0.12
 - Closes braces and quotes
 - Includes `for_each` and `variable` syntax shortcuts (`fore`, `vare`, `varm`)
 


aeschright:
This got caught in an in-between state: the syntax highlighting file will apply itself to .hcl files if the Terraform extension is already loaded. If we want to 100% ignore all .hcl then the file header in `/syntaxes` also needs to be updated.

paultyng:
If you see #225, we mark Terraform content as `hcl` in Markdown for our docs, which complicates this, so we should offer basic HCL syntax highlighting here (and inject to markdown) where we can.

radeksimko:
I just thought it's easier not to support it initially, because if we are going to highlight `hcl` it should use dedicated rules which would be subset of Terraform syntax IMO and as such it e.g. probably shouldn't highlight/recognise references in any way.

I suspect people won't download Terraform extension to get highlighting for generic HCL, but also they might have Terragrunt configs alongside Terraform, which *are* generic HCL and should be highlighted as such.

I understand the pragmatic argument that most of our docs use hcl identifier in markdown and all of these docs do contain Terraform, but also they use wrong identifier, so we will need to address that tech debt sooner or later.

I guess it comes down to comparing the cost of:

1. highlighting Terragrunt configs incorrectly, vs
2. not highlighting existing examples in our docs, vs
3. building a generic HCL highlighting rules

It seems there's perception that (1) has the lowest cost, which I'm happy to respect, but I wanted to make sure we are all making an informed decision 😉

FernandoMiguel:
i load outputs of terraform runs into vscode to read them, and those are HCL highlighted

FernandoMiguel:
and then there's this 
https://github.com/wholroyd/vscode-hcl
which is so old, i assume doesnt support HCL2

fishhead108:
@aeschright is something blocks from merge?

PR #423, opened 2020-06-26
Add setting for workspace root module configuration

### LINKED ISSUE #396: Add ability to explicitly specify root modules as a workspace setting
Users' should be able to manually override root modules for a given Terraform workspace or folder. The config should be scoped `resource`.

```json
"terraform.rootModules": [
  "dir1",
  "dir2/dir3",
  "otherDir"
]
```

The paths should be relative to the root of the folders in the workspace (not sure if we want to allow for initialization outside workspace scope, maybe as a follow-on issue to investigate).

Given the above config and the following tree structure for a single workspace folder:

```tree
.
├── dir1
│   └── main.tf 
├── someDir
│       └── main.tf 
└── dir2
    ├── notRoot
    └── dir3
        └── main.tf 
```

The extension should enumerate all the directories in the setting that match actual directories present and pass those to the LS. In this case it would be:

```
[
  "dir1",
  "dir2/dir3"
]
```

(You'll notice `otherDir` is not passed as it doesn't match anything on disk, and `someDir` is not passed as its not in the settings).

---

Optionally we may also want to support per root module expanded configuration.

```json
"terraform.rootModules": {
  "dir1": {
    "someSetting": ""
  }, 
  "dir2/dir3": {
    "someOtherSetting": ""
  }
}
```

aeschright:
Closes #396


aeschright:
Depends on https://github.com/hashicorp/terraform-ls/issues/189

src/extension.ts
```diff
@@ -58,15 +58,16 @@ export function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(
 		vscode.workspace.onDidChangeConfiguration(
 			(event: vscode.ConfigurationChangeEvent) => {
-				if (!event.affectsConfiguration('terraform.languageServer')) {
+				if (event.affectsConfiguration("terraform")) {
+					const reloadMsg = "Reload VSCode window to apply language server changes";
+					vscode.window.showInformationMessage(reloadMsg, "Reload").then((selected) => {
+						if (selected === "Reload") {
+							vscode.commands.executeCommand("workbench.action.reloadWindow");
```
paultyng, line 65: Why does this need to restart, why can't we just use the configuration RPC hook?

package.json
```diff
@@ -134,6 +134,11 @@
             "maxNumberOfProblems": 100,
             "trace.server": "off"
           }
+        },
+        "terraform.rootModules": {
+          "scope": "resource",
+          "type": "array",
+          "description": "Per-workspace list of module directories for the language server to read"
```
paultyng, line 141: should this have nested proprties, in case we need to support that? not sure if JSON schema can properly convey a string or a property, so to avoid breaking people down the road, maybe we need to go for the nested properties now?

src/extension.ts
paultyng: Just read the LS issue, this is fine if its just the initial implementation and we can improve it over time.

aeschright: @radeksimko and I talked through it this morning, this is the simplest option to start with. We'll have to stagger the language server and extension releases to deploy this, so it makes the most sense to do something we know we can ship on the language server quickly.

package.json
aeschright: By nested you mean the bit under "Optionally we may also want to support per root module expanded configuration." ? I have some concerns about managing this, one of them being that there are eight different layers of configuration scopes that all get rolled up together. The docs note: "Only object value types are merged and all other value types are overridden."

paultyng: The more I think about it we should just deal with that when we get to it and just go with array of strings for now. JSON schema does support `oneOf`, whether VS Code does though I am unsure, and I'd imagine it definitely doesn't support it in the UI.

package.json
```diff
@@ -134,6 +134,11 @@
             "maxNumberOfProblems": 100,
             "trace.server": "off"
           }
+        },
+        "terraform.rootModules": {
+          "scope": "resource",
+          "type": "array",
```
paultyng, line 140: Found this example here: https://github.com/golang/vscode-go/blob/93a25eb84ff1aad58cee3eff9126bb09c857d7ca/package.json#L866-L874
```suggestion
          "type": "array",
          "items": {
            "type": "string"
          },
          "default": []          
```

package.json
```diff
@@ -134,6 +134,15 @@
...
+          "items": {
+            "type": "string"
+          },
+          "default": []
```
paultyng, line 144:
```suggestion
          "default": [],
```

package.json
```diff
@@ -134,6 +134,15 @@
             "maxNumberOfProblems": 100,
             "trace.server": "off"
           }
+        },
+        "terraform.rootModules": {
```
radeksimko, line 138: I guess what we call the option here is orthogonal to how it is sent to the server.

The server doesn't support explicit `workspace/configuration` calls yet so doesn't matter as much, but it will at some point and so it will be asking for an explicit section of a config in the future.

The convention I saw in gopls is that the server asks for sections called `gopls` and `gopls-workspaceName`. I'm not saying we have to care about workspace scoping yet, but I do think we should call the section `terraform-ls` if possible, or somehow separate what's server settings and what's client settings so it's not tempting to mix the two.

package.json
radeksimko: array of strings is currently what the server expects:
https://github.com/hashicorp/terraform-ls/pull/198/files#diff-98ffa6f0726a4bff51793b717c302e0dR13

Changing the type and supporting different types for the same option will probably be difficult, but we can (and probably should, to avoid confusion) deprecate one and introduce another if we need to.

src/extension.ts
radeksimko: The reload will still be necessary with the latest proposed solution (which involves `initializationOptions`), there's some context here to explain why:
https://github.com/hashicorp/terraform-ls/issues/189#issuecomment-652337667

package.json
aeschright: Good point, I think using `terraform-ls` would be best so it's not confusing for people who do choose to use a different language server.

thejosephstevens:
Quick question, will this support wildcards? I have a directory of directories named "something.standard.suffix" and I'd love to just pick them all up with "terraform.rootModules" : "*.standard.suffix". I'd be up for trying to add this myself too if that's out of scope for this PR

aeschright:
> Quick question, will this support wildcards? I have a directory of directories named "something.standard.suffix" and I'd love to just pick them all up with "terraform.rootModules" : "*.standard.suffix". I'd be up for trying to add this myself too if that's out of scope for this PR

Not on the first round, and we're still examining which parts of this should be handled by the extension and which by the language server. If you want to create an issue with your use case, that would be helpful!

PR #428, opened 2020-07-02
Add syntax for object and tuple structural types

### LINKED ISSUE #391: Equals symbol in variable type literals breaks syntax highlighting
Example problem code:

```tf
variable myvar {
  type = object({ key = val })
}
```

[Link to documentation section on Structural Types.](https://www.terraform.io/docs/configuration/types.html#structural-types)

The vscode extension doesn't handle this syntax. All subsequent text is highlighted wrong:

Screen Shot 2020-06-15 at 8 03 11 pm

You can work around this by replacing the equals sign `=` with a colon `:`, which is not well documented but fully supported by the language:

Screen Shot 2020-06-15 at 8 03 25 pm

(`terraform fmt -check test.tf` returns 0 for both snippets.)

aeschright:
Fixes #391

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -84,7 +84,7 @@
 			"end": "\\*/"
 		},
 		"attribute_definition": {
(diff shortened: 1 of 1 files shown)


aeschright:
I haven't run across anything, in general it seemed like tips for implementing new syntaxes were pretty heavy on the "and then you manually hammer on it until something works".

PR #438, opened 2020-07-15
Fix shasum verification

### LINKED ISSUE #437: Install error: SHA sum for terraform-ls_0.5.1_darwin_amd64.zip does not match
This error pops up whenever I open a Terraform code directory in vscode 1.47.0.

radeksimko:
My JavaScript-fu is very rusty, but I believe this fixes #437 which seems to be caused by the local sum calculation done asynchronously and not synchronized with the actual comparison code - at least that's my understanding.

### Before

Screenshot 2020-07-15 at 10 45 40

### After

Screenshot 2020-07-15 at 10 40 41

Fixes #437


src/languageServerInstaller.ts
```diff
@@ -10,6 +10,13 @@ import * as yauzl from 'yauzl';
 
 const releasesUrl = "https://releases.hashicorp.com/terraform-ls";
 
+interface Release {
```
paultyng, line 13: this type is also referenced in the `installPkg` method, you should probably update that as well

radeksimko: Good catch - fixed.

bradennapier:
please merge this asap - this breaks all functionality

PR #444, opened 2020-07-23
Install Refactor

appilon:
This refactor flattens the code as much as possible and fixes an install bug. The conversion did lose one cleanup task (deleting the zip on error), will try and restore that later.


src/extension.ts
```diff
@@ -87,17 +87,18 @@ export function deactivate(): Thenable<void> | undefined {
 async function installThenStart(context: vscode.ExtensionContext, config: vscode.WorkspaceConfiguration) {
 	const command: string = config.get("languageServer.pathToBinary");
 	if (command) { // Skip install/upgrade if user has set custom binary path
-		startLsClient(command, config);
+		await startLsClient(command, config);
 	} else {
 		const installer = new LanguageServerInstaller;
-		const installDir = `${context.extensionPath}/lsp`
-		installer.install(installDir).then(() => {
-			config.update("languageServer.external", true, vscode.ConfigurationTarget.Global);
```
appilon, line 95: @paultyng this change is a bit outside my current understanding of the codebase, just want to double check with you this is what we want?

paultyng: we should move this to the top level commands for enable/disable

appilon: It appears to already be there, could I get approval? We tested this PR together already.

appilon:
@radeksimko the formatting is Alex Pilon flavoured javascript, I like to indent and align chained methods, I find helps keep track of what "layer" you are in once you start putting callbacks everywhere. I will run the vscode formatter but I don't think it will correct that (javascript does not have strict formatting best practices like Go... sadly).

As for the error I have never seen that, what are your repro steps? I will see if I can get that error.

appilon:
@radeksimko I am able to reproduce that error message now

appilon:
@radeksimko fixed!

PR #448, opened 2020-08-04
Preliminary multi folder workspace support

### LINKED ISSUE #434: VS Code Folders should have isolated language server context
I am unsure if this is at all duplicate of #417 since we don't have enough details there, but I managed to reproduce a bug with a great help of a community member [on Discourse](https://discuss.hashicorp.com/t/no-root-module-found/10944/12).

This could also be solved by https://github.com/hashicorp/terraform-ls/issues/179 but I think if we _know_ that two folders just aren't related to each other I think we just shouldn't even attempt to be cross-linking them in any way, i.e. these shouldn't share the same language server session.

--- 

## Hierarchy

```
.
|-- first
    `-- .terraform
`-- second
    `-- .terraform
```

## Steps to reproduce

1. Create a hierarchy of two (initialized) root modules as above
2. Open `first` in VS Code, e.g. via `code first` in the terminal
3. Right click in the Explorer `->` `Add Folder to Workspace`
4. Pick `second` folder

## Actual behaviour

`textDocument/didOpen` was sent to the server, which looked up `first` root module as a candidate for `second`, because they share the same LS session.

## Expected behaviour

New language server session or process is launched for `second` folder and `initialize` request is sent, which triggers an isolated discovery of root module(s) within `second`.

We could also leverage [`workspace/workspaceFolders`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_workspaceFolders), but I am afraid that might require some more significant changes on the server side. Also `workspace/workspaceFolders` is opt-in per LSP and when it isn't supported by the server then most clients (those I had chance to try) just launch isolated session and send new `initialize` request, so I think the extension should just mimic this behaviour for now.

appilon:
This allows multi ls server support for adding/removing folders from the workspace. There may need to be more support for other event hooks (such as open file), for now just focusing on assuming the user is explicitly opening a folder to add to the workspace, this is not tested for terraform modules nested in a top level folder either.

Closes #434


README.md
```diff
@@ -37,7 +37,7 @@ In addition, this new version brings the syntax highlighting up to date with all
 ```json
 {
   "terraform.languageServer": {
-    "enabled": true,
+    "external": true,
```
appilon, line 40: I am a bit unclear on the history here, but the code lead me to believe the old plugin had a setting called `enabled` but our plugin uses a new property called `external`? This README I think was a typo (not reflecting the new property)

aeschright: Yes, that's the case. It probably wasn't too well thought out on my side, but when we were originally prototyping we had both an internal language server option and terraform-ls to switch between. It does give us a flag to look for on old configs though.

src/extension.ts
```diff
@@ -41,110 +40,152 @@ export function activate(context: vscode.ExtensionContext) {
...
-async function startLsClient(cmd: string, config: vscode.WorkspaceConfiguration) {
+function newClient(cmd: string, folder: string) {
 	const binaryName = cmd.split("/").pop();
-	const lsConfig = vscode.workspace.getConfiguration("terraform-ls");
-	const serverArgs: string[] = config.get("languageServer.args");
-	let serverOptions: ServerOptions;
-	let initializationOptions = { rootModulePaths: lsConfig.get("rootModules") };
+	const channelName = `${binaryName}/${folder}`;
+	const serverArgs: string[] = config('terraform').get('languageServer.args');
+	let initializationOptions = { rootModulePaths: config('terraform-ls').get("rootModules") };
```
appilon, line 111: I feel like with the client side now handling folder context (via this PR)... this argument shouldn't be passed along? Locally testing this config is unset for me, so it hasn't caused any bugs.

aeschright: Are you asking if we want to remove the `rootModulePaths` config entirely?

src/extension.ts
```diff
@@ -41,110 +40,152 @@ export function activate(context: vscode.ExtensionContext) {
...
 }
 
-function stopLsClient() {
-	if (!client) {
-		return;
+function config(section: string) {
+	return vscode.workspace.getConfiguration(section);
+}
+
+// this was lifted from a sample, having the folders sorted may not be necessary
```
aeschright, line 179: Probably want to check this and remove the comment before we merge :)

src/extension.ts
appilon: I am unclear on the cases it covers (have not used them myself yet), but maybe?

appilon: I spoke with Paul and understand this now, I load the config with a folder based scope

njuCZ:
@appilon May I ask a question about this feature? the terraform-ls has supported multi-workspace and will automatically scan all root modules, why do we need this feature in vscode client?

appilon:
Hey @njuCZ . `terraform-ls` will discover all modules "downstream" of a single top level folder. The client though is responsible for starting `terraform-ls`, in scenarios where users have multiple top level folders in their workspace (not nested in one another), this feature is needed:

IE
```
/User/alex/terraform_for_work
/User/alex/terraform_for_home
/User/alex/terraform_sandbox
```

njuCZ:
@appilon yes, currently terraform-ls does not support multiple top level workspace, but I think it's a feature that language server should do ? https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize the `InitializeParams` has a field `workspaceFolders?: WorkspaceFolder[] | null;` which I think contains the multiple top level workspace url

paultyng:
@njuCZ for now its simpler to just do an LS per folder, so that configuration scopes can be more easily handled. I think long term that its reasonable to handle it on the LS side if its possible but right now just need to solve these folder issues for users.

radeksimko:
I was OOO for a while, but just to add a bit more context to the LS vs client distinction:

@njuCZ You are absolutely right that LSP does have the concept of workspaces, but that is an _optional_ part of the protocol which also isn't currently implemented by the LS. While we can and probably should support it, we need the client to reflect the optional nature of it and behave in a sensible way when these workspaces aren't explicitly supported. Right now the client would just initialize once and treat all files as part of the same folder, which causes confusion on the server side.

It's my understanding that other servers often deal with this is by just starting new instance for each folder.

PR #454, opened 2020-08-13
Update ls client to major version 6

appilon:
Also updated to latest minor releases of other dependencies


package.json
```diff
@@ -166,26 +166,26 @@
     "test": "npm run test-compile && node ./out/test/runTest.js"
   },
   "dependencies": {
-    "@octokit/rest": "^17.1.3",
-    "@types/semver": "^7.1.0",
+    "@octokit/rest": "^17.11.2",
```
aeschright, line 169: I don't think octokit is still being used

package.json
```diff
@@ -166,26 +166,26 @@
...
+    "@typescript-eslint/parser": "^2.34.0",
     "eslint": "^6.8.0",
     "glob": "^7.1.6",
-    "mocha": "^7.1.1",
-    "typescript": "^3.6.4",
-    "vsce": "^1.77.0",
+    "mocha": "^7.2.0",
+    "typescript": "^3.9.7",
+    "vsce": "^1.78.0",
     "vscode-test": "1.3.0"
```
aeschright, line 189: AFAIK `vscode-test` can also be updated, there's a version 1.4.0 out.

PR #464, opened 2020-08-19
Delete downloaded zips

appilon:
This restores behavior that was lost after #444


package.json
```diff
@@ -175,6 +175,7 @@
   },
   "dependencies": {
     "@types/semver": "^7.3.1",
+    "del": "^5.1.0",
```
paultyng, line 178: 😭

PR #473, opened 2020-08-25
Update README.md

aicarmic:
Updates to readme, aiming to help new users and those running into common issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
 1. Install the extension from the Marketplace
 1. Reload VS Code after the installation (click the reload button next to the extension).
 1. Perform a `terraform init` to provide `terraform-ls` with an up-to-date Provider schema
-1. Open your desired Workspace and/or the root folder containing your Terraform files. Note: see *Known Issues* below about multi-folder workspaces
+1. Open your desired Workspace and/or the root folder containing your Terraform files. Note: see [*Known Issues*](#known-issues) below about multi-folder workspaces
 1. Depending on your settings in VS Code, completion will start automatically (if not inside quotes/string literal, on certain trigger characters), or you can explicitly trigger completion via keyboard combination (Ctrl+Space on Windows, control+space on Mac).
 
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
 1. Install the extension from the Marketplace
 1. Reload VS Code after the installation (click the reload button next to the extension).
 1. Perform a `terraform init` to provide `terraform-ls` with an up-to-date Provider schema
-1. Open your desired Workspace and/or the root folder containing your Terraform files. Note: see [*Known Issues*](#known-issues) below about multi-folder workspaces
+1. Open your desired workspace and/or the root folder containing your Terraform files. Note: see [*Known Issues*](#known-issues) below about multi-folder workspaces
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,17 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
 - Closes braces and quotes
 - Includes `for_each` and `variable` syntax shortcuts (`fore`, `vare`, `varm`)
 
+## Getting Started
+
+**IMPORTANT:** After installing, you must perform a `terraform init` to provide `terraform-ls` with an up-to-date Provider schema. The language-server will not work correctly without first completing this step!
```
radeksimko, line 17:
```suggestion
**IMPORTANT:** After installing, you must perform a `terraform init` to provide `terraform-ls` with an up-to-date provider schemas. The language server will not work correctly without first completing this step!
```
(minor nits) We never capitalize `provider` and technically there can be schemas for more than a single provider.

README.md
```diff
@@ -12,6 +12,17 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
...
+
+1. Install the extension from the Marketplace
+1. Reload VS Code after the installation (click the reload button next to the extension).
```
radeksimko, line 20: Maybe too much context - unsure - but we could link to https://github.com/hashicorp/terraform-ls/issues/222 here - i.e. this step should not be necessary anymore once that issue is resolved

README.md
```diff
@@ -12,6 +12,17 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
...
+1. Perform a `terraform init` to provide `terraform-ls` with an up-to-date Provider schema
```
radeksimko, line 21:
```suggestion
1. Perform a `terraform init` to provide `terraform-ls` with an up-to-date provider schemas
```
AFAIK we never capitalize `provider` anywhere.

README.md
```diff
@@ -12,6 +12,17 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
...
+1. Open your desired Workspace and/or the root folder containing your Terraform files. Note: see [*Known Issues*](#known-issues) below about multi-folder workspaces
```
radeksimko, line 22:
```suggestion
1. Open your desired workspace and/or the root folder containing your Terraform files. Note: see [*Known Issues*](#known-issues) below about multi-folder workspaces
```

I may not be fully familiar with the VS Code naming convention here, but is "workspace" the official name for it? Either way I think it probably shouldn't be capitalized.

README.md
```diff
@@ -12,6 +12,17 @@ The HashiCorp Terraform Visual Studio Code (VS Code) extension adds syntax highl
...
+1. Depending on your settings in VS Code, completion will start automatically (if not inside quotes/string literal, on certain trigger characters), or you can explicitly trigger completion via keyboard combination (Ctrl+Space on Windows, control+space on Mac).
```
radeksimko, line 23: Similar to above we could link to https://github.com/hashicorp/vscode-terraform/issues/373 here, but I'm unsure if the context is helpful.

radeksimko, line 23:
```suggestion
1. Depending on your settings in VS Code, completion will start automatically (if not inside quotes/string literal, on certain trigger characters), or you can explicitly trigger completion via keyboard combination (`Ctrl+Space` by default).
```

radeksimko (commented):
I think we should also mention how to enable the formatting because AFAIK it doesn't get enabled by default which is probably what's tripping up most people, e.g. add

```
    "[terraform]": {
        "editor.formatOnSave": true
    },
```
I'm not sure if there is a GUI equivalent to this.

README.md
aicarmic: I'd say we can leave out for these instructions, and just remove it from the steps once we enhance/add monitoring

aicarmic:
@radeksimko merged in most of your suggestions, thanks for taking a look

radeksimko (approved):

PR #475, opened 2020-08-25
Fix object key syntax highlighting

### LINKED ISSUE #466: `tags = {` breaks syntax highlighting
Not sure if this belongs here or at [hashicorp/terraform-ls](https://github.com/hashicorp/terraform-ls), but the following block appears to break any subsequent stanzas:

```
resource "aws_instance" "test" {
  tags = {
    A = "thisisfine"
    B = "this is not fine"
    C = "neither.is.this"
  }
}

resource "aws_instance" "test2" {
    tags = {}
}

```

## Expected result:

Proper syntax highlighting with `"this is not fine"` and `"niether.is.this"` being highlighted as strings and the next block being highlighted normally.

## Actual Result:

example tf — phishtank-infra 2020-08-20 17-01-03

Removing only `B =` or `C = ` lines still gives a broken block, each of those lines appears to break the parser.  Removing both lines and never having anything but `[a-z0-9]` in the tag value appears to be the only "fix".

aeschright:
Closes #466

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -449,7 +449,7 @@
 					"include": "#comments"
 				},
 				{
-					"match": "\\b((?!null|false|true)[[:alpha:]][[:alnum:]_-]*)\\s*(\\=)\\s*([\\w\"]+)",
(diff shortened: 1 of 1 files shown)


DenWin:
Solves some of the problems, but definetly not all.

image


Here the terraform scripts I used:
- [A.tf.txt](https://github.com/hashicorp/vscode-terraform/files/5126074/A.tf.txt)
- [B.tf.txt](https://github.com/hashicorp/vscode-terraform/files/5126073/B.tf.txt)

Issue #477, opened 2020-08-25
PR #475 does not completely resolve issues #466, #469, #461 or #441


DenWin:
PR #475 does not completely resolve issues #466, #469, #461 or #441

#471 and #472 aren't even touched at all!

![image](https://user-images.githubusercontent.com/13867326/91223266-aae00800-e720-11ea-8f03-6d32cd24971a.png)

Here the code:
```hcl
terraform {
  required_version = "~> 0.13.0"
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "2.24.0"
    }
  }
  assume_role_policy = jsonencode({
    Version       = "2012-10-17",
    Statement     = [
      {
        Effect    = "Allow",
        Principal = {
          Service = "ecs-tasks.amazonaws.com"
        },
        Action    = "sts:AssumeRole"
      }
    ]
  })
}

default_settings = {
    test1 = {
        setting1 = "ami-aaa1e"
        item2 = "x"
        item3 = "${local.tfenv.project_name}-${local.tfenv.environment}"
        item4 = "not_showing_properly"
    }
}
```

DenWin:
After some tinkering in the code I found that this section in terraform.tmGrammar.json seems to be the culprit, though I'm not experienced enough to have complete understanding of potential issues resulting from this fix:

Line 433 and following - to be precise line 435 (replace `"begin": "\\{",` by `"begin": "\\= *\\{",`)
```json
		"objects": {
			"name": "meta.braces.terraform",
			"begin": "\\= *\\{",
			"beginCaptures": {
				"0": {
					"name": "punctuation.section.braces.begin.terraform"
				}
			},
			"end": "\\}",
			"endCaptures": {
				"0": {
					"name": "punctuation.section.braces.end.terraform"
				}
			},
```

aeschright:
The update has been released in v.2.2.0.

PR #478, opened 2020-08-26
Fix additional object key matching issues

### LINKED ISSUE #467: "*/*" breaks syntax highlighting
Plugin version: `2.2.0`

edit: Actually it looks like a duplicate of #429.

Not sure if it's related to #466. Perfectly valid strings break highlighting for the whole file.

It appears `*/*` string crashes Syntax Highlighting because `/*` is treated as a start of a comment.

Here is a valid AWS IAM Policy document:

```
data "aws_iam_policy_document" "example" {
  statement {
    effect = "Allow"

    actions = [
      "s3:GetBucketLocation",
      "s3:ListAllMyBuckets"
    ]

    resources = [
      "arn:aws:s3:::*"
    ]
  }

  statement {
    effect = "Allow"

    actions = [
      "s3:ListBucket",
      "s3:GetObject"
    ]

    resources = [
      "arn:aws:s3:::*",
      "arn:aws:s3:::*/*"
    ]
  }

  statement {
    effect = "Allow"

    actions = [
      "s3:*"
    ]

    resources = [
      aws_s3_bucket.example.arn,
      "${aws_s3_bucket.example.arn}/*"
    ]
  }
}
```
which causes problems:
Screenshot-2020-08-21-153216

aeschright:
Closes #467 and #477

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -449,7 +449,7 @@
 					"include": "#comments"
 				},
 				{
-					"match": "\\b((?!null|false|true)[[:alpha:]][[:alnum:]_-]*)\\s*(\\=)\\s*(\".+\"|[\\w\\._-]+)",
+					"match": "\\b((?!null|false|true)[[:alpha:]][[:alnum:]_-]*)\\s*(\\=)\\s*",
 					"comment": "Literal, named object key",
 					"captures": {
 						"1": {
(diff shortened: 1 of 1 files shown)


DenWin:
The "meta.mapping.key.terraform string.unquoted.terraform" remains incorrectly displayed - e.g. source and version within the azurerm-block:

```hcl
terraform {
  required_version = "~> 0.13.0"
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "2.24.0"
    }
  }
}
```
image

DenWin:
I refer to my comment in #477 (once I change the regex in line 435 the "meta.mapping.key.terraform string.unquoted.terraform" was displayed properly - at least in my little sample.

aeschright:
Hi @DenWin – I've looked and this isn't an error with the regex, but due to the semantic names for different elements. Themes display these in different ways. Here's what it looks like in the one I'm running:

Screen Shot 2020-08-26 at 10 10 15 AM

We're going to stick with this implementation for now. The eventual goal is to move syntax highlighting into the language server, and we'll be able to revisit these decisions then.

PR #483, opened 2020-09-01
Setup integration testing

appilon:
This PR sets up helpers and folder structure for integration testing. The example test, tests document symbols which lands in https://github.com/hashicorp/terraform-ls/pull/265.


radeksimko (commented):
I will let @aeschright review the TypeScript details but these look sensible on first sight to me and I'm 💯 👍  for improving test coverage in general, so thank you for the work!

The only thing I'm worried about is bundling the 30M provider binary in the extension repo. I assume (hope) that it at least doesn't get included in the package distributed through the Marketplace, but even then my experience is that such technical debt is nearly impossible to address later. i.e. git history is immutable and we'll have to live with that decision practically forever.

Would calling `init` as part of the test complicate things a lot?

appilon:
@radeksimko great point, that was included mostly by accident, I will amend the setup to calling init as part of the test

.gitignore
```diff
@@ -4,4 +4,5 @@ node_modules/
 .DS_Store
 .vscode-test
 npm-debug.log
-lsp/
\ No newline at end of file
+lsp/
+.terraform/
```
aeschright, line 8: If this is a test output directory, also put it in the `.vscodeignore` file.

src/languageServerInstaller.ts
```diff
@@ -92,26 +91,24 @@ export class LanguageServerInstaller {
...
 		}
-		let arch = os.arch();
-		switch (arch) {
+		let arch: string;
+		switch (process.arch) {
 			case 'x64':
 				arch = 'amd64'
 				break;
-			case 'x32':
+			case 'ia32':
```
appilon, line 106: 32bit installs were not working

aeschright (approved):
🚀

PR #495, opened 2020-10-09
Add command and statusbar interface for running terraform init

aeschright:


src/extension.ts
```diff
@@ -168,6 +195,34 @@ async function stopClients(folders = folderNames()) {
 	return Promise.all(promises);
 }
 
+// Command interface
+
+function sendRequest(clientName: string, command: string, args?: any) {
```
appilon, line 200: early feedback: arguments is a list of any `any[]`. I was keeping my backend simple and just assuming a flat/simple structure similar to a cli invokation

`rootmodule fileInFocus` which would be an ExecuteCommandParams of 
```
{
    command: "rootmodule",
    arguments: ["theFileURIInFocus"]
}
```
I am unsure of if we can have our arguments be complex like
```
{
    command: "rootmodule",
    arguments: [
        {
            file: "fileURIInFocus"
        }
    ]
}
```
Let me know if you prefer a complex request object or keeping it "flat". I imagine some commands such as a future `validate` we will be keeping flat/parity with the actual terraform cli

appilon: There is certainly a case to be made for the complex structure as we can define types on the Golang side to marshal/unmarshal

src/extension.ts
```diff
@@ -194,20 +206,95 @@ function config(section: string, scope?: vscode.ConfigurationScope) {
...
+	needsInit: boolean
+}
+
+async function rootModulesCommand(client: LanguageClient, documentUri: string): Promise<any> {
+	const requestParams: ExecuteCommandParams = { command: "rootmodules", arguments: [`uri=${documentUri}`] };
+	return execWorkspaceCommand(client, requestParams);
+}
+
+async function rootModules(client: LanguageClient, documentUri: string) {
+	return new Promise<rootModuleResponse>((resolve, reject) => {
```
appilon, line 279: I would like to avoid bringing in `async.js`. Perhaps rewrite function to include retry loop, async/await should be it relatively simple

```javascript
async function rootModules(client: LanguageClient, documentUri: string) {
    let done = false;
    for (let i = 0; i < 2 && !done; i++) {
        const resp = await rootModulesCommand(client, documentUri);
        await sleep(100); // move that sleep helper from the tests to somewhere more common and import it
        done = resp.doneLoading;
    }
    if (!done) {
        throw new Error(`Unable to load root modules for ${documentUri}`);
    }
    return { rootModules: resp.rootModules, needsInit: resp.rootModules.length === 0 };
}
```

No need to `try/catch` even, any error in `rootModulesCommand` will automatically bubble up (one of the best parts of `async/await`)

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "module": "commonjs",
-        "target": "es6",
+        "target": "es2017",
```
appilon, line 4: async.js probably did this, but like to just stick with es6

appilon (changes requested):
Early feedback, overall looking good, just want to avoid dependency.

src/extension.ts
aeschright: That was what I started with, but it felt clunky (not necessarily a problem!). Is there a specific reason you want to avoid adding the dependency, or just to keep it simple?

aeschright:
I updated everything to use the merged workspace pruning code, and add a uuid to manage the client commands.

package.json
```diff
@@ -175,10 +175,12 @@
     "test": "npm run test-compile && node ./out/test/runTest.js"
   },
   "dependencies": {
+    "@types/async": "^3.2.3",
```
appilon, line 178: you should be able to prune this dep now?

src/extension.ts
```diff
@@ -102,10 +147,11 @@ function newClient(cmd: string, location: string) {
 	}
 	let initializationOptions = {};
 	if (rootModulePaths.length > 0) {
-		initializationOptions = { rootModulePaths };
-	}
-	if (excludeModulePaths.length > 0) {
-		initializationOptions = { excludeModulePaths };
+		initializationOptions = { uuid, rootModulePaths };
```
appilon, line 150: super nitpick, could we have the parameter on the object be just `id`, just trying to avoid enforcing the protocol to have to worry about using a specific uuid library implementation, I'm perfectly happy with `uuid` being used, but would like the variable in `initializationOptions` to imply it just needs to be any unique identifier relative to the client.

src/extension.ts
```diff
@@ -102,10 +147,11 @@ function newClient(cmd: string, location: string) {
...
+	} else if (excludeModulePaths.length > 0) {
+		initializationOptions = { uuid, excludeModulePaths };
```
appilon, line 152: if variable name is updated, would apply here too

src/extension.ts
```diff
@@ -102,10 +147,11 @@ function newClient(cmd: string, location: string) {
...
+	} else {
+		initializationOptions = { uuid }
```
appilon, line 154: and here

appilon (changes requested):
Super small comments, overall though looking good and essentially ready for merge, will get the backend work merged for you soon for testing

src/extension.ts
appilon: So not asking the implementation or use of a `uuid` variable name else changes, just at the protocol layer in this if else if else

appilon (dismissed):

src/extension.ts
radeksimko: As mentioned in my other review here https://github.com/hashicorp/terraform-ls/pull/279 I would probably focus more on what this is used for than how it is generated, and call this "command prefix"?

appilon:
Just confirming, we need nail in the protocol to call it `commandPrefix` its niche usage has been documented on the LS settings doc. the commands will now look like
```
commandPrefix.terraform-ls.commandName
```

Sorry for the constant churn, but it's now merged for testing in the langserver

src/extension.ts
```diff
@@ -174,6 +217,52 @@ async function pathToBinary(): Promise<string> {
...
+	const requestParams: ExecuteCommandParams = { command: `${languageClient.commandPrefix}.terraform-ls.rootmodules`, arguments: [`uri=${documentUri}`] };
+	return execWorkspaceCommand(languageClient.client, requestParams);
+}
+
+async function rootModules(languageClient: terraformLanguageClient, documentUri: string): Promise<rootModuleResponse> {
+	let doneLoading = false;
+	let rootModules: rootModule[];
+	for (let attempt = 0; attempt < 2 && !doneLoading; attempt++) {
+		const response = await rootModulesCommand(languageClient, documentUri);
+		await sleep(100);
```
appilon, line 256: I would move sleep  to only if needed
```
const response = await rootModulesCommand(languageClient, documentUri);
doneLoading = response.doneLoading;
rootModules = response.rootModules;
if (!doneLoading) {
  await sleep(100);
}

File: src/test/helper.ts [hunk 9]
```diff
@@ -41,9 +42,7 @@ async function activated() {
 	return _activatedPromise;
 }
 
-export async function sleep(ms: number): Promise<void> {
-	return new Promise(resolve => setTimeout(resolve, ms));
-}
+export const testFolderPath =path.resolve(__dirname, '../../testFixture');
```
appilon, line 45: minor fmt: missing space after `=`

appilon (dismissed):
Very minor feedback, otherwise LGTM after those changes.

appilon:
The failed test seems like a path separator bug on windows, investigating

src/extension.ts
```diff
@@ -43,6 +59,37 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
...
+		vscode.commands.registerCommand('terraform.init', async () => {
+			const selected = await vscode.window.showOpenDialog({
+				canSelectFiles: false,
+				canSelectFolders: true,
+				canSelectMany: false,
+				defaultUri: vscode.workspace.workspaceFolders[0].uri,
+				openLabel: "Initialize"
+
+			});
+			if (selected) {
```
appilon, line 71: nitpick, can `selected` be nil, but `showOpenDialog` didn't throw?

src/extension.ts
```diff
@@ -84,27 +156,28 @@ async function startClients(folders = prunedFolderNames()) {
 	const disposables: vscode.Disposable[] = [];
 	for (const folder of folders) {
 		if (!clients.has(folder)) {
-			const client = newClient(command, folder);
+			const commandPrefix = shortUid.seq();
+			const client = newClient(command, folder, commandPrefix);
 			disposables.push(client.start());
-			clients.set(folder, client);
+			clients.set(folder, { commandPrefix: commandPrefix, client });
```
appilon, line 162:
```suggestion
			clients.set(folder, { commandPrefix, client });
```

src/extension.ts
```diff
@@ -84,27 +156,28 @@ async function startClients(folders = prunedFolderNames()) {
...
 	const f: vscode.WorkspaceFolder = getWorkspaceFolder(location);
 	const serverArgs: string[] = config('terraform').get('languageServer.args');
 	const rootModulePaths: string[] = config('terraform-ls', f).get('rootModules');
 	const excludeModulePaths: string[] = config('terraform-ls', f).get('excludeRootModules');
 	if (rootModulePaths.length > 0 && excludeModulePaths.length > 0) {
-		throw new Error('Only one of rootModules and excludeRootModules can be set at the same time, please remove the conflicting config and reload'); 
+		throw new Error('Only one of rootModules and excludeRootModules can be set at the same time, please remove the conflicting config and reload');
 	}
-	let initializationOptions = { commandPrefix: uid.seq() };
+	let initializationOptions = { commandPrefix: commandPrefix };
```
appilon, line 180:
```suggestion
	let initializationOptions = { commandPrefix };
```

src/vscodeUtils.ts
```diff
@@ -4,6 +4,18 @@ export function config(section: string, scope?: vscode.ConfigurationScope): vsco
...
 }
 
+export function getFolderName(folder: vscode.WorkspaceFolder): string {
+	return normalizeFolderName(folder.uri.toString());
+}
+
+// Make sure that folder uris always end with a slash
+export function normalizeFolderName(folderName: string): string {
+	if (folderName.charAt(folderName.length - 1) !== '/') {
+		folderName = folderName + '/';
```
appilon, line 14: this maaay be the reason for the test failure, although this should be fine because the client should always work in URIs and thus `/`

appilon (changes requested):
Very minor feedback, mostly ready to test/ship!! Just need to chase down the failure, make sure its not a bug in the client/LS, it may just be a bug with the test setup.

src/extension.ts
aeschright: Yes, if the user hits cancel on the file finder, it's not an error there's just no value.

appilon (approved):
LGTM, just try and condense or squash the commits

PR #499, opened 2020-10-15
Prune workspaces

aeschright:
Remove nested workspace folders before starting clients so that multiple language server instances aren't running for the same folder.


appilon (commented):
Looking good, I think a test case could be written

1) export clients on the extension API surface (the object returned from activate).
2) test add two folders to workspace, one of which is nested inside another
3) verify that only 1 client exists

src/extension.ts
```diff
@@ -194,21 +175,46 @@ function config(section: string, scope?: vscode.ConfigurationScope) {
...
+			});
 	}
-	return folders.map(folder => {
-		let result = folder.uri.toString();
-		if (result.charAt(result.length - 1) !== '/') {
-			result = result + '/';
+	return [];
+}
+
+function folderNames(folders: readonly vscode.WorkspaceFolder[] = vscode.workspace.workspaceFolders): string[] {
```
appilon, line 201: nitpick: I would rename this to `prunedFolderNames` since there is logic around pruning children, otherwise it might get used in a situation later where pruning is unwanted

src/test/workspaces.test.ts
```diff
@@ -0,0 +1,22 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import { getDocUri, open } from './helper';
+
+suite('workspace folders', () => {
```
appilon, line 5: Sadly, vscode does not support `workspace.updateWorkspaceFolders` in tests, only caught a glimpse of an error dialog

Screen Shot 2020-10-21 at 3 37 26 PM

Worse, this test asserts that the result of the call is true (when it wasn't.. sort of) and that there is now 2 workspace folders. Manually testing seemed to reveal that the important event `onDidChangeWorkspaceFolders` doesn't fire? Therefore test passes with `clients.size == 1` but for the wrong reasons, if you wrote a test with 2 independant folders, where we expect `clients.size == 2` I suspect it would fail.

I would leave the testfile in the PR, but comment out the test with a summary, maybe one day we can re-enable it. Sorry for harping on this and then having it be unsupported 😞

src/extension.ts
appilon: As an alternative to the test from below, I would move this helper and it's dependant helpers to the `utils` file, and add a unit test validating that `["a/", "a/b/"]` gets pruned to `["a/"]`, that is about all we can do.

appilon (changes requested):
LGTM, but small reshuffle due to vscode bug/limitations

appilon (approved):

PR #512, opened 2020-11-12
Read LS version from stdout

### LINKED ISSUE #418: Read LS version from stdout+stderr
As per https://github.com/hashicorp/terraform-ls/issues/185 printing out version output to stderr was an unfortunate oversight.

https://github.com/hashicorp/vscode-terraform/blob/1c882b35bb52e834214fad186c167f8ecaf04804/src/languageServerInstaller.ts#L32

I'm not sure of all the implications for old versions lying around, but I think we could at least start reading from both stdout and stderr, with `stdout` having priority (if it's not empty)?

radeksimko:
Closes #418

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -52,11 +52,12 @@ interface Release {
 export class LanguageServerInstaller {
 	public async install(directory: string): Promise<void> {
 		const { version: extensionVersion } = require('../package.json');
-		const lspCmd = `${directory}/terraform-ls --version`;
+		const lsVersionCmd = `${directory}/terraform-ls --version`;
 		const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`;
 		let isInstalled = true;
 		try {
-			var { stderr: installedVersion } = await exec(lspCmd);
+			var { stdout, stderr } = await exec(lsVersionCmd);
+			var installedVersion = stdout || stderr;
 		} catch (err) {
 			// TODO: verify error was in fact binary not found
 			isInstalled = false;


src/languageServerInstaller.ts
```diff
@@ -52,11 +52,12 @@ interface Release {
 export class LanguageServerInstaller {
 	public async install(directory: string): Promise<void> {
 		const { version: extensionVersion } = require('../package.json');
-		const lspCmd = `${directory}/terraform-ls --version`;
+		const lsVersionCmd = `${directory}/terraform-ls --version`;
 		const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`;
 		let isInstalled = true;
 		try {
-			var { stderr: installedVersion } = await exec(lspCmd);
+			var { stdout: stdout, stderr: stderr } = await exec(lsVersionCmd);
```
appilon, line 59: object destructuring only need the `name: othername` if you want to pull out a key, but rename it to `othername` as I did before. if you keep the keyname intact you can simplify to
```
var { stdout, stderr } = await exec(lsVersionCmd);
```

src/languageServerInstaller.ts
```diff
@@ -52,11 +52,12 @@ interface Release {
...
+			var installedVersion = stdout != "" ? stdout : stderr;
```
appilon, line 60: Can be rewritten as
```
var installedVersions = stdout || stderr;
```

appilon (changes requested):

src/languageServerInstaller.ts
radeksimko: I see, let me fix that 😄

appilon (approved):
LGTM, but please hold until latest release goes out

PR #516, opened 2020-11-16
Switch installer to use js-releases for downloading the language server

aeschright:


appilon (approved):
The gh action tests passing should validate this is all working since it downloads the language server

PR #523, opened 2020-12-02
Prepare for semantic token based highlighting

radeksimko:
## Dev Environment

I pinned the version to the latest LTS via NVM as instructed.

## Dependency Changes

Bumping `vscode-languageclient` to 7.0.0 enables the client to speak LSP 3.16 (version which introduces semantic tokens), which is still "upcoming" and so that new part of the protocol may still change until it's finally stabilized.

FWIW the official Go VS Code extension also uses 7.0.0 series: https://github.com/golang/vscode-go/blob/master/package.json#L60 and so do some other popular extensions.

I picked the same patch version as the Go extension uses, which [raises the VSCode requirement to `^1.46.0`](https://github.com/microsoft/vscode-languageserver-node/blob/875bf6c1a4313a5245691c063607e5f031e35308/client/src/node/main.ts#L25). We could in theory bump to any later version, but that also means cutting off any VS Code versions older than `1.51`.

The Go extension actually **doesn't** _formally_ require `1.46` probably because gopls isn't enabled by default and the requirement is imposed by the language client which doesn't engage until LS is enabled. Our extension relies on the language server in many more ways and suggests installation by default, so I'm more tempted to require 1.46 for the extension on Marketplace level, but not doing so would just show error popup mentioning the incompatibility, so it's probably not a big deal.

https://github.com/hashicorp/vscode-terraform/blob/f87955055de8640a6c0506a7751d9ac7b74f62ec/package.json#L9-L11

All changes to LSP are additional and hence any existing 3.15 and earlier functionality should not be affected.

## Grammar Change

The block type token association is being changed to reflect that semantic tokens use custom scoping (with some fallbacks) and the existing scope `storage.type` isn't available there so language server would be unable to match it which would lead to confusing UX where some block types are coloured differently than others. Here I am therefore changing to one that *is* available (`entity.name.type`), which means we can match the style.

### Before

Screenshot 2020-12-07 at 09 59 00 Screenshot 2020-12-07 at 09 59 21

### After

Screenshot 2020-12-07 at 11 59 14 Screenshot 2020-12-07 at 11 59 30

--- 

## LS PR

The language server side is still WIP, but can be tried out by compiling https://github.com/hashicorp/terraform-ls/pull/331


package-lock.json
```diff
@@ -1,8 +1,2783 @@
 {
   "name": "terraform",
   "version": "2.3.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
```
appilon, line 7: weird... empty string key?

appilon:
@radeksimko can you rerun the install commands but with node LTS, you can version switch with [nvm](https://github.com/nvm-sh/nvm)

```
➜ nvm install --lts
Installing latest LTS version.
v14.15.1 is already installed.
Now using node v14.15.1 (npm v6.14.8)
```

radeksimko:
@appilon Done, PTAL.

appilon (changes requested):
I believe the engines entry key needs a bump to 1.46 also now?

appilon (approved):
My preference is for the types and runtime/marketplace filter to remain consistent, although I do understand in our situation it is technically possible to leave the runtime lower (1.42). We should let @aeschright chime in so we're all on the same page.

aeschright:
> My preference is for the types and runtime/marketplace filter to remain consistent, although I do understand in our situation it is technically possible to leave the runtime lower (1.42). We should let @aeschright chime in so we're all on the same page.

I agree, and given that 1.46 was released in May I think it's safe to assume that anyone using our extension has upgraded at least to that point.

PR #526, opened 2020-12-08
Test Github Action setup-terraform

aeschright:


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v2
       - name: Install Terraform
-        uses: hashicorp/setup-terraform@v1
+        uses: hashicorp/setup-terraform@v1.3.1
```
radeksimko, line 19: I'm not sure if this is GitHub Actions, or the way we release our Action, but I would expect `@v1` to just pull the latest `v1` version. Is there a reason why it doesn't work that way? Could we make it work that way?

radeksimko (dismissed):

.github/workflows/test.yml
aeschright: Sorry, I should have added a note that you don't need to review this. I'm just testing a problem in `setup-terraform` so I need the specific version. I'll move the `v1` tag once it's fixed.

PR #527, opened 2020-12-09
use the amd64 binary on Apple Silicon Macs for now

### LINKED ISSUE #521: "Unable to install terraform-ls" macOS arm64
I am getting an error installing terraform-ls. 

OS: macOS 11 arm64 (Apple Silicon)
vscode: 1.52.0-exploration
commit: 88c7381b77df9601914ca914f80c700fee35825a
Electron: 11.0.2
Chrome: 87.0.4280.67
Node.js: 12.18.3
V8: 8.7.220.25-electron.0
OS: Darwin arm64 20.1.0

This may be related to #507 but there is no error message in the Extension Host Log

codekitchen:
This fixes #521

I'm guessing there won't be a native build of terraform-ls until at least February 2021 when Golang 1.16 is released with `GOARCH=arm64` support. So it seems worth doing this for now, the amd64 terraform-ls binary works great.

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -64,6 +64,13 @@ export class LanguageServerInstaller {
 			case 'ia32':
 				arch = '386'
 				break;
+			case 'arm64':
+				if (platform === 'darwin') {
+					// On Apple Silicon, install the amd64 version and rely on Rosetta2
+					// until a native build is available.
+					arch = 'amd64'
+				}
+				break;
 		}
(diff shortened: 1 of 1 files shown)


aeschright (approved):
Thanks for submitting this! You're correct that we won't be able to build a binary for this platform until the tooling catches up.

radeksimko:
Just FYI there is a relevant issue on LS side tracking arm64 support: https://github.com/hashicorp/terraform-ls/issues/342

PR #530, opened 2020-12-15
Add more os platforms to test matrix

aeschright:


appilon (approved):
AWESOME! Please squash at/before merge.

PR #532, opened 2021-01-04
Extract Go/NodeJS OS/arch translation logic

radeksimko:
This will allow installation of solaris builds and `linux/arm64` when it's published: https://github.com/hashicorp/terraform-ls/pull/351

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -52,32 +52,14 @@ export class LanguageServerInstaller {
 		const destination = `${installDir}/terraform-ls_v${release.version}.zip`;
 		fs.mkdirSync(installDir, { recursive: true }); // create install directory if missing
 	
-		let platform = process.platform.toString();
-		if (platform === 'win32') {
-			platform = 'windows';
-		}
-		let arch: string;
-		switch (process.arch) {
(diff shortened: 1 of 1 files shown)


src/languageServerInstaller.ts
```diff
@@ -96,8 +78,36 @@ export class LanguageServerInstaller {
 		});
 	}
 
-	removeOldBinary(directory: string, platform: string): void {
-		if (platform === "windows") {
+	goOs(): string {
```
appilon, line 81: I would probably not add to the Class definition and just call the functions directly. Enforcing style in JS is much more subjective so I tend to go with the flow, but in general I almost never use Classes and opt for more "pure functions and passed objects". So my subjective bias is to lean away from the class use.

appilon (approved):
I left my 0.02$ but nothing explicitly wrong, LGTM

src/languageServerInstaller.ts
radeksimko: Fair point. 👍🏻  I will move it into functions.
Funny enough I did think about that possibility, but I "went with the flow" as you say 😄

appilon (approved):

PR #536, opened 2021-01-05
Add setting for experimental features

### LINKED ISSUE #534: Expose Experimental Features Setting
The language server now has support for experimental features users can opt into. The first of which is validation of a folder on file save.

https://github.com/hashicorp/terraform-ls/pull/340

We need to expose a setting on the `terraform-ls` setting called `experimentalFeatures` and within that allow setting `validateOnSave = true`. Then we must pass the `experimentalFeatures` object to the language server within the `initializationOptions` object

aeschright:
- Boolean: validateOnSave

Closes #534

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,19 @@
           },
           "default": [],
           "description": "Per-workspace list of module directories for the language server to exclude"
-        }
+        },
+        "terraform-ls.experimentalFeatures": {
+          "scope": "resource",
+          "type": "object",
+          "description": "Experimental (opt-in) terraform-ls features",
+          "properties": {
+            "validateOnSave": {
+              "scope": "resource",
+              "type": "boolean",
+              "default": false
+            }
+          }
+        }  
       }
     },
     "commands": [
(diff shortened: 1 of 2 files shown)


aeschright:
Here's what it looks like:
Screen Shot 2021-01-05 at 3 56 14 PM
Screen Shot 2021-01-05 at 3 56 37 PM
Screen Shot 2021-01-05 at 3 56 49 PM

package.json
```diff
@@ -152,7 +152,19 @@
           },
           "default": [],
           "description": "Per-workspace list of module directories for the language server to exclude"
-        }
+        },
+        "terraform-ls.experimentalFeatures": {
+          "scope": "resource",
```
appilon, line 157: Looks great overall, something I don't know well enough yet is the scopes, it's documented here

https://code.visualstudio.com/api/references/contribution-points

is `resource` the one we want?

appilon (approved):
Overall it LGTM but can we double check how scopes work?

appilon:
I can confirm though the setting works!

aeschright:
`window` would be the other option that could fit, but `resource` seems the most flexible so that's what I went with. It matches how we've done the rootModule config as well.

appilon (approved):

PR #540, opened 2021-01-11
Add terraform validate command

### LINKED ISSUE #538: Implement Validate Command Palette
With #536 getting merged, we can also implement validation triggered via command pallete. The workspace command is
```
terraform.validate
```
and takes an argument to specify the folder to run from `uri=<folder>`

aeschright:
Closes #538

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -292,7 +292,7 @@ async function rootModules(languageClient: terraformLanguageClient, documentUri:
 	return { rootModules: rootModules, needsInit: rootModules.length === 0 };
 }
 
-async function terraformCommand(command: string) {
+async function terraformCommand(command: string): Promise<any> {
 	if (vscode.window.activeTextEditor) {
 		const documentUri = vscode.window.activeTextEditor.document.uri;
 		const languageClient = getDocumentClient(documentUri);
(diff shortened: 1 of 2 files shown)


aeschright:
The text of the warning message could be better, if you have suggestions.

src/extension.ts
```diff
@@ -302,9 +292,25 @@ async function rootModules(languageClient: terraformLanguageClient, documentUri:
...
+			const selected = await vscode.window.showQuickPick(modules.rootModules.map(m => m.uri), { canPickMany: false });
+			selectedModule = selected[0];
+		} else {
+			selectedModule = modules.rootModules[0].uri;
+		}
+		const requestParams: ExecuteCommandParams = { command: `${languageClient.commandPrefix}.terraform-ls.terraform.${command}`, arguments: [`uri=${selectedModule}`] };
+		return execWorkspaceCommand(languageClient.client, requestParams);
+	} else {
+		vscode.window.showWarningMessage(`Open a module then run terraform ${command} again`);
+		return Promise.resolve();
```
appilon, line 312: `async` keyword means you can just code like a normal func and return your value normally, it will already be wrapped in a promise.resolve. so you can just `return;`

appilon (approved):
lgtm, small nitpick

src/extension.ts
aeschright: Good catch, thanks.

appilon (approved):

PR #543, opened 2021-01-18
Add for expression snippets

reegnz:
Add snippets to for expressions for both maps and lists.

### DIFF
--- a/snippets/terraform.json
+++ b/snippets/terraform.json
@@ -61,16 +61,16 @@
 	"For Expression (List)": {
 		"prefix": "for_list",
 		"body": [
-			"for ${3:item} in ${2:list} :",
-			"${4:item}"
+			"for ${1:item} in ${2:list} :",
+			"${3}"
 		],
 		"description": "Construct a list or map by transforming and filtering elements in another list."
 	},
 	"For Expression (Map)": {
 		"prefix": "for_map",
 		"body": [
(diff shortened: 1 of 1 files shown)


reegnz:
Example on how the snippets work:

for_list
for_map

NoseyNick (approved):

apetalik-cloudera (approved):
LGTM.

aeschright:
Thanks for submitting this! I'd like to keep the tab order on all of the snippets consistent, so it remains in the same order as when someone would be typing out the full text. If you can update that, I'll be happy to approve the PR.

reegnz:
Sure, will fix that it next week.

reegnz:
@aeschright Sorry for the delay, coompletely forgot about this PR. Fixed the tab ordering.

radeksimko:
Hi @reegnz 
Sorry for the delay in reviewing your PR.

Our longer-term plan is to actually remove the existing snippets from the extension and handle this functionality within the language server. The server can provide more contextual completion - e.g. avoid inserting `for ... in ...` to an empty file where it isn't valid, but only to an argument value after `=`.

This functionality is tracked under https://github.com/hashicorp/terraform-ls/issues/527 and is on our near-term roadmap, alongside other expression types as outlined in https://github.com/hashicorp/vscode-terraform/issues/641

With that in mind I'm going to close this PR.

Thank you for your contribution and understanding.

PR #554, opened 2021-02-08
Update js-releases dep to fix build on GitHub Actions

aeschright:
Ref: https://github.com/actions/setup-node/issues/214


radeksimko (approved):
Sad situation, but understandable fix/workaround 😭

PR #557, opened 2021-02-11
Add telemetry for error reporting

aeschright:


radeksimko (approved):
LGTM in principle.

We discussed the default fields and agreed to test this internally to double check there's nothing sensitive being collected - which there shouldn't be (according to the package docs).

I am still curious though why does it not automatically generate the UUIDs here as it probably should, according to the docs:

Screenshot 2021-02-12 at 10 44 56

Maybe VS Code just doesn't do that in the dev mode? 🤔  🤷🏻‍♂️

PR #560, opened 2021-02-16
Use version JSON output of LS during installation

radeksimko:
This change makes the extension leverage a new flag introduced in LS 0.13.0 via https://github.com/hashicorp/terraform-ls/pull/386 so that we can merge and release https://github.com/hashicorp/terraform-ls/pull/407

The usual disclaimer of my TS/JS skills applies here - i.e. any criticism is welcomed 😄

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -15,10 +15,10 @@ export class LanguageServerInstaller {
 		let installedVersion: string;
 		try {
 			installedVersion = await getLsVersion(directory);
-			console.log(`Found installed LS ${installedVersion}`);
-			isInstalled = true
+			isInstalled = true;
 		} catch (err) {
-			console.log(`failed to get version: ${err}`);
+			// TODO: verify error was in fact binary not found
+			isInstalled = false;
 		}
 
 		const currentRelease = await getRelease("terraform-ls", "latest", userAgent);
@@ -98,15 +98,17 @@ async function getLsVersion(dirPath: string): Promise<string> {
 	fs.accessSync(lsBinPath, fs.constants.X_OK);
 
 	try {
-		var { stdout } = await exec(`${lsBinPath} version -json`);
-		let jsonOutput = JSON.parse(stdout);
+		let jsonCmd: { stdout: string };
+		jsonCmd = await exec(`${lsBinPath} version -json`);
+		let jsonOutput = JSON.parse(jsonCmd.stdout);
 		return jsonOutput.version
 	} catch (err) {
 		// assume older version of LS which didn't have json flag
 		if (err.status != 0) {
 			try {
-				var { stdout, stderr } = await exec(`${lsBinPath} -version`);
-				let version = stdout || stderr
+				let plainCmd: { stdout: string, stderr: string };
+				plainCmd = await exec(`${lsBinPath} -version`);
+				let version = plainCmd.stdout || plainCmd.stderr
 				return version
 			} catch (err) {
 				throw err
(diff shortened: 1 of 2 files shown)


src/languageServerInstaller.ts
```diff
@@ -9,16 +9,16 @@ import { Release, getRelease } from '@hashicorp/js-releases';
...
 	public async install(directory: string): Promise<void> {
 		const { version: extensionVersion } = require('../package.json');
-		const lsVersionCmd = `${directory}/terraform-ls --version`;
 		const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`;
-		let isInstalled = true;
+		let isInstalled = false;
 		try {
-			var { stdout, stderr } = await exec(lsVersionCmd);
-			var installedVersion = stdout || stderr;
+			var installedVersion = await getLsVersion(directory);
```
aeschright, line 15:
```suggestion
			const installedVersion = await getLsVersion(directory);
```

src/languageServerInstaller.ts
```diff
@@ -9,16 +9,16 @@ import { Release, getRelease } from '@hashicorp/js-releases';
...
+			console.log(`Found installed LS ${installedVersion}`);
```
aeschright, line 16: The console log and error functions don't get seen by the user, so we should either throw the error to make sure it's handled, or use some other kind of output.

src/languageServerInstaller.ts
```diff
@@ -9,16 +9,16 @@ import { Release, getRelease } from '@hashicorp/js-releases';
...
+			isInstalled = true
 		} catch (err) {
-			// TODO: verify error was in fact binary not found
-			isInstalled = false;
+			console.log(`failed to get version: ${err}`);
```
aeschright, line 19: Same as above – also we still aren't confirming that the error is because the language server isn't installed, so the comment could stay.

src/languageServerInstaller.ts
```diff
@@ -91,6 +91,31 @@ export class LanguageServerInstaller {
 	}
 }
 
+async function getLsVersion(dirPath: string): Promise<string> {
+	const lsBinPath = `${dirPath}/terraform-ls`;
+
+	fs.accessSync(lsBinPath, fs.constants.X_OK);
+
+	try {
+		var { stdout } = await exec(`${lsBinPath} version -json`);
```
aeschright, line 100: Switch `var` to let and move the variable declaration outside the block.

aeschright (changes requested):

src/languageServerInstaller.ts
radeksimko: I think in this particular case it was mostly intentional to keep it silent as from UI perspective finding LS version is an implementation detail, but I thought it might be helpful for debugging later?

src/languageServerInstaller.ts
radeksimko: True, but AFAIK this codepath is hit regularly on a fresh installation (_including_ when the binary is not found), so it's probably not something we'd want to notify the user about? I guess the best solution we'd be looking for eventually is to separate out upgrading and fresh installation and then treat this condition differently in each case?

aeschright: Yes, a new install is the most likely scenario, so we don't need to throw the error in this case. My preference is just to not leave in any logging that the user can't access.

radeksimko: For some reason I thought `console.log` would appear in one of the Output panes, but it looks like I'm wrong. 😄 

Makes sense to leave it out then 👌🏻

aeschright (approved):
Looks good!

PR #563, opened 2021-02-19
[HOLD] Install arm64 instead of amd64 for darwin

radeksimko:
This should not be merged until https://github.com/hashicorp/terraform-ls/pull/350 is merged and released, i.e. when there is an actual darwin/arm64 build of LS available.

That PR is currently also on hold due to https://github.com/syndbg/goenv/pull/164

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -139,10 +139,6 @@ function goArch(): string {
 	if (arch === 'x64') {
 		return 'amd64';
 	}
-	if (arch === 'arm64' && process.platform.toString() === 'darwin') {
-		// On Apple Silicon, install the amd64 version and rely on Rosetta2
-		// until a native build is available.
-		return 'amd64';
-	}
+
 	return arch;
 }


pszypowicz:
Hi, just a kind reminder, that's all of your requirements seems to be meet :)

radeksimko:
@pszypowicz Thanks for the reminder.

Yes, the conditions I mentioned are met. We just had an internal conversation about how the unavailability of darwin/arm64 build of Terraform may affect the experience. There is certainly plan to release it soon, but not before significant number of official providers are released for that platform.

I'm sure that this could be worked around by passing `-tf-exec` flag to ensure that Terraform runs via Rosetta, but I don't think that's a great first-time UX (forcing users to change configuration to make it "just work"), so I think we're going to wait until the build is available.

billinghamj:
Terraform itself is now fully natively available too! 😄  Along with most/all major providers 👌 I think it may be time!

radeksimko:
@billinghamj The team has plans to release native darwin/arm64 support soon, but it has not done so yet from what I know:
https://releases.hashicorp.com/terraform/0.15.0/

It is possible that you compiled Terraform for that platform yourself, or downloaded a binary compiled by a third party?
This may work, but it's not explicitly tested/supported by the maintaining team yet.

billinghamj:
A lot of people get terraform from brew, which has been distributing it compiled for arm64 for many months. Hashicorp's insistence to keep waiting for other dependencies has resulted in Terraform being - by far - the slowest to make progress on Apple Silicon support across all the tools I use

I mean no ill will to you, but as a policy it's really very frustrating and has unnecessarily delayed every single component of Terraform :\ Maybe if someone of the engineers at Hashicorp used M1 machines, you might realise how bad it's been 😞

radeksimko:
Terraform 1.0.2 was released with darwin/arm64 support.

There is a bug fix release scheduled to go out soon, but I'll make sure this is part of the following release, most likely next week.

src/languageServerInstaller.ts
```diff
@@ -153,11 +153,7 @@ function goArch(): string {
 	if (arch === 'x64') {
 		return 'amd64';
 	}
-	if (arch === 'arm64' && process.platform.toString() === 'darwin') {
-		// On Apple Silicon, install the amd64 version and rely on Rosetta2
-		// until a native build is available.
-		return 'amd64';
```
ansgarm, line 159: Is this correct? Shouldn't it be `arm64` at least for Terraform 1.0.2?

radeksimko: That is a good point! We could do version filtering here 🤔 

I will try patching that on Monday 👍🏻

radeksimko: Actually these are LS versions, not Terraform versions and first LS for `darwin/arm64` was released [in February](https://github.com/hashicorp/terraform-ls/releases/tag/v0.14.0), and given that we don't really allow installing old versions & users are prompted to upgrade 24 hours after LS release I feel there's little benefit in adding that extra complexity.

In fact I would argue that we _want_ users to upgrade, so if returning an error for an _old_ version can help this, then I'd say it's even better?

ansgarm (approved):

PR #570, opened 2021-02-26
Cancel language server install when upgrade message is closed

### LINKED ISSUE #519: VSCode language server popup installs latest on clicking x (to close)
Apologies for posting this here, but it triggers me that whenever there's a new language server update of terraform on VSCode and i click on X in order to close the popup, it immediately installs the new language server. Is this something that no other people are experiencing because I haven't seen a related ticket like this or I haven't searched for the right keywords. Thank you for your attention.

aeschright:
Fixes #519

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -9,7 +9,6 @@ import { Release, getRelease } from '@hashicorp/js-releases';
 export class LanguageServerInstaller {
 	public async install(directory: string): Promise<void> {
 		const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
-		const lsVersionCmd = `${directory}/terraform-ls --version`;
 		const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`;
 		let isInstalled = false;
 		let installedVersion: string;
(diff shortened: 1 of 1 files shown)


src/languageServerInstaller.ts
```diff
@@ -40,21 +39,15 @@ export class LanguageServerInstaller {
...
 	}
 
 	async installPkg(release: Release, installDir: string, userAgent: string): Promise<void> {
 		const destination = `${installDir}/terraform-ls_v${release.version}.zip`;
 		fs.mkdirSync(installDir, { recursive: true }); // create install directory if missing
 	
-		let os = goOs();
-		let arch = goArch();
+		const os = platformOs();
+		const arch = platformArch();
```
radeksimko, line 50: the _go_ in the name here was actually somewhat intentional, because the names can be Go specific - hence the conditions translating `win32` to `windows` or `x64` to `amd64`.

radeksimko (dismissed):

src/languageServerInstaller.ts
aeschright: Ah, that makes sense. I was thinking of it as a language server thing rather than the Go environment.

radeksimko (approved):

PR #577, opened 2021-03-08
Add stopClient and execWorkspaceCommand telemetry events

### LINKED ISSUE #576: Telemetry doesn't track when the client is stopped
To provide more context to reported errors, a telemetry event should be sent when a client is deactivated.

aeschright:
Fixes #576

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -278,6 +278,7 @@ function getDocumentClient(document: vscode.Uri): terraformLanguageClient {
 }
 
 function execWorkspaceCommand(client: LanguageClient, params: ExecuteCommandParams): Promise<any> {
+	reporter.sendTelemetryEvent('execWorkspaceCommand', { command: params.command });
 	return client.sendRequest(ExecuteCommandRequest.type, params);
 }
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -232,6 +232,7 @@ async function stopClients(folders = prunedFolderNames()) {
 		if (clients.has(folder)) {
 			promises.push(clients.get(folder).client.stop());
 			clients.delete(folder);
+			reporter.sendTelemetryEvent('stopClient');
```
radeksimko, line 235: There is probably some complexity cost to weigh this against, but is it worth tracking the context? i.e. pass an extra argument to `stopClients` which (if sent as a 2nd argument of `sendTelemetryEvent`) can tell us whether this was manually requested stop, or how was it triggered?

radeksimko (dismissed):
Just one comment in addition to that in-line one:

It may be worth tracking also command executions, given the traces we have seen in AppInsights, so we can correlate these events.

Don't treat any of this as a blocker though. 👍🏻

src/extension.ts
aeschright: I'd be worried about creating a lot of noise that wouldn't tell us all that much. I think we'll get more out of logging the command execution (and I could add that now, it's just one more line).

radeksimko: Fair enough 👍🏻

radeksimko (approved):
:shipit:

PR #579, opened 2021-03-09
Add telemetry for tracking language server installed and upgrade versions

aeschright:


src/languageServerInstaller.ts
```diff
@@ -17,11 +22,13 @@ export class LanguageServerInstaller {
 			isInstalled = true;
 		} catch (err) {
 			// TODO: verify error was in fact binary not found
+			this.reporter.sendTelemetryException(err);
```
radeksimko, line 25: Is there a way of knowing where was an exception raised from or add a flag somehow? For example if we end up looking at generic filesystem exceptions, a file not found can be because it was never written/installed, or because it was removed after installation so it might be helpful to know which case it is.

src/languageServerInstaller.ts
```diff
@@ -33,6 +40,7 @@ export class LanguageServerInstaller {
 		}
 
 		try {
+			this.reporter.sendTelemetryEvent('installingLs', { version: currentRelease.version });
 			await this.installPkg(currentRelease, directory, userAgent);
 		} catch (err) {
 			vscode.window.showErrorMessage('Unable to install terraform-ls');
```
radeksimko, line 46: Is it worth logging the installation exception here too?

radeksimko (approved):

src/languageServerInstaller.ts
radeksimko: Oh I guess we'd see that hopefully from the trace. 💡

src/languageServerInstaller.ts
aeschright: This one gets logged by the caller in `extension.ts`.

radeksimko: I see 👍🏻

PR #580, opened 2021-03-09
Improve error message for failed language server install

### LINKED ISSUE #578: Improve installation error handling for common filesystem issues
We observed some filesystem related issues in AppInsights.

On that note I tried to launch VS Code to let the extension auto-install LS after revoking some permissions on the `lsp` folder:

```
cd ~/.vscode/extensions/hashicorp.terraform-2.7.0
rm -f lsp/terraform-ls
sudo chown root:admin lsp
sudo chmod 700 lsp
```

which then resulted in a suboptimal (not very helpful) error message:

Screenshot 2021-03-08 at 21 11 40

a quick look into the log also brings up this line

> [2021-03-08 21:10:41.038] [exthost] [error] Error: EACCES: permission denied, scandir '/Users/radeksimko/.vscode/extensions/hashicorp.terraform-2.7.0/lsp' terraform.enableLanguageServer

which explains the issue, but not many users may look there.

A good error message is actionable, which in this case could be something like _"Folder .../lsp is not writable, cannot install terraform-ls"_ or just somehow include the error message shown in the log.

--- 

Lack of permissions may be the most common problem that would come to mind (on overly restrictive FS permissions of some corporate-IT-controlled machines), but ideally we should surface any FS related issues that prevent the downloaded stream of bytes of `terraform-ls` from being successfully written to the destination path.

aeschright:
Fixes #578

Screen Shot 2021-03-09 at 2 50 10 PM

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -250,7 +250,6 @@ async function pathToBinary(): Promise<string> {
 			try {
 				await installer.install(installDir);
 			} catch (err) {
-				vscode.window.showErrorMessage(err);
 				reporter.sendTelemetryException(err);
 				throw err;
 			} finally {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -147,7 +147,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 	);
 
 	if (enabled()) {
-		await vscode.commands.executeCommand('terraform.enableLanguageServer');
+		try {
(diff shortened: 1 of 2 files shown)


aeschright:
That top error message isn't that helpful to users, so I made another change to catch and report it, letting the extension still be activated. Now all they'll see is this:
Screen Shot 2021-03-09 at 3 10 23 PM

radeksimko (approved):
LGTM, thank you 🚢

PR #583, opened 2021-03-10
Match correct language server binary name per platform

### LINKED ISSUE #571: VSCode keeps installing  ls 0.14.0 and fails to start language server
Every time I open VSCode in a terraform repo, I get 2 pop ups in the bottom right of the screen. See screenshot attachments.

The first is a message saying 'Installed terraform-ls 0.14.0.'. I get this every time, so I assume the install is actually failing

The second is a message is 'Couldn't start client Language Server:'

I've tried uninstalling and reinstalling the extension and uninstalling and reinstalling VSCode entirely. I've also made sure to run `terraform init` as well as reset the `terraform.languageServer` block in the settings JSON as instructed in the README.

terraform-ls_message

aeschright:
Fixes #571


src/languageServerInstaller.ts
```diff
@@ -104,9 +100,7 @@ export class LanguageServerInstaller {
 }
 
 async function getLsVersion(dirPath: string): Promise<string> {
-	const lsBinPath = `${dirPath}/terraform-ls`;
-
-	fs.accessSync(lsBinPath, fs.constants.X_OK);
+	fs.accessSync(lsBinPath(dirPath), fs.constants.X_OK);
```
cdonnellytx, line 103: Unfortunately it looks like the method references the `lsBinPath` const in several locations; you'll need to assign it to a variable.

We'll need a new variable name though; maybe `binPath`?
```suggestion
	const binPath = lsBinPath(dirPath);

	fs.accessSync(binPath, fs.constants.X_OK);
```

(I couldn't fix the others; GitHub wouldn't let me add notes on those lines.)

aeschright: Good catch!

src/languageServerInstaller.ts
```diff
@@ -150,3 +145,11 @@ function goArch(): string {
...
 	return arch;
 }
+
+function lsBinPath(directory: string): string {
+	if (goOs() === "windows") {
+		return `${directory}/terraform-ls.exe`;
+	} else {
+		return `${directory}/terraform-ls`;
+	}
+}
```
radeksimko, line 155:
```suggestion
}

```

radeksimko (approved):

src/languageServerInstaller.ts
radeksimko: Missing EOF newline.

PR #584, opened 2021-03-11
Rescue version check errors on language server install

aeschright:
Filter and log unexpected error types

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -21,7 +21,9 @@ export class LanguageServerInstaller {
 			installedVersion = await getLsVersion(directory);
 			isInstalled = true;
 		} catch (err) {
-			if  (err.code !== 'ENOENT') { // error is not due to a missing terraform-ls binary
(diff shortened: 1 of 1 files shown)


src/languageServerInstaller.ts
```diff
@@ -21,8 +21,12 @@ export class LanguageServerInstaller {
 			installedVersion = await getLsVersion(directory);
 			isInstalled = true;
 		} catch (err) {
-			// TODO: verify error was in fact binary not found
-			this.reporter.sendTelemetryException(err);
+			// If no terraform-ls binary is found, getLsVersion will produce the error "ENOENT: no such file or directory"
+			// In that situation, it's safe to continue. Otherwise we want to stop and log the error.
```
radeksimko, line 25:
```suggestion
			// Most of the time, getLsVersion would produce "ENOENT: no such file or directory"
			// on a fresh installation (unlike upgrade). It’s also possible that the file or directory
			// is inaccessible for some other reason, but we catch that separately.
```

radeksimko (approved):

radeksimko (approved):

PR #587, opened 2021-03-16
Update client telemetry

aeschright:
A couple of changes that might help us get more visibility into errors.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -217,6 +217,10 @@ function newClient(cmd: string, location: string, commandPrefix: string) {
 		documentSelector: [{ scheme: 'file', language: 'terraform', pattern: `${f.uri.fsPath}/**/*` }],
 		workspaceFolder: f,
 		initializationOptions: initializationOptions,
+		initializationFailedHandler: (error) => {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -170,9 +171,17 @@ async function startClients(folders = prunedFolderNames()) {
 		if (!clients.has(folder)) {
 			const commandPrefix = shortUid.seq();
 			const client = newClient(command, folder, commandPrefix);
+			client.onReady().then(() => {
+				reporter.sendTelemetryEvent('startClient');
+			});
+			client.onDidChangeState((event) => {
+				if (event.newState === ClientState.Stopped) {
+					clients.delete(folder);
```
radeksimko, line 179: Is there a reason `clients.delete` should be called from here, as opposed to `stopClients`?

aeschright: This way we can ensure it's always in sync with when the client actually stops, even if that's triggered by something other than our own code.

radeksimko: I see! 👍🏻

radeksimko (approved):

PR #589, opened 2021-03-16
Normalize language server installer file paths

aeschright:
Fixes problems on some Windows environments

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -261,7 +261,7 @@ async function pathToBinary(): Promise<string> {
 	if (!_pathToBinaryPromise) {
 		let command: string = config('terraform').get('languageServer.pathToBinary');
 		if (!command) { // Skip install/upgrade if user has set custom binary path
-			const installDir = path.join(extensionPath, '/lsp');
+			const installDir = path.join(extensionPath, 'lsp');
 			const installer = new LanguageServerInstaller(reporter);
 			try {
 				await installer.install(installDir);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -260,7 +261,7 @@ async function pathToBinary(): Promise<string> {
 	if (!_pathToBinaryPromise) {
 		let command: string = config('terraform').get('languageServer.pathToBinary');
 		if (!command) { // Skip install/upgrade if user has set custom binary path
-			const installDir = `${extensionPath}/lsp`;
+			const installDir = path.join(extensionPath, '/lsp');
```
radeksimko, line 264: The documentation doesn't seem very clear on how it handles such case (except that it "allows" it), but shouldn't the platform-dependent separators be now omitted? i.e. shouldn't `/lsp` be just `lsp`?

src/extension.ts
```diff
@@ -270,7 +271,7 @@ async function pathToBinary(): Promise<string> {
 			} finally {
 				await installer.cleanupZips(installDir);
 			}
-			command = `${installDir}/terraform-ls`;
+			command = path.join(installDir, '/terraform-ls');
```
radeksimko, line 274: Same question here ^

radeksimko (commented):
This makes sense in principle, I'm just not sure about how `path.join` actually works when there is a platform-dependent separator inside of the path segment.

Generally though if it doesn't need to be there, it probably shouldn't?

src/extension.ts
aeschright: My impression was that it would normalize the separator as well, but it works without the slash so I'll go with that.

radeksimko (approved):

PR #594, opened 2021-03-18
Fix branch links for terraform-ls and vscode-terraform in readme and changelog

paulmilbank:
Terraform-LS repo has deleted master, so redirects to the default of main. 
Vscode-terraform still has the master branch, so the links from VSCode readme take you to the old changelog, not the updated one on the main branch.


radeksimko (approved):

radeksimko:
Thanks.

PR #595, opened 2021-03-22
Check for language server updates every 24 hours

### LINKED ISSUE #588: Check for language server updates daily
Currently the extension will look for a new version of terraform-ls every time it's activated, which may not be for several days if a user leaves VSCode open. We can have it check again every 24hrs so that users have access to bugfixes more consistently.

aeschright:
Fixes #588

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -8,14 +8,15 @@ import TelemetryReporter from 'vscode-extension-telemetry';
 import { exec } from './utils';
 import { Release, getRelease } from '@hashicorp/js-releases';
 
+const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
+
 export class LanguageServerInstaller {
 	constructor(
-		private reporter: TelemetryReporter
+		private reporter: TelemetryReporter,
+		private userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`
 	) { }
 
 	public async install(directory: string): Promise<void> {
-		const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
-		const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscode.version}`;
 		let isInstalled = false;
 		let installedVersion: string;
 		try {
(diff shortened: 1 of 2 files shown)


src/languageServerInstaller.ts
```diff
@@ -32,31 +36,45 @@ export class LanguageServerInstaller {
 			isInstalled = false;
```
appilon, line 36: I would just return true here early (needsInstall because isInstalled == false)

src/languageServerInstaller.ts
```diff
@@ -32,31 +36,45 @@ export class LanguageServerInstaller {
...
-			if (semver.gt(currentRelease.version, installedVersion, { includePrerelease: true })) {
-				const selected = await vscode.window.showInformationMessage(`A new language server release is available: ${currentRelease.version}. Install now?`, 'Install', 'Cancel');
-				if (selected !== 'Install') { // selected is undefined if the user closes the message
-					return;
-				}
-			} else {
-				return;
-			}
+		this.release = await this.latestRelease();
+		if (!isInstalled) {
```
appilon, line 40: we can remove this block with the early return

src/languageServerInstaller.ts
```diff
@@ -32,31 +36,45 @@ export class LanguageServerInstaller {
...
 		} catch (err) {
 			vscode.window.showErrorMessage(`Unable to install terraform-ls: ${err.message}`);
 			throw err;
 		}
 
-		this.showChangelog(currentRelease.version);
+		this.showChangelog(this.release.version);
+	}
+
+	_latestReleasePromise: Promise<Release>;
```
appilon, line 66: I'm not sure we want to cache this result? It seemed like updates were checked for on a schedule now, so I think we don't want to cache the response, that way when a new release goes out, the logic will show a mismatch between what the user has installed and the new latest

appilon (changes requested):
Minor feedback, I may be mistaken on the caching thing, if so then overall looks good!

src/languageServerInstaller.ts
aeschright: I'll remove it, I was thinking I would need it between checking whether there's an update and installing it, but I went in a different direction for that part.

appilon (approved):
LGTM

PR #604, opened 2021-04-09
Update syntax highlighting for Terraform 0.15

### LINKED ISSUE #603: Deprecate list and map in syntax highlighting
The functions `list` and `map` will be fully [deprecated in Terraform 0.15](https://github.com/hashicorp/terraform/pull/26818) and the syntax highlighting should reflect this.

aeschright:
Closes #603

I fixed another small (but recurrent) bug while I was in there, which is on the second commit.

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -264,11 +264,11 @@
 			]
 		},
 		"string_interpolation": {
-			"begin": "(\\G|[^%$])[%$]{",
+			"begin": "(\\G|[^%$])([%$]{)",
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
It would be really useful to have tests, but I guess that's something to look into as part of the overhaul to a more generic HCL grammar.

Relatedly a generic HCL grammar will/should likely be Terraform version agnostic and not care about what particular functions are supported.

magnetikonline:
@aeschright thank you so much - this fixes that annoying visual bug with string interpolation running over to the right hand side! yay! 🎉

PR #608, opened 2021-04-19
Update incorrect dates to CHANGELOG.md

altern8tif:
Changed the incorrect years from `2020` to `2021` for the last 3 versions.


aeschright (approved):
Thanks for fixing that! 2020 was long enough as is 😅

PR #612, opened 2021-04-23
Update js-releases dep to resolve security issue

### LINKED ISSUE #611: Update terraform-ls release verification to use new public key
HashiCorp has rotated and revoked the public GPG key we use to verify terraform-ls packages due to [a reported vulnerability](https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512). For this project, that means that the dependency [js-releases](https://github.com/hashicorp/js-releases/pull/9) needs to be updated with the new GPG key. As soon as that update is published and we can verify that previous terraform-ls releases have been re-signed, we'll push out an update.

In the interim, new users of this extension might experience errors when installing the language server. Existing users should be unaffected, since terraform-ls releases are on hold while this incident is being resolved. Please see the [security discussion thread](https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512) if you have questions about the incident as a whole.

aeschright:
The GPG key used to sign the package signature files has been rotated per [HCSEC-2021-12](https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512). This update will allow to continue validating `terraform-ls` packages as part of the install process.

Closes #611

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,16 +13,15 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest, macos-11.0, ubuntu-latest]
-        node-version: [12.x, 14.x, 16.x]
     runs-on: ${{ matrix.os }}
 
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 14.x
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 14.x
       - name: Set up Xvfb (Ubuntu)
         run: |
           /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,11 +13,16 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest, macos-11.0, ubuntu-latest]
+        node-version: [12.x, 14.x, 16.x]
     runs-on: ${{ matrix.os }}
 
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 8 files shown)


radeksimko (approved):
The test failures seem unrelated, so LGTM, but it would be nice to get those failures addressed async.

src/test/helper.ts
```diff
@@ -30,8 +30,6 @@ async function activated() {
 			} else {
 				console.log('hashicorp.terraform is already active');
 			}
-			// make sure language server download is complete
-			await ext.exports.pathToBinary();
```
radeksimko, line 34: Just to persist context for the future - why is this being removed?

.github/workflows/test.yml
```diff
@@ -18,6 +18,10 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
```
radeksimko, line 24: I wish we could somehow sync this up with the `.nvmrc` file at some point 🤔

package.json
```diff
@@ -7,7 +7,7 @@
   "publisher": "hashicorp",
   "license": "MPL-2.0",
   "engines": {
-    "vscode": "^1.46.0"
+    "vscode": "^1.50.1"
```
radeksimko, line 10: Is there a particular feature/change between these versions you were hoping would address the test failures?

radeksimko (approved):
Nothing major, just some questions about context which would be good to add for future readers that may stumble upon this commit/PR later.

src/test/helper.ts
aeschright: It was no longer doing what it claimed to -- setting the path is now separate from installing the language server.

.github/workflows/test.yml
aeschright: That could be handy, the main reason I added this to the test.yml is in case the default on the system changes.

package.json
aeschright: I've been keeping this set to a six-month window so that we aren't trying to maintain compatibility way outside of what people are likely to be using. It doesn't affect the test issues.

PR #615, opened 2021-04-29
Add hcl alternative extension

RustyF:
hcl is the default extension used for packer scripts converted to hcl syntax.  Without this change, it is necessary to manually select the file type of "terraform".

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
           "terraform"
         ],
         "extensions": [
-          ".tf"
+          ".tf",
+          ".hcl"
         ],
         "configuration": "./language-configuration.json"
       },


radeksimko:
Hi @RustyF  
Sorry for the long delay in reviewing your PR.

The grammar which is currently in use by the extension is largely Terraform-specific, which is why I'd be very hesitant to highlighting just any "HCL" code with it, as this PR proposes.

We do have plans to publish a more generic grammar as per https://github.com/hashicorp/vscode-terraform/issues/638 and it is likely that this grammar would replace the existing one, but even then such grammar would - in the context of this (Terraform) extension - would still be used to highlight **Terraform** code, not any other HCL (i.e. not Packer).

However a [planned generic HCL extension](https://github.com/hashicorp/vscode-terraform/issues/639) could theoretically highlight Packer files as they follow helpful convention `*.pkr.hcl` and that extension would most likely claim `*.hcl`.

With that in mind I'm going to close this PR, but feel free to subscribe to the mentioned issues and/or contribute with any ideas and opinions there.

Thank you for your understanding.

PR #633, opened 2021-05-14
Rename rootModules command to module.callers

### LINKED ISSUE #632: Rename rootModules command interface
Because the internal representation of modules in terraform-ls has changed, the [`rootModules` language server command is going to be updated](https://github.com/hashicorp/terraform-ls/pull/508) to `modules.callers`. In order for the extension to continue to access a list of modules that can be used to call `terraform init`, the command interface will need to be updated to match it.

aeschright:
Pending https://github.com/hashicorp/terraform-ls/pull/508
Closes #632

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,6 +10,7 @@ import {
 	Executable,
 	State as ClientState
 } from 'vscode-languageclient/node';
+import { Utils } from 'vscode-uri'
 import * as path from 'path';
 import ShortUniqueId from 'short-unique-id';
 import TelemetryReporter from 'vscode-extension-telemetry';
@@ -25,7 +26,6 @@ import {
 } from './vscodeUtils';
 import {
 	SingleInstanceTimeout,
-	sleep
 } from './utils';
 
 interface terraformLanguageClient {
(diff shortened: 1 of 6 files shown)


radeksimko:
I pushed some more changes, all of which should be explained in the commit message.

One pending question from me is that `relPath` (previously "friendly name") is no longer used anywhere, so I'm inclined to just drop it from the API entirely.

src/extension.ts
```diff
@@ -320,50 +324,45 @@ function execWorkspaceCommand(client: LanguageClient, params: ExecuteCommandPara
...
+interface moduleCaller {
 	uri: string,
-	name: string
+	rel_path: string
 }
 
 interface moduleCallersResponse {
-	moduleCallers: module[],
-	needsInit: boolean
+	v: number,
```
aeschright, line 333: Is this the API version? Even though it's not being referenced, let's give it a more descriptive name.

aeschright:
> One pending question from me is that `relPath` (previously "friendly name") is no longer used anywhere, so I'm inclined to just drop it from the API entirely.

Makes sense to me, if the language server is no longer displaying it to the user, there's no reason to share that.

src/extension.ts
radeksimko: yep, updated.

radeksimko:
> Makes sense to me, if the language server is no longer displaying it to the user, there's no reason to share that.

See https://github.com/hashicorp/terraform-ls/pull/525

radeksimko (dismissed):

radeksimko (approved):

PR #634, opened 2021-05-16
feat: use arm64 ls builds on darwin

johanneswuerbach:
Native arm64 builds are now available https://github.com/hashicorp/terraform-ls/releases/tag/v0.16.3


radeksimko:
Thank you for the PR but I will close this as a duplicate of #563 which will be merged most likely next week.

PR #642, opened 2021-05-21
Update changelog URL

alexjurkiewicz:
Link to the latest copy of the changelog document, rather than release page.

There are no changelog notes for terraform-ls GitHub releases, so link directly to the changelog file instead.

Since this plugin is only used to install the latest terraform-ls version, link to the main branch `CHANGELOG.md`, rather than the version as of the appropriate tag. Sometimes the changelog is modified after the tag with forgotten information, and we don't want to hide that from users.

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -107,7 +107,7 @@ export class LanguageServerInstaller {
 		vscode.window.showInformationMessage(`Installed terraform-ls ${version}.`, "View Changelog")
 			.then(selected => {
 				if (selected === "View Changelog") {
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @alexjurkiewicz ,
Thanks for the contribution and sorry for the confusion.

The problem here is that the `terraform-ls` changelog updating is [not automated yet](https://github.com/hashicorp/terraform-ls/issues/63) and I just forgot to update it for the last release (manually).
It is updated now: https://github.com/hashicorp/terraform-ls/releases/tag/v0.17.0

I would still prefer for the extension to keep linking to a particular version, rather than just `main`, especially because `main` contains the _whole_ changelog which also it often contains changes for the upcoming unreleased version.

Although the extension does install latest versions only at this point, it may not stay that way, see https://github.com/hashicorp/vscode-terraform/issues/628

PR #654, opened 2021-05-27
Avoid launching more servers if server supports multiple folders

### LINKED ISSUE #627: Only launch new instances of LS if it doesn't support multiple folders
The extension communicates via client capabilities that it supports multiple folders:

```json
{
    "processId": 85078,
    "clientInfo": {
        "name": "vscode",
        "version": "1.56.0"
    },
    "rootPath": "/private/var/workspace/tf-test/github",
    "rootUri": "file:///private/var/workspace/tf-test/github",
    "capabilities": {
        "workspace": {
            "workspaceFolders": true,
```
this can also be seen by the client sending all opened folders, as expected in the `initialize` request:

```json
    "workspaceFolders": [
        {
            "uri": "file:///private/var/workspace/tf-test/local",
            "name": "local"
        },
        {
            "uri": "file:///private/var/workspace/tf-test/netlify",
            "name": "netlify"
        },
        {
            "uri": "file:///private/var/workspace/tf-test/github",
            "name": "github"
        }
    ]
```

The LS therefore should have a way of opting into multi-folder mode via server capabilities and prevent the extension from launching more processes.

However when it does so:

```json
{
    "capabilities": {
        "workspace": {
            "workspaceFolders": {
                "supported": true,
                "changeNotifications": "folder-change"
            }
        }
```

the extension still launches 3 instances of LS (1 for each of 3 folders) anyway.

WIP branch with experimental multi-folder support is available from https://github.com/hashicorp/terraform-ls/pull/502

### LINKED ISSUE #672: Completion of module `source` and `version`
## Problem Statement

Module `source`s are not always easy to get right and often times these are copy-pasted, or typed manually (which increases the risk of typos). Either way the user would benefit from some extra help there.

```hcl
module "example" {
  source = # HERE
}
```

Relatedly users would benefit from being able to pick a version from a list, instead of typing it in.

### `source`

 - [x] User will be provided with any known/tracked local modules (i.e. local file paths)
 - [x] User will be provided with any matching module sources from the Registry
 - [ ] Required inputs will be pre-filled, if the feature is enabled

https://user-images.githubusercontent.com/287584/164701473-35df4ded-ecb9-420e-907d-e38ebe83ca4e.mov

### `version`

User will be provided with a list of published versions for a given Terraform Registry-published module.

https://user-images.githubusercontent.com/287584/164701664-856fb6ef-5aab-47d9-98b5-697bda3b4935.mov

Typing will filter versions and versions will be sorted with the latest ones on top (unlike in the recorded demo).

## Proposal

https://github.com/hashicorp/terraform-ls/issues/869

## Implementation Notes

LSP autocompletion part of the spec uses `isIncomplete` which we should probably use for the dynamic results from the API
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

> for speed clients should be able to filter an already received completion list if the user continues typing. Servers can opt out of this using a [CompletionList](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList) and mark it as isIncomplete.

radeksimko:
Fixes #627

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -203,26 +203,48 @@ async function startClients(folders = prunedFolderNames()) {
 	console.log('Starting:', folders);
 	const command = await pathToBinary();
 	const disposables: vscode.Disposable[] = [];
-	for (const folder of folders) {
-		if (!clients.has(folder)) {
-			const commandPrefix = shortUid.seq();
-			const client = newClient(command, folder, commandPrefix);
-			client.onReady().then(() => {
-				reporter.sendTelemetryEvent('startClient');
-			});
-			client.onDidChangeState((event) => {
-				if (event.newState === ClientState.Stopped) {
-					clients.delete(folder);
-					reporter.sendTelemetryEvent('stopClient');
-				}
-			});
-			disposables.push(client.start());
-			clients.set(folder, { commandPrefix, client });
-		} else {
-			console.log(`Client for folder: ${folder} already started`);
+
+	const firstClient = configureClient(command, folders[0])
+	firstClient.client.onReady().then(() => {
+		reporter.sendTelemetryEvent('startClient');
+		const multiFoldersSupported = firstClient.client.initializeResult.capabilities.workspace?.workspaceFolders?.supported;
+
+		console.log("Multi-folder support: ", multiFoldersSupported);
+		const remainingFolders = folders.slice(1);
+		for (const folder of remainingFolders) {
+			if (multiFoldersSupported) {
+				// Reuse the same client if server supports multiple workspace folders
+				clients.set(folder, firstClient);
+				continue
+			}
+
+			if (!clients.has(folder)) {
+				const c = configureClient(command, folders[0]);
+				disposables.push(c.client.start());
+				clients.set(folder, c);
+			} else {
+				console.log(`Client for folder: ${folder} already started`);
+			}
 		}
-	}
+	});
+
+	disposables.push(firstClient.client.start());
+	clients.set(folders[0], firstClient);
+	
 	return disposables;
+}
+
+function configureClient(binPath:string, folder: string) {
+	const commandPrefix = shortUid.seq();
+	const client = newClient(binPath, folder, commandPrefix);
+
+	client.onDidChangeState((event) => {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -199,28 +199,51 @@ async function startClients(folders = prunedFolderNames()) {
...
-		} else {
-			console.log(`Client for folder: ${folder} already started`);
+
+	const firstClient = configureClient(command, folders[0])
+	firstClient.client.onReady().then(() => {
+		reporter.sendTelemetryEvent('startClient');
+		const wsSupported = firstClient.client.initializeResult.capabilities.workspace.workspaceFolders.supported
+
+		console.log("Multi-folder support: ", wsSupported)
+		for (var i = 1; i < folders.length; i++) {
```
radeksimko, line 209: Is there a cleaner/better way of iterating over an array while skipping the first item? Something like `folders[1:]` in Go or Python? 😄

src/extension.ts
```diff
@@ -199,28 +199,51 @@ async function startClients(folders = prunedFolderNames()) {
...
-					clients.delete(folder);
-					reporter.sendTelemetryEvent('stopClient');
-				}
-			});
-			disposables.push(client.start());
-			clients.set(folder, { commandPrefix, client });
-		} else {
-			console.log(`Client for folder: ${folder} already started`);
+
+	const firstClient = configureClient(command, folders[0])
```
radeksimko, line 203: Not sure `firstClient` is the best name for what this is, but I couldn't come up with a better one :|

aeschright: How about `primaryClient`?

radeksimko: Yeah, that's better, although I was more referring to the fact that the variable contains the commandPrefix + the client instance.

src/extension.ts
aeschright: In javascript it's `folders.slice(1)`

src/extension.ts
aeschright: I agree "client" is getting a little overloaded – one direction you could go is to use `client`, `clientWithPrefix`, `clientWithConfig`, etc.

Issue #628, opened 2021-05-11
Allow pinning language server to a version


radeksimko:
Ideally users should not have any strong reasons to be pinning LS to any particular older version. This assumes that latest version always provides the latest (security) patches and stability.

However sometimes it is useful, esp. when debugging to be able to pin to an older version, so that the user can verify that a bug they're experiencing is only present in particular versions.

I imagine this could be exposed to the user in the form of a field in settings, say "terraform-ls version" which could be either a selectbox of last N versions, or text box accepting version constraints such as `~> 0.12`. I don't have any strong opinion on the UI side of things, also I'm not sure what's realistic within the VS Code API constraints.

jpogran:
This was implemented in #656

PR #656, opened 2021-06-04
(GH-628) Add semver based version pin for Language Server

glennsarti:
Partially solves #628 

This PR does not address any UI for selecting a version pin.  It merely adds all of the plumbing to make the version pin actually work.  A later PR would then add the UI layer.

---

Previously the Language Server would always update to the latest version however
there are scenarios where using an older version is required.  This commit adds
a new `version` configuration setting which is a semver range (as per the
@hashicorp/js-releases module). This setting is then used to find an appropriate
release when calling the Hashi release API.

This commit also adds validation on the `version` setting to ensure it is a
valid string and warns the user of the potentially bad setting.

This commit then updates the Language Server installer to not only upgrade the
server, but also downgrade it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,10 @@
               ],
               "default": "off",
               "description": "Traces the communication between VS Code and the language server."
+            },
+            "required_version": {
+              "type": "string",
+              "description": "The required version of the Language Server described as a semantic version string, for example '^2.0.1' or '> 1.0'. Defaults to latest available version."
             }
           },
           "default": {
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
               "default": "off",
               "description": "Traces the communication between VS Code and the language server."
             },
-            "required_version": {
+            "requiredVersion": {
               "type": "string",
               "description": "The required version of the Language Server described as a semantic version string, for example '^2.0.1' or '> 1.0'. Defaults to latest available version."
             }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,7 +6,7 @@ import {
 import { LanguageClient } from 'vscode-languageclient/node';
 import { Utils } from 'vscode-uri'
 import TelemetryReporter from 'vscode-extension-telemetry';
-import { LanguageServerInstaller } from './languageServerInstaller';
+import { LanguageServerInstaller, isValidVersionString, defaultVersionString } from './languageServerInstaller';
 import { ClientHandler, TerraformLanguageClient } from './clientHandler';
 import { config, prunedFolderNames } from './vscodeUtils';
 import { SingleInstanceTimeout } from './utils';
@@ -36,6 +36,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 			await config('terraform').update('languageServer', { enabled: undefined, external: true }, vscode.ConfigurationTarget.Global);
 		} catch (err) {
 			console.error(`Error trying to erase pre-2.0.0 settings: ${err.message}`);
+		}
+	}
+
+	if (config('terraform').has('languageServer.required_version')) {
+		const langServerVer = config('terraform').get('languageServer.required_version', defaultVersionString)
+		if (!isValidVersionString(langServerVer)) {
+			vscode.window.showWarningMessage(`The Terraform Language Server Version string '${langServerVer}' is not a valid semantic version and will be ignored.`);
 		}
 	}
 
@@ -165,7 +172,7 @@ async function updateLanguageServer(clientHandler: ClientHandler, lsPath: Server
 		// skip install if a language server binary path is set
 		if (!lsPath.hasCustomBinPath()) {
 			const installer = new LanguageServerInstaller(lsPath, reporter);
-			const install = await installer.needsInstall();
+			const install = await installer.needsInstall(config('terraform').get('languageServer.version', defaultVersionString));
 			if (install) {
 				await clientHandler.stopClients();
 				try {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -39,8 +39,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 		}
 	}
 
-	if (config('terraform').has('languageServer.required_version')) {
-		const langServerVer = config('terraform').get('languageServer.required_version', defaultVersionString)
+	if (config('terraform').has('languageServer.requiredVersion')) {
+		const langServerVer = config('terraform').get('languageServer.requiredVersion', defaultVersionString)
 		if (!isValidVersionString(langServerVer)) {
 			vscode.window.showWarningMessage(`The Terraform Language Server Version string '${langServerVer}' is not a valid semantic version and will be ignored.`);
 		}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -172,7 +172,7 @@ async function updateLanguageServer(clientHandler: ClientHandler, lsPath: Server
 		// skip install if a language server binary path is set
 		if (!lsPath.hasCustomBinPath()) {
 			const installer = new LanguageServerInstaller(lsPath, reporter);
-			const install = await installer.needsInstall(config('terraform').get('languageServer.version', defaultVersionString));
+			const install = await installer.needsInstall(config('terraform').get('languageServer.requiredVersion', defaultVersionString));
 			if (install) {
 				await clientHandler.stopClients();
 				try {
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -9,6 +9,11 @@ import { ServerPath } from './serverPath';
 import { Release, getRelease } from '@hashicorp/js-releases';
 
 const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
+export const defaultVersionString = "latest";
+
+export function isValidVersionString(value: string): boolean {
+	return semver.validRange(value,  { includePrerelease: true, loose: true }) !== null;
+}
 
 export class LanguageServerInstaller {
 	constructor(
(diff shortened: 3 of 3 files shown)


glennsarti:
For reviewers (ping @radeksimko)

* Local testing worked to upgrade/downgrade the language server
* I'm not 100% sure on the wording of the setting, or `console.log` statements.  Happy to work on that.
* I added some more debugging to help future-you when someone logs an issue that the updater isn't working
* This could be tested but I'm not sure how much could be done by automated unit tests. Open to thoughts about that

glennsarti:
ping to keep this alive

radeksimko:
Thanks for the PR, it is well alive 😄  I just haven't had a moment to review it.

Thanks for the patience.

glennsarti:
Those pesky PR bots close things if you're not careful :-)

radeksimko:
FWIW our bots are careful enough to never close active PRs unless a maintainer assigned a particular label to them 😉 

https://github.com/hashicorp/vscode-terraform/blob/5ad88b2e7e71fcd3a437777a3676af30d395ee5f/.github/workflows/stale.yml#L15

package.json
```diff
@@ -123,6 +123,10 @@
               ],
               "default": "off",
               "description": "Traces the communication between VS Code and the language server."
+            },
+            "version": {
```
radeksimko, line 127: Nitpick: Can we call this `required_version`? Just to have it aligned with the naming convention within Terraform and to communicate more explicitly that a constraint is expected (rather than just a version) https://www.terraform.io/docs/language/settings/index.html#specifying-a-required-terraform-version

src/extension.ts
```diff
@@ -60,6 +61,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 		}
 	}
 
+	if (config('terraform').has('languageServer.version')) {
+		const langServerVer = config('terraform').get('languageServer.version', defaultVersionString)
+		if (!isValidVersionString(langServerVer)) {
+			vscode.window.showWarningMessage(`The Terraform Language Server Version string '${langServerVer}' is not a valid semantic version and will be ignored.`);
+		}
+	}
```
radeksimko, line 69: Thank you for adding the validation!

Is there a particular reason why do it here, instead of prior to updating below? 🤔 

Relatedly I think that the user would benefit from even more validation - e.g.

1. check for conflict between `pathToBinary` and the version
2. check that there is at least 1 version which satisfies the constraint

Re (2) the current UX when I try to set a constraint which doesn't satisfy any version is not ideal:

Screenshot 2021-06-17 at 16 33 50

radeksimko (commented):
It looks good functionally, I just left some comments about the UX.

I appreciate that the UI element is left for later 👍🏻  but I still feel that even with JSON-only config options we should provide a decent experience.

package.json
glennsarti: Sure that can be done.

src/extension.ts
glennsarti: > Is there a particular reason why do it here, instead of prior to updating below? 

There feels like there are two layers of validation here..

**"Static" Validation** (Doesn't need to call out to "anything")
Is the `required_version` string even a valid semver string?
Did the user also specify a conflicting setting e.g. `pathToBinary`?

**"Dynamic" Validation** (Needs to call out to "something" for validation)
Does the valid semver string constraint actually return at least 1 possible version.?

So the reason I "static" validation at the top was to validate the settings as soon as possible and prompt the user. Note that this validation is synchronous and blocking the extension activation so it needs to be fast. (TODO - Need to check if `showWarningMessage` blocks the extension activation.  Don't think so)

Now, as you suggested, I could move the validation to be later...that's fine. This is just the approach I took.

glennsarti: > Re (2) the current UX when I try to set a constraint which doesn't satisfy any version is not ideal:

UX wise we are constrained by VScode. 

Interesting... that UI isn't from the updater code....

Ahh ok... it happens when Fresh install of extension and a valid semver string but has no versions which satisfies the constraint.

Yeah, may need to change the code  to request the default version if  no applicable version could be found AND there is no existing install of the LS.

package.json
glennsarti: Done.

src/extension.ts
glennsarti: Ok... so now it will attempt to find the required_version, and fall back to the default version, and then disable updating if it can't find any version.

This doesn't fix another problem I _think_ I see, but that was already there.

* First time user with the extension installed (but not activated)
* ... with no internet connection (or blocked getting to the API endpoint)
* ... It can't download the LSP, so the extension will throw the error.

glennsarti:
... and rebased.  @radeksimko   Back for review

src/extension.ts
radeksimko: Yes - I am aware this is a problem - or that the UX is confusing at the least - I already started working on one part of the problem in https://github.com/hashicorp/vscode-terraform/pull/691

I reckon there's still more to address in that area though.

radeksimko (commented):
I am sincerely sorry for causing the conflicts here, but I did some refactoring and changed the error handling recently, so that exceptions are caught and reported to the user instead of silently ignored.

Which is not to say that I'm suggesting to use exception for reporting a validation problem - falling back `latest` + displaying warning is still good course of action. 👍🏻 

I like the code otherwise.

All that said I assume that we'll need to rebase this PR. Are you willing to do it? Alternatively I can take it over.

glennsarti:
Rebased... Just need to retest manually

src/extension.ts
glennsarti: I noticed that you've now got a Client Handler, so in theory "we/you" should be able to start the extension and NOT the language server/client part so you can degrade gracefully.  Definitely on the right track!

glennsarti:
Did some quick testing and looks good.

glennsarti:
Back to you @radeksimko

package.json
```diff
@@ -124,6 +124,10 @@
               ],
               "default": "off",
               "description": "Traces the communication between VS Code and the language server."
+            },
+            "required_version": {
```
radeksimko, line 128:
```suggestion
            "requiredVersion": {
```
Just to maintain the consistency with other config options here

src/extension.ts
```diff
@@ -39,6 +39,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 		}
 	}
 
+	if (config('terraform').has('languageServer.required_version')) {
+		const langServerVer = config('terraform').get('languageServer.required_version', defaultVersionString)
```
radeksimko, line 43:
```suggestion
	if (config('terraform').has('languageServer.requiredVersion')) {
		const langServerVer = config('terraform').get('languageServer.requiredVersion', defaultVersionString)
```

src/extension.ts
```diff
@@ -165,7 +172,7 @@ async function updateLanguageServer(clientHandler: ClientHandler, lsPath: Server
 		// skip install if a language server binary path is set
 		if (!lsPath.hasCustomBinPath()) {
 			const installer = new LanguageServerInstaller(lsPath, reporter);
-			const install = await installer.needsInstall();
+			const install = await installer.needsInstall(config('terraform').get('languageServer.version', defaultVersionString));
```
radeksimko, line 175:
```suggestion
			const install = await installer.needsInstall(config('terraform').get('languageServer.requiredVersion', defaultVersionString));
```

src/languageServerInstaller.ts
```diff
@@ -19,10 +24,30 @@ export class LanguageServerInstaller {
...
-	public async needsInstall(): Promise<boolean> {
+	private async getRequiredVersionRelease(versionString: string): Promise<Release> {
 		try {
-			this.release = await getRelease("terraform-ls", "latest", this.userAgent);
+			const release = await getRelease("terraform-ls", versionString, this.userAgent);
+			console.log(`Found Terraform language server version ${release.version} which satisfies range '${versionString}'`);
+			return release;
 		} catch (err) {
+			if (versionString == defaultVersionString) { throw err; }
+			console.log(`Error while finding Terraform language server release which satisfies range '${versionString}' and will reattempt with '${defaultVersionString}': ${err}`)
```
radeksimko, line 34:
```suggestion
			console.log(`Error while finding Terraform language server release which satisfies range '${versionString}' and will reattempt with '${defaultVersionString}': ${err}`)
			vscode.window.showWarningMessage(`No version matching ${versionString} found, searching for ${defaultVersionString} instead`);
```

I think this would be a valuable feedback to the end user also.

src/languageServerInstaller.ts
```diff
@@ -43,12 +69,18 @@ export class LanguageServerInstaller {
 		}
 
 		this.reporter.sendTelemetryEvent('foundLsInstalled', { terraformLsVersion: installedVersion });
-		const upgrade = semver.gt(this.release.version, installedVersion, { includePrerelease: true });
-		if (upgrade) {
+
+		if (semver.eq(this.release.version, installedVersion, { includePrerelease: true })) {
+			// Already at the specified versoin
```
radeksimko, line 74:
```suggestion
			// Already at the specified version
```

radeksimko (dismissed):
This looks good to me from UX perspective! Thank you for the rebase and addressing the feedback.

I just left you some last minor suggestions. I will probably commit those in & merge the PR the next few hours unless you address it first, just so I can expedite the PR and get it shipped with the next release which is imminent. 😄

glennsarti:
> I just left you some last minor suggestions. I will probably commit those in & merge the PR the next few hours unless you address it first,

Passed my bedtime ... PR is yours to do with as you will.

radeksimko (approved):

radeksimko:
Thank you for the PR.

PR #661, opened 2021-06-11
Add support for `tfvars` files

### LINKED ISSUE #426: Please add support for formatting .tfvars files
The old version of the Terraform extension formatted `.tfvars` files so why doesn't the new version?  Please restore this feature.

Thanks

beandrad:
To add completion/hover/highlighting on tfvars (variable files).

Fixes #426

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,3483 @@
 {
   "name": "terraform",
   "version": "2.12.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "terraform",
+      "version": "2.12.1",
+      "license": "MPL-2.0",
+      "dependencies": {
+        "@hashicorp/js-releases": "^1.4.0",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,8 +1,3482 @@
 {
   "name": "terraform",
   "version": "2.12.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
radeksimko, line 4: Happy to upgrade - Just for posterity - which version of NPM do you have installed locally?

I wish there was a way of pinning it, like Node version via https://github.com/hashicorp/vscode-terraform/blob/main/.nvmrc 🤔

beandrad: I have version 7.6.3, a bit old 🙈

radeksimko: which is still newer than mine (6.14.8) 😂  🙈 

... which explains the `lockfileVersion` bump I guess ...

radeksimko (approved):
LGTM, thanks!

PR #674, opened 2021-06-20
feat: add terraform destroy command

### LINKED ISSUE #646: Implement `terraform destroy` as a command
## Background

The extension currently implements the following:

 - `terraform init`
   - proxied to Terraform LS: https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#terraforminit
 - `terraform plan`
   - executed directly in the integrated terminal
 - `terraform apply`
   - executed directly in the integrated terminal
 - `terraform validate`
   - proxied to Terraform LS: https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#terraformvalidate

The difference in how each command is integrated relate to how the output/outcome is mappable to LSP and how it can be communicated to the user.

## Proposal

Implement `terraform destroy` as command executable from the command palette.

It may be easiest to just implement it initially so that it executes in the integrated terminal. We can later assess whether any integration with LS is useful.

dbanck:
Hi all!

This PR adds the `Terraform: destroy` command and resolves #646.

When using the existing `terraformCommand` function to run the command, the user will see this prompt before the command is executed inside the terminal.

CleanShot 2021-06-20 at 17 14 27@2x

I think this prompt doesn't convey the potential danger of running destroy.

## Additional prompts

There are (at least) two more options for adding a prominent prompt before running the command.

**Option A**
This will open a modal at the position of the command palette of VSCode and ask for user confirmation.

I like this option because it keeps the user focus at the same position and locks the editor until a decision is made. 

CleanShot 2021-06-20 at 17 22 23@2x

**Option B**
This will open a warning message in the bottom right corner where most notifications appear and ask for user confirmation.

Opening the message at this position could go unnoticed, and it could seem that the command didn't do anything at all.

CleanShot 2021-06-20 at 17 23 09@2x

## Solution

Which option do you prefer? This PR currently implements option A.

I couldn't find any tests for the other registered commands, so I added none for this new command. But I would suggest adding some kind of tests for commands, too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,10 @@
       {
         "command": "terraform.plan",
         "title": "Terraform: plan"
+      },
+      {
+        "command": "terraform.destroy",
+        "title": "Terraform: destroy"
       }
     ]
   },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -105,7 +105,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 			await terraformCommand('validate');
 		}),
 		vscode.commands.registerCommand('terraform.destroy', async () => {
-			await terraformCommand('destroy', false);
+			const result = await vscode.window.showWarningMessage('Do you really want to destroy all resources?', { modal: true}, 'Yes');
+
+			if (result) {
+				await terraformCommand('destroy', false);
+			}
 		}),
 		vscode.workspace.onDidChangeConfiguration(
 			async (event: vscode.ConfigurationChangeEvent) => {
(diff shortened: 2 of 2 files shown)


jpogran:
Thanks @dbanck for the contribution! Apologies we haven't replied sooner.

I feel like the modal dialog is necessary here because the action is potentially destructive. The other option is too temporary and easily missed.

I don't see this handling the prompt for user confirmation by passing `-auto-approve`. Are you expecting the user to see the prompt in the terminal and know what to do? I'm asking because I'm walking through the UX here and trying to find a good balance. Longer term, we're trying to figure out how we surface these kinds of commands that either require user input or should show the user results. A similar scenario is https://github.com/hashicorp/vscode-terraform/pull/700.

While we figure this out, I'm going to mark this as a draft and add this to our planning board.

PR #688, opened 2021-07-15
Fix DocumentSelector for multi-folder workspace

### LINKED ISSUE #685: Doesn't work in multi-folder workspaces
vscode-terraform: 2.13.0
terraform-ls: 0.19.0

Scenario 1, have following VSCode workspace structure:
```json
{
  "folders": [
    {
      "name": "TF Project 1",
      "path": "tf1"
    },
    {
      "name": "TF Project 2",
      "path": "tf2"
    },
    {
      "name": "TF Project 3",
      "path": "tf3"
    }
  ]
}
```
*Result:*: doesn't work in anything but `TF Project 1`. I.e. it works only for the first folder, if i reorder and move Project 2 to the top, it's starting to work here, but not in Project 1

Scenario 2:
```json
{
  "folders": [
    {
      "name": "NON TF Project",
      "path": "nontf"
    },
    {
      "name": "TF Project 1",
      "path": "tf1"
    },
    {
      "name": "TF Project 2",
      "path": "tf2"
    },
    {
      "name": "TF Project 3",
      "path": "tf3"
    }
  ]
}
```

*Result:*": Doesn't work anywhere

radeksimko:
Fixes #685 

### Does it affect me?

This bug was introduced in 2.12.0 and affects any folders within workspaces of multiple folders which are not descendants of the 1st folder in the workspace.

For example, if workspace has two folders, and
 - 1st one `/my-repo` and 2nd `/my-repo/subfolder` then that workspace is **not** affected by the bug
 - 1st one `/my-repo` and 2nd `/another-repo` then that workspace **is** affected by the bug

The bug affects only users on language server `0.19+` (which is the first version that supports multiple folders natively).

### How did this happen?

Prior to this patch we would always launch the first instance of a client/server, to obtain the server capabilities and see if the server supports multiple folders. We would then continue launching more instances for any more folders if it doesn't, or reuse the same instance if it does.

The problem with reusing the same instance is that the instance is launched with a very specific document selector:

https://github.com/hashicorp/vscode-terraform/blob/583bdf59a19d46a0f5e2325f0d7a11379e0bd337/src/extension.ts#L284-L285

This prevents the client from receiving/processing documents outside of that first folder.

We cannot just remove the `pattern` part because that would make potentially multiple instances of the server (which don't support multi-folder natively) to **each** receive documents for the **whole workspace** (all folders) - resulting in duplicate requests and e.g. duplicate completion items.

Language clients are also immutable after launch, which includes the `documentSelector`.

### How is it being fixed?

Instead of _universally_ reusing the same client, we "optimistically" launch first one which has the "greedy" document selector - so that if the server does support multi-folder, no further action is needed.

If server does _not_ support multiple folders though we will shut down that instance and launch new ones with more narrowly scoped `documentSelector` - each targeting its own folder only.

--- 

While testing this I noticed the following error:

```
rejected promise not handled within 1 second: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
stack trace: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
	at Socket.write (internal/streams/writable.js:292:11)
```

The root cause is basically the upstream LSP library not waiting for `exit` request to be received before closing the channel.
This was fixed in https://github.com/microsoft/vscode-languageserver-node/pull/776 but to my knowledge that error isn't surfaced to the user and doesn't actually affect the user in any way. The server is already able to shut itself down when it receives `EOF` anyway - it just logs that as "dirty" shutdown.

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -23,7 +23,7 @@ export interface TerraformLanguageClient {
 	client: LanguageClient
 }
 
-let clients: Map<string, TerraformLanguageClient> = new Map();
+const clients: Map<string, TerraformLanguageClient> = new Map();
 
 /**
  * ClientHandler maintains lifecycles of language clients
@@ -33,7 +33,7 @@ let clients: Map<string, TerraformLanguageClient> = new Map();
 export class ClientHandler {
 	private shortUid: ShortUniqueId;
 	private pathToBinary: string;
-	private supportsMultiFolders: boolean = true;
+	private supportsMultiFolders = true;
 
 	constructor(private context: vscode.ExtensionContext, private reporter: TelemetryReporter ) {
 		this.shortUid = new ShortUniqueId();
@@ -41,13 +41,13 @@ export class ClientHandler {
 		if (this.pathToBinary) {
 			this.reporter.sendTelemetryEvent('usePathToBinary');
 		} else {
-			let installPath = path.join(context.extensionPath, 'lsp');
+			const installPath = path.join(context.extensionPath, 'lsp');
 			this.pathToBinary = path.join(installPath, 'terraform-ls');
 		}
-	};
+	}
 
 	public StartClients(folders?: string[]): vscode.Disposable[] {
-		let disposables: vscode.Disposable[] = [];
+		const disposables: vscode.Disposable[] = [];
 
 		if (this.supportsMultiFolders) {
 			if (this.GetClient()?.client.needsStart()) {
@@ -57,7 +57,7 @@ export class ClientHandler {
 
 			console.log('Starting client');
 
-			let tfClient = this.newTerraformClient();
+			const tfClient = this.newTerraformClient();
 			tfClient.client.onReady().then(async () => {
 				this.reporter.sendTelemetryEvent('startClient');
 				const multiFoldersSupported = tfClient.client.initializeResult.capabilities.workspace?.workspaceFolders?.supported;
@@ -82,7 +82,7 @@ export class ClientHandler {
 			for (const folder of folders) {
 				if (!clients.has(folder)) {
 					console.log(`Starting client for ${folder}`);
-					let folderClient = this.newTerraformClient(folder);
+					const folderClient = this.newTerraformClient(folder);
 					folderClient.client.onReady().then(() => {
 						this.reporter.sendTelemetryEvent('startClient');
 					});
@@ -95,7 +95,7 @@ export class ClientHandler {
 			}
 		}
 		return disposables;
-	};
+	}
 
 	private newTerraformClient(location?: string): TerraformLanguageClient {
 		const cmd = this.pathToBinary;
@@ -170,7 +170,7 @@ export class ClientHandler {
 			revealOutputChannelOn: 4 // hide always
 		};
 
-		let client = new LanguageClient(
+		const client = new LanguageClient(
 			id,
 			name,
 			serverOptions,
@@ -185,10 +185,10 @@ export class ClientHandler {
 		});
 
 		return {commandPrefix, client}
-	};
+	}
 
 	public async StopClients(folders?: string[]): Promise<void[]> {
-		let promises: Promise<void>[] = [];
+		const promises: Promise<void>[] = [];
 
 		if (this.supportsMultiFolders) {
 			promises.push(this.stopClient(""));
@@ -206,7 +206,7 @@ export class ClientHandler {
 			promises.push(this.stopClient(folder));
 		}
 		return Promise.all(promises);
-	};
+	}
 
 	private async stopClient(folder: string): Promise<void> {
 		if (!clients.has(folder)) {
@@ -221,7 +221,7 @@ export class ClientHandler {
 			}
 			console.log(`Client stopped for ${folder}`);
 		}).then(() => {
-			let ok = clients.delete(folder);
+			const ok = clients.delete(folder);
 			if (ok) {
 				if (folder === "") {
 					console.log('Client deleted');
@@ -238,7 +238,7 @@ export class ClientHandler {
 		}
 
 		return clients.get(this.clientName(document.toString()));
-	};
+	}
 
 	private clientName(folderName: string, workspaceFolders: readonly string[] = sortedWorkspaceFolders()): string {
 		folderName = normalizeFolderName(folderName);
@@ -248,5 +248,5 @@ export class ClientHandler {
 			folderName = getFolderName(getWorkspaceFolder(outerFolder));
 		}
 		return folderName;
-	};
+	}
 }
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -1,0 +1,252 @@
+import * as vscode from 'vscode';
+import ShortUniqueId from 'short-unique-id';
+import {
+	Executable,
+	LanguageClient,
+	LanguageClientOptions,
+	ServerOptions,
+	State,
+	DocumentSelector
+} from 'vscode-languageclient/node';
+import {
+	config,
+	getFolderName,
+	getWorkspaceFolder,
+	normalizeFolderName,
+	sortedWorkspaceFolders
+} from './vscodeUtils';
+import * as path from 'path';
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export interface TerraformLanguageClient {
+	commandPrefix: string,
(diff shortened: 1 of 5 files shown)


src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+			]
+			rootModulePaths = config('terraform-ls', wsFolder).get('rootModules');
+			excludeModulePaths = config('terraform-ls', wsFolder).get('excludeRootModules');
+			outputChannel = vscode.window.createOutputChannel(channelName);
+			outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')} for folder: ${location}`);
+		} else {
+			documentSelector = [
+				{ scheme: 'file', language: 'terraform' },
+				{ scheme: 'file', language: 'terraform-vars' }
+			]
```
radeksimko, line 132: This is the key part where we launch the client with appropriate selector depending on how it's being launched.

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+} from './vscodeUtils';
+import * as path from 'path';
+import TelemetryReporter from 'vscode-extension-telemetry';
+
+export interface TerraformLanguageClient {
+	commandPrefix: string,
+	client: LanguageClient
+}
+
+let clients: Map<string, TerraformLanguageClient> = new Map();
```
ansgarm, line 26:
```suggestion
const clients: Map<string, TerraformLanguageClient> = new Map();
```
It is common to use const when the variable is never reassigned. The project contains a config for eslint that would also error for this (it is installed as a dependency but not run in scripts).
You can run it via `./node_modules/.bin/eslint .` or via `yarn eslint .` (if you have yarn installed – as yarn falls back to the `.bin` directory where npm complains instead).

radeksimko: Thanks for pointing it out - I wasn't even aware we had a linter set up.

I already installed VS Code ESLint extension and fixed all the errors and any warnings in the new code. I also added a check to CI.

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+		this.pathToBinary = config('terraform').get('languageServer.pathToBinary');
+		if (this.pathToBinary) {
+			this.reporter.sendTelemetryEvent('usePathToBinary');
+		} else {
+			const installPath = path.join(context.extensionPath, 'lsp');
+			this.pathToBinary = path.join(installPath, 'terraform-ls');
+		}
+	}
+
+	public StartClients(folders?: string[]): vscode.Disposable[] {
```
ansgarm, line 49:
```suggestion
	public startClients(folders?: string[]): vscode.Disposable[] {
```
TypeScript style would be keeping a lowercase letter 🙂

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+			if (event.newState === State.Stopped) {
+				clients.delete(location);
+				this.reporter.sendTelemetryEvent('stopClient');
+			}
+		});
+
+		return {commandPrefix, client}
+	}
+
+	public async StopClients(folders?: string[]): Promise<void[]> {
```
ansgarm, line 190: "lowerCamelCase" here as well :)

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+				if (folder === "") {
+					console.log('Client deleted');
+					return
+				}
+				console.log(`Client deleted for ${folder}`);
+			}
+		});
+	}
+
+	public GetClient(document?: vscode.Uri): TerraformLanguageClient {
```
ansgarm, line 235: [same](https://youtu.be/IYH7_GzP4Tg?t=26) here one last time :)

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+					// restart is needed to launch folder-focused instances
+					console.log('Restarting clients as folder-focused');
+					await this.StopClients(folders);
+					this.supportsMultiFolders = false;
+					this.StartClients(folders);
+				}
+			});
+
+			disposables.push(tfClient.client.start());
+			clients.set("", tfClient);
```
ansgarm, line 76: I like the idea of using an empty String. Maybe use a constant for that?
Like e.g. `const DEFAULT_CLIENT="";` or `SINGLE_CLIENT` or `MULTI_CLIENT` or whatever makes the most sense.

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+			console.log('Starting client');
+
+			const tfClient = this.newTerraformClient();
+			tfClient.client.onReady().then(async () => {
+				this.reporter.sendTelemetryEvent('startClient');
+				const multiFoldersSupported = tfClient.client.initializeResult.capabilities.workspace?.workspaceFolders?.supported;
+				console.log(`Multi-folder support: ${multiFoldersSupported}`);
+
+				if (!multiFoldersSupported) {
+					// restart is needed to launch folder-focused instances
```
ansgarm, line 67: I'm probably lacking knowIedge here: Is this also happening if only a single workspace folder is used? (If yes: Is a restart required in that case?)

src/clientHandler.ts
ansgarm: Looks good 👍

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+				{ scheme: 'file', language: 'terraform-vars' }
+			]
+			rootModulePaths = config('terraform-ls').get('rootModules');
+			excludeModulePaths = config('terraform-ls').get('excludeRootModules');
+			outputChannel = vscode.window.createOutputChannel(channelName);
+			outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
+		}
+
+		if (rootModulePaths.length > 0 && excludeModulePaths.length > 0) {
+			throw new Error('Only one of rootModules and excludeRootModules can be set at the same time, please remove the conflicting config and reload');
```
ansgarm, line 140: I personally wouldn't know how to reload 😄 Is there a restart language server command?

src/clientHandler.ts
```diff
@@ -0,0 +1,252 @@
...
+		const clientOptions: LanguageClientOptions = {
+			documentSelector: documentSelector,
+			workspaceFolder: wsFolder,
+			initializationOptions: initializationOptions,
+			initializationFailedHandler: (error) => {
+				this.reporter.sendTelemetryException(error);
+				return false;
+			},
+			outputChannel: outputChannel,
+			revealOutputChannelOn: 4 // hide always
```
ansgarm, line 170: There seems to be `RevealOutputChannelOn.Never` for that ([example](https://github.com/microsoft/vscode-extension-samples/blob/main/lsp-user-input-sample/client/src/extension.ts#L21)). ([Source](https://github.com/microsoft/vscode-languageserver-node/blob/0cb3812e7d540ef3a904e96df795bc37a21de9b0/client/src/common/client.ts#L358))

ansgarm (commented):
Only some code-style stuff to make it look less like Go and a question 🚀

src/clientHandler.ts
radeksimko: Good catch!

src/clientHandler.ts
radeksimko: Good idea - I will create `MULTI_FOLDER_CLIENT`.

src/clientHandler.ts
radeksimko: No restart is needed because the client/server is already (optimistically) being launched with the right parameters for multi-folder. These parameters just aren't relevant for folder-specific instances. I tried to describe it more verbosely in the PR description above.

src/clientHandler.ts
radeksimko: So the server isn't even launched at this point - this would be reload of VS Code and interestingly we already have some logic to prompt the reload with a single button upon any change to the relevant parts of the configuration:
https://github.com/hashicorp/vscode-terraform/blob/583bdf59a19d46a0f5e2325f0d7a11379e0bd337/src/extension.ts#L108-L116

Now that you mention it though it looks like that the exception just blocks the process of creating a client and is never shown to the user - which is not ideal. That said this is not something being introduced as part of this PR, that bug was there for some time now (apparently since this error handling was introduced), so we can deal with that in a separate PR.

ansgarm (approved):
:shipit:

PR #691, opened 2021-07-16
Improve error handling

### LINKED ISSUE #681: Return human-friendly error when LS is not found
## Problem Statement

When the extension cannot find the language server, it returns a vague error message:

> Couldn't start client Language Server: file:///...

Screenshot 2021-06-29 at 14 40 54

## Reproduction

Use e.g. the following settings to point it to a non-existing path:

```json
    "terraform.languageServer": {
        "pathToBinary": "/bla/blah/terraform-ls",
        "args": [
            "serve",
        ]
    }
```

## Expected Behaviour

A human-readable error message, such as

```
Language server not found at /bla/blah/terraform-ls (provided via "pathToBinary" config option)
```

radeksimko:
Depends on #696 

Fixes #681

--- 

As discussed in https://github.com/hashicorp/vscode-terraform/pull/688#discussion_r670994631 we would previously try to communicate conflicting config options via an exception, which was never caught anywhere. In case user happened to set the two conflicting options the server would just (silently) never launch and they wouldn't receive any UI feedback.

With this PR that error is now surfaced correctly as error popup:

Screenshot 2021-07-16 at 17 50 07

--- 

Additionally this PR also improves error handling in cases where path to LS is for some reason wrong or inaccessible. The error surfaced by the upstream library would otherwise look like this:

Screenshot 2021-07-16 at 17 40 19

so we make it **more helpful** by surfacing the real error messages and adding a hint in cases where user has specified the path themselves:

Screenshot 2021-07-16 at 17 41 55
Screenshot 2021-07-16 at 17 42 49

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "vscode-extension-telemetry": "^0.1.7",
         "vscode-languageclient": "^7.0.0-next.12",
         "vscode-uri": "^3.0.2",
+        "which": "^2.0.2",
         "yauzl": "^2.10.0"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


ansgarm (commented):
Nice work 👍 I only have the same feedback such as in #696 but as soon as that is addressed, this is fine to go 🚀

package-lock.json
```diff
@@ -18,6 +18,7 @@
         "vscode-extension-telemetry": "^0.1.7",
         "vscode-languageclient": "^7.0.0-next.12",
         "vscode-uri": "^3.0.2",
+        "which": "^2.0.2",
```
radeksimko, line 21: I'm not sure why does the addition of this single dependency cause such a messy diff 🤷🏻

ansgarm: Maybe there were outdated things in the Lockfile (so it was not consistent with the dependencies in package.json anymore). But I don't know 😄

ansgarm (approved):

PR #694, opened 2021-07-19
fix: launch LS even if path contains escapable characters

radeksimko:
Unblocks #691 

--- 

This fixes a long undiscovered bug which would cause the LS to never launch
in situation where there is an escapable character (such as whitespace)
in the *default* path leading to the LS binary.

The root cause is string concatenation of full commands (and
`child_process.exec` API expecteding already escaped string)
when checking LS version, instead of passing arguments individually.

This problem would also never be surfaced to the user due to some other
issues with error reporting (that can be addressed in a separate PR).

### Users Affected

This affects all v2 versions from what I can tell (based on the codepath).

### Reproduction

First time launch after fresh installation (or upgrade) on Windows with account called `Radek Simko` (with whitespace) + empty settings:

Screenshot 2021-07-19 at 16 10 34

After restarting VS Code:
Screenshot 2021-07-19 at 16 12 12

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -119,13 +119,13 @@ async function getLsVersion(dirPath: string): Promise<string> {
 	fs.accessSync(binPath, fs.constants.X_OK);
 
 	try {
-		const jsonCmd: { stdout: string } = await exec(`${binPath} version -json`);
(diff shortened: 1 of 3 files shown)


ansgarm (approved):
Ah nice! I think I need to write down a todo to check that we also do this everywhere in the CDKTF 😄

PR #696, opened 2021-07-20
Extract logic for determining LS paths

radeksimko:
Unblocks #691 

--- 

Path to either the LS binary or to the installation folder is needed in multiple places and sadly isn't as trivial. This extraction ensures that the same mistakes (such as forgetting to reflect Windows exe extensions or path separators) aren't repeated.

Windows would normally still find and execute a binary even without the "exe" extension so this bug shouldn't be affecting Windows users at this point, but some further checks (testing file existence/permissions) will be added in the above PR and that requires knowing the real binary name.

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -4,20 +4,15 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
 import TelemetryReporter from 'vscode-extension-telemetry';
-
 import { exec } from './utils';
+import { ServerPath } from './serverPath';
 import { Release, getRelease } from '@hashicorp/js-releases';
 
 const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
 
-interface pathGetter {
-	installPath(): string;
-	binPath(): string
-}
-
 export class LanguageServerInstaller {
 	constructor(
-		private lsPath: pathGetter,
+		private lsPath: ServerPath,
 		private reporter: TelemetryReporter
 	) { }
 
@@ -120,7 +115,7 @@ export class LanguageServerInstaller {
 	}
 }
 
-async function getLsVersion(lsPath: pathGetter): Promise<string> {
+async function getLsVersion(lsPath: ServerPath): Promise<string> {
 	const binPath = lsPath.binPath();
 	fs.accessSync(binPath, fs.constants.X_OK);
 
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -10,9 +10,14 @@ import { Release, getRelease } from '@hashicorp/js-releases';
 
 const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
 
+interface pathGetter {
+	installPath(): string;
+	binPath(): string
+}
+
 export class LanguageServerInstaller {
 	constructor(
-		private directory: string,
+		private lsPath: pathGetter,
 		private reporter: TelemetryReporter
 	) { }
 
@@ -30,7 +35,7 @@ export class LanguageServerInstaller {
 
 		let installedVersion: string;
 		try {
-			installedVersion = await getLsVersion(this.directory);
+			installedVersion = await getLsVersion(this.lsPath);
 		} catch (err) {
 			// Most of the time, getLsVersion will produce "ENOENT: no such file or directory"
 			// on a fresh installation (unlike upgrade). It’s also possible that the file or directory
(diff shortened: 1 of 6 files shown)


src/vscodeUtils.ts
```diff
@@ -1,4 +1,42 @@
 import * as vscode from 'vscode';
+import * as path from 'path';
+
+export class ServerPath {
+	private folderName = 'lsp';
+	private customBinPath: string;
+	public customBinPathOptionName = 'languageServer.pathToBinary';
```
ansgarm, line 7: I suspect this is not going to change dynamically at runtime any time soon?
I think it would be more common to put this as
`const CUSTOM_BIN_PATH_OPTION_NAME = 'languageServer.pathToBinary';`
above the `export class ServerPath {` line 🙂

ansgarm: Same probably for `folderName`.

src/utils.ts
```diff
@@ -12,25 +11,6 @@ export function exec(cmd: string, args: readonly string[]): Promise<{ stdout: st
 	});
 }
 
-export function httpsRequest(url: string, options: https.RequestOptions = {}, encoding = 'utf8'): Promise<string> {
```
ansgarm, line 15: Was that old unused code?

src/languageServerInstaller.ts
```diff
@@ -10,9 +10,14 @@ import { Release, getRelease } from '@hashicorp/js-releases';
 
 const extensionVersion = vscode.extensions.getExtension('hashicorp.terraform').packageJSON.version;
 
+interface pathGetter {
```
ansgarm, line 13: Interfaces starting with a lower-case letter are [uncommon](https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#interface)

ansgarm: Maybe also move this interface next to the implementation of the PathGetter (= `ServerPath`) to be exported and reused from there? (It is duplicated at another place as well).
I personally would also be fine with having no interface and just using `ServerPath` as the type.

src/utils.ts
radeksimko: Yep, see the commit message - I pushed it intentionally in a separate one 😉

src/languageServerInstaller.ts
radeksimko: Happy to rename it to start with upper-case 👍🏻 

For the interface - this is actually intentional and I don't see those as duplicates?

The `ServerPath` implementation implements _all_ of the methods/attributes and then the installer only requires subset of those, and `ClientHandler` requires different subset - hence they each have their own declaration of the interface depending on the needs.

Is this not how you'd use interfaces in TypeScript? It is a relatively common pattern in Go to reduce the exposure and limit responsibilities.

src/vscodeUtils.ts
radeksimko: Yeah, that's fair - I had the tendency to do it actually. The only reason I didn't is because constants seem to be scoped to the whole file in TypeScript, not a class AFAICT?

I could extract the class into its own file (`serverPath.ts`) I suppose - I just don't like the idea of a seemingly random constant lingering in a file called "utils" 😂 😭  but I haven't had the time yet to break down the logic here and get rid of "utils".

src/languageServerInstaller.ts
radeksimko: That said its seems like a minor change with the size of the class anyway - and if I extract it into its own file then that addresses another concern I had - which is dependency cycles - so yeah - I can avoid interfaces altogether.

src/utils.ts
ansgarm: Ah, I did not look at the commits individually 😅

src/languageServerInstaller.ts
ansgarm: Yeah, I suspected something like that. My personal style would've just been to keep it simple that's why I commented on that :) I did not notice that the subset differed – I just glanced over the naming and suspected them to be the same.
In the end I'm happy with either way 👍

src/vscodeUtils.ts
ansgarm: Ah, I see that reasoning 👍 You could also add a `readonly` modifier and change it to a `static` to it to make it more of a constant value.
```suggestion
	static readonly customBinPathOptionName = 'languageServer.pathToBinary';
```
I think you'd also have to adjust the places it's used and replace `this` with the name of the class then.

radeksimko: I extracted it into its own file. 👍🏻

ansgarm (approved):

PR #700, opened 2021-07-22
Create command contribution for terraform apply

### LINKED ISSUE #648: Implement `Terraform: Apply` as a command
## Problem Statement

Terraform allows the user to apply changes to their infrastructure via `terraform apply`, which is likely the most commonly used command.

Users would benefit from being able to apply changes directly from the editor.

## Technical Details

The extension currently implements the following:

 - `terraform init`
   - proxied to Terraform LS: https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#terraforminit
 - `terraform plan`
   - executed directly in the integrated terminal
 - `terraform validate`
   - proxied to Terraform LS: https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#terraformvalidate

The difference in how each command is integrated relate to how the output/outcome is mappable to LSP and how it can be communicated to the user.

## Expected User Experience

User is able to open a command palette and choose "Terraform: Apply", similar to how they can do that for `terraform plan` currently:

Screenshot 2021-05-24 at 12 11 46

Then user will be prompted to confirm the action via prompt.

Then the user will see the `apply` command executed in the VS Code terminal, similar to `terraform plan`:

Screenshot 2021-05-24 at 12 13 00

## Proposal

Implement `terraform apply` as command executable from the command palette.

It may be easiest to just implement it initially so that it executes in the integrated terminal. We can later assess whether any integration with LS is useful.

kamilturek:
Closes #648.
---
Command `terraform.apply` is already registered so this PR only adds it to extension manifest.
https://github.com/hashicorp/vscode-terraform/blob/b5c8b6c7855c98918b6fe4d2fea0f11d1c438599/src/extension.ts#L66-L68
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -194,6 +194,10 @@
         "title": "Terraform: validate"
       },
       {
+        "command": "terraform.apply",
+        "title": "Terraform: Apply"
+      },
+      {
         "command": "terraform.plan",
         "title": "Terraform: plan"
       }


jpogran:
Thanks @kamilturek for the contribution! Apologies on the delay in response, but we appreciate the help.

The reason we haven't surfaced the existing `terraform apply` command yet is that it will open the current terminal and just paste the command there. We haven't decided on the best UX here yet. If we enable it as is, it won't know if the command succeeded or failed, if terraform was on the path or not, if it was in the correct folder, etc.

What are your thoughts on this? While we discuss, I'm going to mark this as draft and put it on our planning board.

PR #741, opened 2021-08-15
src/clientHandler: add support for certain CLI options

danishprakash:
* terraformExecPath
* terraformExecTimeout
* terraformExecLogFilePath

Closes (partially) https://github.com/hashicorp/terraform-ls/issues/240

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,21 @@
           "default": [],
           "description": "Per-workspace list of module directories for the language server to read"
         },
+        "terraform-ls.terraformExecPath": {
+          "scope": "resource",
+          "type": "string",
+          "description": "Path to the Terraform binary"
+        },
+        "terraform-ls.terraformExecTimeout": {
+          "scope": "resource",
+          "type": "string",
+          "description": "Overrides Terraform execution timeout (e.g. 30s)"
+        },
+        "terraform-ls.terraformLogFilePath": {
+          "scope": "resource",
+          "type": "string",
+          "description": "Path to a file for Terraform executions to be logged into (TF_LOG_PATH) with support for variables (e.g. Timestamp, Pid, Ppid) via Go template syntax {{.VarName}}"
+        },
         "terraform-ls.excludeRootModules": {
           "scope": "resource",
           "type": "array",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -149,6 +149,24 @@
           "default": [],
           "description": "Per-workspace list of module directories for the language server to read"
         },
+        "terraform-ls.terraformExecPath": {
+          "scope": "resource",
+          "type": "string",
+          "default": [],
+          "description": "Path to Terraform binary"
```
radeksimko, line 156:
```suggestion
          "description": "Path to the Terraform binary"
```

package.json
```diff
@@ -149,6 +149,24 @@
...
+          "scope": "resource",
+          "type": "string",
+          "default": [],
+          "description": "Overrides Terraform execution timeout (e.g. 30s)"
+        },
+        "terraform-ls.terraformExecLogFilePath": {
+          "scope": "resource",
+          "type": "string",
+          "default": [],
+          "description": "Path to a file for Terraform executions to be logged into with support for variables (e.g. Timestamp, Pid, Ppid) via Go template syntax {{.VarName}}"
```
radeksimko, line 168:
```suggestion
          "description": "Path to a file for Terraform executions to be logged into (TF_LOG_PATH) with support for variables (e.g. Timestamp, Pid, Ppid) via Go template syntax {{.VarName}}"
```

package.json
```diff
@@ -149,6 +149,24 @@
...
+        "terraform-ls.terraformExecTimeout": {
+          "scope": "resource",
+          "type": "string",
+          "default": [],
+          "description": "Overrides Terraform execution timeout (e.g. 30s)"
+        },
+        "terraform-ls.terraformExecLogFilePath": {
+          "scope": "resource",
+          "type": "string",
+          "default": [],
```
radeksimko, line 167: The default type seems wrong (array) for string arguments but more importantly I don't think we need any defaults for these options in particular? Do you mind removing the defaults?

danishprakash: You're right, didn't notice that while copying over this block from an existing piece. Updated.

radeksimko (approved):
LGTM, thanks.

PR #745, opened 2021-08-27
(maint) Set syntax and format style guidelines

jpogran:
Update eslint and prettier packages and add style and format guidelines to editorconfig and VS Code worskpace settings file.

This aligns to company standards for code style and formatting and ensures consistent spacing, line endings, and file encodings. This does not enable any new eslint rules, as that is out of scope for this line of work.

This also configures VS Code to use prettier and eslint for all typescript, javascript and json formatting and use eslint and prettier to enforce guidelines.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  "endOfLine": "lf",
+  "printWidth": 120,
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false
+}
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.{diff,md}]
+trim_trailing_whitespace = false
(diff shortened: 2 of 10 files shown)


.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "endOfLine": "lf",
+  "printWidth": 120,
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "all",
+  "useTabs": false
+}
```
ansgarm, line 9: `endOfLine`, `semi`, `tabWidth` and `useTabs` could be omitted as the selected settings match the [defaults](https://prettier.io/docs/en/options.html) already

ansgarm (approved):
For the CDK for Terraform we chosed to use the Prettier defaults [without any adjustments](https://github.com/hashicorp/terraform-cdk/blob/main/.prettierrc.json) to leave discussions about preferred code style to those working on Prettier.

I personally am happy with most things as long as they're consistent in the codebase, so me mentioning this is more of a sidenote 😄 If you e.g. like `'` more than `"` go for it – you will have to read that code daily 😄

.prettierrc
jpogran: That is a good point, and one I generally hew towards. With the codebase having a mix of all of these things (sometimes in the same file) I leaned toward being explicit instead of implicit for clarity.  This way when each file is formatted as we add things down the line, it's still clear where the rules are coming from.

jpogran:
Thanks @ansgarm for the review!

.prettierrc
ansgarm: Oh, you're right! Explicit beats implicit 👍

PR #755, opened 2021-08-31
TCP Client Handler

jpogran:
Add TCP connection settings

This commit adds `terraform.experimentalFeatures.languageServer.tcp.port` to enable connecting to a terraform-ls instance using TCP instead of STDIO. The terraform-ls instance has to be already running, this will not start it for the user.

This also adds a custom error handler to the LanguageClient implementation. This custom implementation attempts to start the LS and returns after one failure, but ensures an error window is raised when it fails to start or connect to a terraform-ls instance. This is needed because the LanguageClient handles errors by default but relies on the OutputChannel not being hidden to show errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -310,6 +310,16 @@
             ],
             "description": "Arguments to pass to language server binary"
           },
+          "terraform.languageServer.tcp.port": {
+            "order": 4,
+            "type": [
+              "number",
+              null
+            ],
+            "scope": "machine-overridable",
+            "default": null,
+            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running language server process."
+          },
           "terraform.languageServer.ignoreSingleFileWarning": {
             "order": "3",
             "scope": "window",
@@ -435,15 +445,6 @@
             "scope": "window",
             "type": "boolean",
             "default": false
-          },
-          "terraform.experimentalFeatures.languageServer.tcp.port": {
-            "type": [
-              "number",
-              null
-            ],
-            "scope": "machine-overridable",
-            "default": null,
-            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running language server process."
           },
           "terraform-ls.experimentalFeatures": {
             "scope": "window",
--- a/package.json
+++ b/package.json
@@ -443,7 +443,7 @@
             ],
             "scope": "machine-overridable",
             "default": null,
-            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running Language Server is already running Language Server process."
+            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running language server process."
           },
           "terraform-ls.experimentalFeatures": {
             "scope": "window",
--- a/package.json
+++ b/package.json
@@ -436,6 +436,15 @@
             "type": "boolean",
             "default": false
           },
+          "terraform.experimentalFeatures.languageServer.tcp.port": {
+            "type": [
+              "number",
+              null
+            ],
+            "scope": "machine-overridable",
+            "default": null,
+            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running Language Server is already running Language Server process."
+          },
           "terraform-ls.experimentalFeatures": {
             "scope": "window",
             "type": "object",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -38,7 +38,7 @@ const documentSelector: DocumentSelector = [
   { scheme: 'file', language: 'terraform' },
   { scheme: 'file', language: 'terraform-vars' },
 ];
-const outputChannel = vscode.window.createOutputChannel(brand);
+export const outputChannel = vscode.window.createOutputChannel(brand);
 export let terraformStatus: vscode.StatusBarItem;
 
 let reporter: TelemetryReporter;
@@ -112,7 +112,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     reporter.sendTelemetryEvent('usePathToBinary');
   }
   const serverOptions: ServerOptions = await getServerOptions(lsPath);
-  // outputChannel.appendLine(`Launching language server: ${executable.command} ${executable.args?.join(' ')}`);
 
   const initializationOptions = getInitializationOptions();
   const clientOptions: LanguageClientOptions = {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -12,7 +12,7 @@ import {
   ServerOptions,
 } from 'vscode-languageclient/node';
 import { Utils } from 'vscode-uri';
-import { clientSupportsCommand, getInitializationOptions, getServerExecutable } from './utils/clientHelpers';
+import { clientSupportsCommand, getInitializationOptions, getServerOptions } from './utils/clientHelpers';
 import { GenerateBugReportCommand } from './commands/generateBugReport';
 import { ModuleCallsDataProvider } from './providers/moduleCalls';
 import { ModuleProvidersDataProvider } from './providers/moduleProviders';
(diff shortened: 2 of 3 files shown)


jpogran:
@radeksimko I agree there should be some sort of validation of settings, but in this case we won't be able rely on whether `args` or `pathToBinary` are specified with `tcp_port`. Both `args` and `pathToBinary` both provide defaults, and so are always specified.

To address this we could warn if `args` and `pathToBinary` both have the default values and `tcp_port` is specified. This requires us storing the default values in code as well as in the `package.json` manifest. This seems prone to error and fragile, but provides the most notification to the user of what is an error state.

An alternative is to leave this as connecting if there is a `tcp_port` set, and have VS Code fail and show the normal error message. We then edit the `tcp_port` description to say that this will override any other setting.

radeksimko:
I see, I didn't realize the trouble with default settings 😢 

In that case I'm ok with no validation, but I still feel like we should improve the UX when `tcp_port` is forgotten in the config and server is no longer running. Currently it seems to just "silently" error into the Output pane:

Screenshot 2021-09-14 at 19 04 21


Could we perhaps raise an exception or something which causes the error to be more visible in a popup or something?

package.json
```diff
@@ -436,6 +436,15 @@
...
             "default": false
           },
+          "terraform.experimentalFeatures.languageServer.tcp.port": {
+            "type": [
+              "number",
+              null
+            ],
+            "scope": "machine-overridable",
+            "default": null,
+            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running Language Server is already running Language Server process."
```
radeksimko, line 446:
```suggestion
            "markdownDescription": "Language server TCP port to connect to. This is not compatible with `#terraform.languageServer.path#`. This is used when you want the extension to connect via TCP to an already running language server process."
```

src/extension.ts
```diff
@@ -111,12 +111,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
     reporter.sendTelemetryEvent('usePathToBinary');
   }
-  const executable = await getServerExecutable(lsPath);
-  const serverOptions: ServerOptions = {
-    run: executable,
-    debug: executable,
-  };
-  outputChannel.appendLine(`Launching language server: ${executable.command} ${executable.args?.join(' ')}`);
+  const serverOptions: ServerOptions = await getServerOptions(lsPath);
+  // outputChannel.appendLine(`Launching language server: ${executable.command} ${executable.args?.join(' ')}`);
```
radeksimko, line 115:
```suggestion
```

src/utils/clientHelpers.ts
```diff
@@ -1,18 +1,53 @@
...
+    const inspect = vscode.workspace.getConfiguration('terraform').inspect('languageServer.path');
+    if (inspect !== undefined) {
+      ///
+      if (inspect.globalValue || inspect.workspaceFolderValue || inspect.workspaceValue) {
+        vscode.window.showWarningMessage(
+          'You cannot use experimentalFeatures.languageServer.tcp.port with terraform.languageServer.path. Ignoring terraform.languageServer.path and proceeding to connect via TCP',
+        );
+      }
+    }
+
```
radeksimko, line 21:
```suggestion
    outputChannel.appendLine(`Connecting to language server via TCP at localhost:%d`, port);
```
I guess we'd need to get the `outputChannel` somewhere down here, but I'd say it's worth doing it as this particular line proved useful for debugging in the past - i.e. knowing how the server was launched, with what arguments etc.

src/utils/clientHelpers.ts
```diff
@@ -1,18 +1,53 @@
...
+      };
+      return Promise.resolve(result);
+    };
+    return serverOptions;
+  }
+
   const cmd = await lsPath.resolvedPathToBinary();
   const serverArgs = config('terraform').get<string[]>('languageServer.args', []);
-
+  // this.outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
```
radeksimko, line 39:
```suggestion
  outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
```
I guess we'd need to get the outputChannel somewhere down here, but I'd say it's worth doing it as this particular line proved useful for debugging in the past - i.e. knowing how the server was launched, with what arguments etc.

src/utils/clientHelpers.ts
```diff
@@ -1,18 +1,53 @@
...
 
-export async function getServerExecutable(lsPath: ServerPath): Promise<Executable> {
+export async function getServerOptions(lsPath: ServerPath): Promise<ServerOptions> {
+  let serverOptions: ServerOptions;
+
+  const port: number | undefined = config('terraform').get('experimentalFeatures.languageServer.tcp.port');
+  if (port) {
+    const inspect = vscode.workspace.getConfiguration('terraform').inspect('languageServer.path');
+    if (inspect !== undefined) {
+      ///
```
radeksimko, line 14:
```suggestion
```

package.json
```diff
@@ -436,6 +436,15 @@
             "type": "boolean",
             "default": false
           },
+          "terraform.experimentalFeatures.languageServer.tcp.port": {
```
radeksimko, line 439: Is there a particular reason you want to roll this out as experimental? At least the code change looks relatively straight-forward and I wouldn't expect many people actually end up using it for day2day development anyway, so there's already reduced risk/exposure from that angle.

radeksimko (approved):
I'm frankly not sure if passing the `OutputChannel` like I proposed is the right way to do logging (forgive me my lack of TypeScript conventions here 😅 ), or whether the underlying code should even do logging like that, but I do think we should be logging the fact that we're connecting via TCP and what port or how we're launching the server (with what arguments).

Similarly it would be useful if we also logged successful connection, but I understand that could require some reshuffling of code probably.

Minor in-line suggestions/question aside, this LGTM!

src/utils/clientHelpers.ts
```diff
@@ -1,18 +1,55 @@
...
+import * as vscode from 'vscode';
+import { Executable, InitializeResult, ServerOptions, StreamInfo } from 'vscode-languageclient/node';
 import { config } from './vscode';
 import { ServerPath } from './serverPath';
+import { outputChannel } from '../extension';
+
+export async function getServerOptions(lsPath: ServerPath): Promise<ServerOptions> {
+  let serverOptions: ServerOptions;
+
+  const port: number | undefined = config('terraform').get('experimentalFeatures.languageServer.tcp.port');
```
dbanck, line 11:
```suggestion
  const port = config('terraform').get<number>('experimentalFeatures.languageServer.tcp.port');
```

We can pass a type to `get`, so we don't have to hard-code it.

src/extension.ts
```diff
@@ -111,12 +111,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
   if (lsPath.hasCustomBinPath()) {
     reporter.sendTelemetryEvent('usePathToBinary');
   }
-  const executable = await getServerExecutable(lsPath);
-  const serverOptions: ServerOptions = {
-    run: executable,
-    debug: executable,
-  };
-  outputChannel.appendLine(`Launching language server: ${executable.command} ${executable.args?.join(' ')}`);
+  const serverOptions: ServerOptions = await getServerOptions(lsPath);
```
dbanck, line 117:
```suggestion
  const serverOptions = await getServerOptions(lsPath);
```

TS can infer the return type from `getServerOptions` here

package.json
jpogran: I was were trying to imply this wasn't a primary workflow we supported for every day use. If we think it's fine alongside the primary workflow I'm ok with removing the experimental part

src/utils/clientHelpers.ts
```diff
@@ -1,18 +1,56 @@
...
+    if (inspect !== undefined) {
+      ///
+      if (inspect.globalValue || inspect.workspaceFolderValue || inspect.workspaceValue) {
+        vscode.window.showWarningMessage(
+          'You cannot use experimentalFeatures.languageServer.tcp.port with terraform.languageServer.path. Ignoring terraform.languageServer.path and proceeding to connect via TCP',
+        );
+      }
+    }
+
+    serverOptions = () => {
```
dbanck, line 23:
```suggestion
    serverOptions = async () => {
```

If we make this function `async` we can avoid the manual creation of a promise. We could even return a bit earlier:
```ts
return {
  writer: socket,
  reader: socket,
};
```

dbanck (commented):
Nice work! I added some small comments with regards to TS types

This review doesn't include the just added `errorHandler`. I'll take a look at it later

src/utils/clientHelpers.ts
jpogran: This PR has been open so long, I've forgotten parts!

Your suggestions work because of setting `"type" :["number",null ]` and `"default": null` for `terraform.experimentalFeatures.languageServer.tcp.port`. If you remove those, it will default to 0 and cause the behavior I describe.

So, leaving the setting attributes up and above and your suggestions works together.

dbanck (approved):
Nice work! 👍

PR #757, opened 2021-09-03
Improve First Use Experience

jpogran:
When the extension starts, if there isn't a `.terraform` folder in the current workspace, this warning will appear:

image

Clicking on `Run terraform init` opens the command palette with the init command pre-populated.

image

Clicking on `More Information` will open a browser window to https://www.terraform.io/docs/cli/commands/init.html.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,3 +1,5 @@
+import * as fs from 'fs';
+import * as path from 'path';
 import * as vscode from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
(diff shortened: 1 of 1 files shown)


jpogran:
As discussed in the linked issue, there was a popup message informing the user that terraform init was needed. This proved to be distracting and undesirable, as well as being unable to be quieted/ignored. While this PR does have the ability to set a value to remember not to show the message, there isn't a strong set of use cases where we *need* to perform this action.

Closing this in favor of other initiatives.

PR #771, opened 2021-09-14
fix: avoid considering output panes as editors

### LINKED ISSUE #744: URI "output:extension-output-hashicorp.terraform-%231" is not valid
## Versions

This bug is reproducible in:
 - [X] the **latest** version of the extension (below)
 - [X] the **latest** version of the language server (below)

### Extension

```
Released on
2/12/2016, 03:31:09
Last updated
7/22/2021, 10:24:21
Identifier
hashicorp.terraform
```

### Language Server

```
Output too large, see attached
```
[terraform-ls-debug.out.tar.gz](https://github.com/hashicorp/vscode-terraform/files/7061218/terraform-ls-debug.out.tar.gz)

### VS Code

```
Version: 1.59.1
Commit: 3866c3553be8b268c8a7f8c0482c0c0177aa8bfa
Date: 2021-08-19T11:53:52.479Z (1 wk ago)
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
OS: Darwin x64 19.6.0
```

### Operating System

```
OS X Catalina 10.15.7
Darwin (redacted) 19.6.0 Darwin Kernel Version 19.6.0: Tue Jun 22 19:49:55 PDT 2021; root:xnu-6153.141.35~1/RELEASE_X86_64 x86_64
```

### Terraform Version

```
Terraform v0.15.3
on darwin_amd64

```

radeksimko:
Fixes #744

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -281,6 +281,11 @@ export class ClientHandler {
     return clients.get(this.clientName(document.toString()));
   }
 
+  public clientSupportsCommand(cmdName: string, document?: vscode.Uri): boolean {
+    const commands = this.getClient(document).client.initializeResult.capabilities.executeCommandProvider?.commands;
+    return commands.includes(cmdName);
+  }
+
   private clientName(folderName: string, workspaceFolders: readonly string[] = sortedWorkspaceFolders()): string {
     folderName = normalizeFolderName(folderName);
     const outerFolder = workspaceFolders.find((element) => folderName.startsWith(element));
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -123,7 +123,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     }),
     vscode.window.onDidChangeVisibleTextEditors(async () => {
       const textEditor = getActiveTextEditor();
-      await updateTerraformStatusBar(textEditor.document.uri, lsPath);
+      await updateTerraformStatusBar(textEditor.document.uri);
     }),
   );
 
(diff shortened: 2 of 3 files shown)


src/clientHandler.ts
```diff
@@ -279,6 +279,11 @@ export class ClientHandler {
     return clients.get(this.clientName(document.toString()));
   }
 
+  public clientSupportsCommand(cmdName: string, document?: vscode.Uri): boolean {
+    const commands = this.getClient(document).client.initializeResult.capabilities.executeCommandProvider?.commands;
```
jpogran, line 283: In each of these `.client` calls, there is a possible exception where the extension is activated, the editor is activated, but the client is not ready yet because the language client and LS haven't initialized yet.

If we use the `onReady` method and chain to `.then()` and await the result we can be reasonably sure we won't throw exceptions.

This will be an infrequent case because our LS starts so quickly, but I encountered this when developing the module view. If a user has an active window/panel open that triggers this before the LS has started and is ready, then an error is shown to the client that they don't know what to do about.

radeksimko: raised a separate PR with patch for this https://github.com/hashicorp/vscode-terraform/pull/778 then I can rebase this one, hopefully retaining the same ClientHandler API 🤞🏻

radeksimko: Rebased this PR - PTAL

jpogran (approved):
Everything is an editor didn't ya know

PR #776, opened 2021-09-15
ci: Read node version from nvmrc

radeksimko:
This is to make future updates of Node.js easier, to prepare for automation of Marketplace publishing and also to align version management with what we do in LS and elsewhere:

https://github.com/hashicorp/terraform-ls/blob/356136b9444ecf2274b9d28947f18e8a71f9fd0e/.github/workflows/ci.yml#L70-L90

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,15 +7,22 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     steps:
       - name: checkout repo
         uses: actions/checkout@v2
 
+      - name: Read NodeJS version
+        id: nodejs-version
+        run: |
+          content=`cat ./.nvmrc`
+          echo "::set-output name=content::$content"
+
       - name: setup node
         uses: actions/setup-node@v1
         with:
-          node-version: '12.x'
+          # See https://github.com/actions/setup-node/issues/32
+          node-version: ${{ steps.nodejs-version.outputs.content }}
 
       - name: clean install dependencies
         run: npm ci    
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,10 +15,17 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
-      - name: Use Node.js 14.x
+      -
+        name: Read Node.js version (Unix)
+        id: nodejs-version
+        run: |
+          content=`cat ./.nvmrc`
+          echo "::set-output name=content::$content"
+      - name: Setup Node.js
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -12,10 +12,17 @@ jobs:
       - name: checkout repo
         uses: actions/checkout@v2
 
+      - name: Read NodeJS version
+        id: nodejs-version
+        run: |
+          $content = Get-Content .\.nvmrc -Raw
```
jpogran, line 18: Much as it pains me to say it, we don't need a Windows CI runner here. We can build on ubuntu and speed up things here, while also keeping the shell languages to a minimum

radeksimko: That is a good point - I don't remember why we picked Windows originally. That said I reckon we'd go back to Windows if/when we repurpose this action for publishing preview builds? https://github.com/hashicorp/vscode-terraform/issues/774

I _guess_ `vsce` works on other platforms as well, but I'm also _guessing_ that it works best on Windows? But these are all guesses honestly - I'm happy to do this on Linux - it would also probably save us some 💰  on GHA minutes.

.github/workflows/test.yml
```diff
@@ -15,10 +15,17 @@ jobs:
...
+        id: nodejs-version
+        run: |
+          content=`cat ./.nvmrc`
+          echo "::set-output name=content::$content"
+      - name: Setup Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 14.x
+          # See https://github.com/actions/setup-node/issues/32
+          node-version: ${{ steps.nodejs-version.outputs.content }}
```
jpogran, line 28: The one drawback I see from making this tied to nvmrc is that we lose the ability to make the node version a matrix and test on LTS and current versions of node. This would be useful to catch features dependent on specific package versions, but I don't think we're at that point yet. In practice, it's more important to us to test on different versions of terraform than node, so this is probably bike shedding.

radeksimko: Yeah - I'm open to introducing a matrix of Node.js versions here, but I reckon we'd still want to be testing against the version used by nvm.

radeksimko: Either way I'd prefer to leave that for another day/PR.

jpogran: Agreed

.github/workflows/build.yml
jpogran: I can confirm from prior experience `vsce` works on nix as well as it does on windows. We wouldn't have to use Windows to publish preview or normal builds, the OS we build on doesn't affect the OS we deploy to.

jpogran (approved):
LGTM!

PR #777, opened 2021-09-15
ci: constraint Terraform for testing to `~1.0`

radeksimko:
We generally expect the extension to work with latest stable versions of Terraform, but this gives us more predictability and better reproducible test failures (if any occur).

https://www.npmjs.com/package/semver#ranges


jpogran (approved):
This sets us up nicely for matrix testing later on 👍🏼

PR #778, opened 2021-09-15
fix: avoid tracking client which is not ready yet

radeksimko:
This is to address a long-term issue which was also spotted in https://github.com/hashicorp/vscode-terraform/pull/771#discussion_r709226812


jpogran (approved):
I think this is a good first step at resolving some of the client connections issues we've been having.

PR #784, opened 2021-09-21
ci: change windows-style slash to unix-style

radeksimko:
This is to address the following failure:
https://github.com/hashicorp/vscode-terraform/actions/runs/1258148142


jpogran (approved):
The one true path

PR #790, opened 2021-09-23
deps: bump vscode-extension-telemetry to 0.4.0

radeksimko:
This upgrade brings in https://github.com/microsoft/vscode-extension-telemetry/pull/70 amongst other changes, which should provide more privacy-respectful data collection.


radeksimko:
This will require vscode >= 1.55.0 because of https://github.com/microsoft/vscode-extension-telemetry/commit/572a1130b1485842d0b94e0df3892c12c4223139 which effectively introduced dependency on `isNewAppInstall` which was introduced to VS Code in https://github.com/microsoft/vscode/commit/95d3c1e63b7f73491db1f2bde7228763d0cd99ce

PR #792, opened 2021-09-23
Prevent Module View Provider from failing to call LS

jpogran:
This detects whether the currently installed Language Server supports the module calls command before calling the Language Server command. This prevents the provider from throwing an error with a LS that doesn't have the command.


radeksimko (approved):
LGTM 🚢

PR #794, opened 2021-09-23
Update launch configuration and tasks

### LINKED ISSUE #622: Terraform binary is required to debug the extension
While trying to develop the extension I followed the instructions in `DEVELOPMENT.MD`.  Pressing F5 to start a debug session just hung. Turns out there were two issues.

1. In `.vscode/tasks.json` the default `"reveal"` is all set to `never`.  So I could never see why it failed

2. The debug "Launch Client" configuration calls the "npm watch" task.  However, there are actually two tasks defined for "npm watch" in `tasks.json`.  The one you'd expect, and `initAndWatch`.  I'm guessing the debug configuration took the last defined one (`initAndWatch`) and tried to run that, which tried to call `terraform init` and failed because I did not have terraform installed.

Bottom line, the development instructions should be updated to include that terraform is a required piece of software for development OR the `tasks.json` file should be cleaned up so the correct task gets called instead.

This is not an issue with the extension per-se, it's an onboarding developers issue.

jpogran:
This modifies `launch.json` and `tasks.json` to only run `terraform init` when running the tests through the ExtensionHost. This allows debugging the extension without running terraform each time, taking the total time to start the ExtensionHost from several seconds to less than one.

This also reduces tasks.json task references to npm scripts to one entry for `npm watch` and consolidates the `terraform init` task to one entry.

It also removes the need to have terraform installed just to debug the extension, as the primary reason is to debug is interactions with the language server which doesn't use the binary for most operations.

Fixes #622

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,68 +1,32 @@
 {
-	"version": "2.0.0",
-	"tasks": [
-		{
-			"type": "npm",
-			"script": "compile",
-			"group": "build",
-			"presentation": {
-				"panel": "dedicated",
-				"reveal": "never"
-			},
-			"problemMatcher": [
-				"$tsc"
-			]
-		},
-		{
-			"type": "npm",
-			"script": "watch",
-			"isBackground": true,
-			"group": {
-				"kind": "build",
-				"isDefault": true
-			},
-			"presentation": {
-				"panel": "dedicated",
-				"reveal": "never"
-			},
-			"problemMatcher": [
-				"$tsc-watch"
-			]
-		},
-		{
-			"label": "init",
-			"type": "process",
-			"command": "terraform",
-			"args": [
-				"init"
-			],
-			"options": {
-				"cwd": "${workspaceFolder}/testFixture"
-			},
-			"presentation": {
-				"panel": "dedicated",
-				"reveal": "never"
-			},
-		},
-		{
-			"label": "initAndWatch",
-			"type": "npm",
-			"script": "watch",
-			"isBackground": true,
-			"group": {
-				"kind": "build",
-				"isDefault": true
-			},
-			"presentation": {
-				"panel": "dedicated",
-				"reveal": "never"
-			},
-			"problemMatcher": [
-				"$tsc-watch"
-			],
-			"dependsOn": [
-				"init"
-			]
-		},
-	]
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "type": "npm",
+      "script": "watch",
+      "problemMatcher": "$tsc-watch",
+      "isBackground": true,
+      "presentation": {
+        "reveal": "never"
+      },
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "terraformInitAndWatch",
+      "type": "process",
+      "command": "terraform",
+      "args": ["init"],
+      "options": {
+        "cwd": "${workspaceFolder}/testFixture"
+      },
+      "presentation": {
+        "panel": "dedicated",
+        "reveal": "never"
+      },
+      "dependsOn":"npm: watch"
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -1,68 +1,32 @@
...
+      "presentation": {
+        "reveal": "never"
+      },
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      }
+    },
+    {
+      "label": "terraformInitAndWatch",
```
radeksimko, line 18: Thank you for that more specific name here! 👍🏻

package.json
```diff
@@ -270,12 +270,12 @@
...
   "scripts": {
     "vscode:prepublish": "npm run compile",
-    "package": "vsce package",
-    "compile": "tsc -b",
-    "watch": "tsc -b -w",
-    "lint": "eslint src/**/*.ts",
-    "test-compile": "tsc -p ./",
-    "test": "npm run test-compile && node ./out/test/runTest.js"
+    "compile": "tsc -p ./",
+    "watch": "tsc -watch -p ./",
```
radeksimko, line 274: I like that these long flags avoid the need to read manual in most cases, is it worth also replacing `-p` with `-project`?

btw. `tsc --help` doesn't explicitly say that these long single-hyphen options are supported, but I suppose they are. 🤷🏻‍♂️

package.json
```diff
@@ -270,12 +270,12 @@
   },
   "scripts": {
     "vscode:prepublish": "npm run compile",
-    "package": "vsce package",
-    "compile": "tsc -b",
-    "watch": "tsc -b -w",
-    "lint": "eslint src/**/*.ts",
-    "test-compile": "tsc -p ./",
-    "test": "npm run test-compile && node ./out/test/runTest.js"
+    "compile": "tsc -p ./",
```
radeksimko, line 273: If I understand it correctly `-p` (unlike `-b`) would _not_ recompile the dependencies, am I right?

```
 -p FILE OR DIRECTORY, --project FILE OR DIRECTORY  Compile the project given the path to its configuration file, or to a folder with a 'tsconfig.json'.
 -b, --build                                        Build one or more projects and their dependencies, if out of date
```

I'm fine with it either way as I often usually blow away `node_modules` relatively regularly anyway when changing dependencies, so I'm not quite convinced that even `-b` does good enough job in terms of recompilation anyway - or maybe I just have Node.js set up in some wrong way.

package.json
```diff
@@ -270,12 +270,12 @@
...
+    "watch": "tsc -watch -p ./",
+    "lint": "eslint src --ext ts",
+    "pretest": "npm run compile && npm run lint",
```
radeksimko, line 276: What is the use case for the `pretest` task?

radeksimko (approved):

package.json
jpogran: Conventions? Parameter specifications rules? Who needs them? 😭 

I'm following conventions I've seen in other projects here, but if you want more concrete reasoning I'll look into it.

package.json
jpogran: I changed these because that's what the VS Code Yeoman templates specify as the default: https://github.com/microsoft/vscode-generator-code/blob/c3dc4980adb4aeb6cede8393f356ddd715fd397b/generators/app/templates/ext-command-ts/package.json#L26-L32.

FWIW calling `tsc` directly will go away when we move to `esbuild` or to `webpack`. Then we get a whole other set of challenges with debugging minified code 😭

package.json
jpogran: Another thing that comes from the default `yeoman` templates. Anything prepended with `pre` gets run before the similarly named script. So `pretest` gets run before `test. This reminds me I forgot to remove `npm compile` from `test.

radeksimko (approved):

PR #799, opened 2021-09-29
Prefill Required Attributes

jpogran:


radeksimko (dismissed):
Can you confirm that this would not require us to delay releasing this option as part of next extension version? LS will actually raise warnings when unknown config options are passed.

Unless it already works that way I think we could avoid sending the option there entirely if it's (default) `false`, so that it only affects people who explicitly opt-in and do so too early, before upgrading LS.

jpogran:
Adding to the existing `terraform-ls.experimentalFeatures` won't send anything automatically because the default value for `terraform-ls.experimentalFeatures` is undefined. Only when the user adds the property to their config would it send something.

With that in mind, I think this is safe to release at the same time as the LS

PR #811, opened 2021-10-14
Install LS into dedicated (persistent) global storage

radeksimko:
Closes https://github.com/hashicorp/vscode-terraform/issues/739

--- 

This changes where the language server gets installed upon first extension activation, or upgrade.

Previously we would install it under:

(macOS) `/Users/radeksimko/.vscode/extensions/hashicorp.terraform-2.15.0/lsp/terraform-ls`

now we install it into:

(macOS) `/Users/radeksimko/Library/Application Support/Code/User/globalStorage/hashicorp.terraform/bin/terraform-ls`

The intended benefit of this change is that LS doesn't need to be re-installed whenever an extension version changes and also we can reduce the likelihood of running into restrictive FS permissions which may rightly be applied to the `~/.vscode` in some cases.

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -125,9 +125,16 @@ export class LanguageServerInstaller {
       throw new Error(`Install error: no matching terraform-ls binary for ${os}/${arch}`);
     }
     try {
-      this.removeOldBinary();
+      fs.unlinkSync(this.lsPath.binPath());
     } catch {
       // ignore missing binary (new install)
+    }
+
+    try {
+      fs.unlinkSync(this.lsPath.legacyBinPath());
+    } catch {
+      // clean up may fail for new installation
+      // or in new versions where this path is no longer in use
     }
 
     return vscode.window.withProgress(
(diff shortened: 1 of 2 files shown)


radeksimko:
@husunal Just to understand the policy and context, what path(s) your company _does_ allow to run executables from? Also is there any justification that the company has for preventing execution from these paths?

FWIW You can maintain a custom installation by following these steps depending on your platform https://github.com/hashicorp/terraform-ls#installation and setting path in settings under `terraform.languageServer > pathToBinary`

https://github.com/hashicorp/vscode-terraform/blob/943a57b0a2859ae93242db66f8876e126b944998/package.json#L100-L105

The downside is that it will not be automatically updated for you - you would have to rely on the relevant package manager in your Linux distribution or Homebrew in macOS.

husunal:
We are using `VSCODE_EXTENSIONS` env variable to install all extensions to the specific allowed path. I'd like to manage it via VS Code extensions, not manually or via package manager so I won't use `pathToBinary` setting.

It would be good to support `VSCODE_EXTENSIONS` env var.

PR #814, opened 2021-10-15
Publish preview builds

jpogran:
This adds a new Github Action that publishes the `hashicorp.terraform-preview` extension to the VS Code Marketplace.

https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform-preview

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,54 @@
+name: Publish Preview release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish:
+    name: Publish Preview to VS Code Marketplace
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Set preview in package.json
+        id: set-preview
+        run: |
+          ./build/preview
+
+      - name: Read extension version
+        id: ext-version
+        run: |
+          content=`cat ./package.json | jq -r .version`
+          echo "::set-output name=content::$content"
+
+      - name: Read Node.js version
+        id: nodejs-version
+        run: |
+          content=`cat ./.nvmrc`
+          echo "::set-output name=content::$content"
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          # See https://github.com/actions/setup-node/issues/32
+          node-version: ${{ steps.nodejs-version.outputs.content }}
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: vsce package
+        run: npx -- vsce package --out=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
+
+      - name: Check latest published version
+        run: |
+          export EXTENSION_ID=`cat package.json | jq -r '.publisher + "." + .name'`
+          npx -- vsce show --json $EXTENSION_ID | jq '.versions[0]'
+
+      - name: vsce publish
+        run: npx -- vsce publish --no-git-tag-version --packagePath=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
+        env:
+          VSCE_PAT: ${{ secrets.VSCE_PAT }}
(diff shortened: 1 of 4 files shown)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,67 @@
...
+          content=`cat ./package.json | jq -r .version`
+          echo "::set-output name=content::$content"
+
+      - name: Ensure version matches with tag
+        if: ${{ github.ref != format('refs/tags/v{0}', steps.ext-version.outputs.content) }}
+        run: |
+          echo "Version mismatch!"
+          echo "Received ref: ${{ github.ref }}"
+          echo "Expected ref: refs/tags/v${{ steps.ext-version.outputs.content }}"
+          exit 1
```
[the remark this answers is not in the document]
jpogran: Good catch!

build/nightly.md
```diff
@@ -0,0 +1,13 @@
+# Terraform Preivew Visual Studio Code Extension
+
+> ### **ATTENTION: This is the *preview* version of the [VS Code Terraform extension](https://github.com/golang/vscode-terraform), used for early feedback and testing.**
+> It is contains previews of new features and bug fixes that are still under review or test. Therefore, this extension can be broken or unstable at times. If you are looking for the stable version,
+please install [the default VS Code Terraform extension](https://marketplace.visualstudio.com/items?itemName=hashicorp.terraform) instead.
+>
+> **NOTE: The stable Terraform extension (`hashicorp.terraform`) cannot be used at the same time as the Terraform Preview extension (`hashicorp.terraform-preview`)**. If you have installed both extensions, you **must disable or uninstall** one of them. For further guidance, read the [documentation on how to disable an extension](https://code.visualstudio.com/docs/editor/extension-gallery#_disable-an-extension).
```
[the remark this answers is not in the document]
jpogran: Agreed.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,53 @@
...
+          echo "::set-output name=content::$content"
+
+      - name: Read Node.js version
+        id: nodejs-version
+        run: |
+          content=`cat ./.nvmrc`
+          echo "::set-output name=content::$content"
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
```
dbanck, line 34: Would it be possible to use `v2` here?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,53 @@
...
+          # See https://github.com/actions/setup-node/issues/32
+          node-version: ${{ steps.nodejs-version.outputs.content }}
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: vsce package
+        run: npx -- vsce package --out=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
+
+      - name: Check latest published version
```
dbanck, line 45: Do we need this step?

build/preview
```diff
@@ -0,0 +1,27 @@
...
+ROOT_RELATIVE_DIR=$(dirname "${SCRIPT_RELATIVE_DIR}")
+
+cd $ROOT_RELATIVE_DIR
+
+# get current version info
+VERSION=`git log -1 --format=%cd --date="format:%Y.%-m.%-d%H"`
+
+# Get existing name & description
+DISPLAY_NAME="$(jq -r .displayName package.json) Preview"
+DESCRIPTION="$(jq -r .description package.json) Preview"
```
dbanck, line 15: The description currently has brackets around `Preview`. Do we want to keep these?

CleanShot 2021-11-09 at 14 21 40@2x

dbanck (approved):
LGTM! 🌟 

I only left a few minor comments/suggestions.

.github/workflows/preview.yml
jpogran: Good point. This was a copypasta from our https://github.com/hashicorp/vscode-terraform/blob/main/.github/workflows/publish.yml#L33 so we should probably update that there too

.github/workflows/preview.yml
jpogran: Technically no, but it is useful to verify in the logs of the action that the version was properly set. This may be premature, but I'm thinking of a case where the magic `build/preview` script does not update the version correctly and we need to see what was done.

build/preview
jpogran: Good catch, fixed!

radeksimko (approved):
ship-let-the-ship-begin

PR #823, opened 2021-10-20
Process telemetry from language server

radeksimko:
Depends on https://github.com/hashicorp/terraform-ls/pull/681 but the release does _not_ need to be coordinated. Older versions of LS will just ignore the experimental capability.

Regarding documentation, I believe the existing paragraph already reflects the existing situation which won't really change with this PR: https://github.com/hashicorp/vscode-terraform#telemetry

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -10,7 +10,7 @@ type TelemetryEvent = {
 };
 
 export class TelemetryFeature implements StaticFeature {
-  constructor(private client: BaseLanguageClient, private reporter: TelemetryReporter) {
+  constructor(private client: BaseLanguageClient, reporter: TelemetryReporter) {
     this.client.onTelemetry((event: TelemetryEvent) => {
       if (event.v != TELEMETRY_VERSION) {
         console.log(`unsupported telemetry event: ${event}`);
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -6,7 +6,7 @@ const TELEMETRY_VERSION = 1;
 type TelemetryEvent = {
   v: number;
   name: string;
-  properties: { string };
+  properties: { [key: string]: unknown };
 };
 
 export class TelemetryFeature implements StaticFeature {
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1,0 +1,38 @@
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { BaseLanguageClient, ClientCapabilities, StaticFeature } from 'vscode-languageclient';
+
+const TELEMETRY_VERSION = 1;
+
+type TelemetryEvent = {
+  v: number;
+  name: string;
+  properties: { string };
+};
+
+export class TelemetryFeature implements StaticFeature {
+  constructor(private client: BaseLanguageClient, private reporter: TelemetryReporter) {
+    this.client.onTelemetry((event: TelemetryEvent) => {
+      if (event.v != TELEMETRY_VERSION) {
+        console.log(`unsupported telemetry event: ${event}`);
+        return;
+      }
+
+      reporter.sendRawTelemetryEvent(event.name, event.properties);
+    });
+  }
+
+  public fillClientCapabilities(capabilities: ClientCapabilities): void {
+    if (!capabilities['experimental']) {
+      capabilities['experimental'] = {};
+    }
+    capabilities['experimental']['telemetryVersion'] = TELEMETRY_VERSION;
+  }
+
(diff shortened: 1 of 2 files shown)


src/telemetry.ts
```diff
@@ -0,0 +1,38 @@
+import TelemetryReporter from 'vscode-extension-telemetry';
+import { BaseLanguageClient, ClientCapabilities, StaticFeature } from 'vscode-languageclient';
+
+const TELEMETRY_VERSION = 1;
+
+type TelemetryEvent = {
+  v: number;
+  name: string;
+  properties: { string };
```
dbanck, line 9: To avoid an implicit any here, we should specify a type for the object values, too.
```suggestion
  properties: { [key: string]: unknown };
```

src/telemetry.ts
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export class TelemetryFeature implements StaticFeature {
+  constructor(private client: BaseLanguageClient, private reporter: TelemetryReporter) {
+    this.client.onTelemetry((event: TelemetryEvent) => {
```
dbanck, line 14: `onTelemetry` returns a Disposable. Do we need to keep track of it and dispose it in `this.dispose`? Maybe @jpogran knows what's the convention for this.

src/telemetry.ts
```diff
@@ -0,0 +1,38 @@
...
+const TELEMETRY_VERSION = 1;
+
+type TelemetryEvent = {
+  v: number;
+  name: string;
+  properties: { string };
+};
+
+export class TelemetryFeature implements StaticFeature {
+  constructor(private client: BaseLanguageClient, private reporter: TelemetryReporter) {
```
dbanck, line 13: `reporter` isn't used outside the constructor, so you might not need to create a class property for it.

If we want to keep it for later usage, we could use `this.reporter` in line 20.

dbanck (commented):
I've added a few suggestions for now and want to dig into disposables for a bit

src/telemetry.ts
jpogran: I don't think so, as they are just events. If you were to keep track to call in dispose, you'd be keeping track of possibly hundreds or thousands of events to dispose at extension deactivate. That seems like a memory problem there.

What I'm more concerned about is that the `StaticFeature` class says `initialize` is called after the client receives the initialize request from the server but before the client sends the initialized notification to the server. We're setting up the handler in the constructor which isn't part of the above guarantee. According to https://github.com/Microsoft/vscode-languageserver-node#303-client-server-and-json-rpc, you must use the `send|on` methods after `onReady` returns, otherwise exceptions can happen. I think we should move this to the `initialize` method to ensure it's called at the correct time.

With that said, I'm not sure this needs to be a `StaticFeature`. This all occurs only if the LS sends a telemetry event. Registering a handler for something that never happens isn't an error condition. I think we could make this a three line addition to `ClientHandler` instead of an entire new class and initialization concern. Thoguhts?

radeksimko: That is a good point @jpogran - I will update the PR!

radeksimko: @jpogran actually I don't see how else I could fill in the client capabilities without using either `StaticFeature` or `DynamicFeature` https://github.com/microsoft/vscode-languageserver-node/blob/f16ca298dac93a99b2338129a2149a0ada9021dc/client/src/common/client.ts#L3716-L3760

radeksimko: > I think we should move this to the initialize method to ensure it's called at the correct time.
> https://github.com/Microsoft/vscode-languageserver-node#303-client-server-and-json-rpc

Strictly speaking we do not have any telemetry events coming from the server before `initialize`, so this probably won't matter today, but what you linked above seems to contradict LSP spec, which does allow telemetry notifications to be sent before:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize

> Until the server has responded to the `initialize` request with an `InitializeResult`, the client must not send any additional requests or notifications to the server. In addition the server is not allowed to send any requests or notifications to the client until it has responded with an `InitializeResult`, with the exception that during the `initialize` request the server is allowed to send the notifications `window/showMessage`, `window/logMessage` and **`telemetry/event`** as well as the `window/showMessageRequest` request to the client.

radeksimko: Here I'm assuming that `onReady()` would only return when initialization is finished, i.e. before `initialized` is sent from the client - please do correct me if I'm wrong.

dbanck (approved):

PR #829, opened 2021-10-28
Silent Langauge Server Installation

### LINKED ISSUE #815: Make LS installation silent by default
Closes https://github.com/hashicorp/vscode-terraform/issues/783

--- 

## Problem Statement

Whenever the user first installs the extension, or we release a new version of the language server, they're presented with a prompt such as this one:

Screenshot 2021-10-18 at 13 01 33

Then after confirmation the user is further presented with a confirmation popup linking to the Changelog:

Screenshot 2021-10-18 at 13 03 02

This behaviour was initially intended to make the process more transparent, based on the assumption that users want to have full control of what is being installed and when. However this seems misaligned with real VS Code user expectations and (relatedly) with how other popular VS Code extensions behave. This is also misaligned with how extensions themselves are updated in VS Code.

The default workflow involves VS Code updating extensions automatically on the background without any user interaction. In some cases (where the update requires reload) the update is staged until the user decides to click the button in Extensions pane or restarts VS Code.

Screenshot 2021-10-18 at 13 14 37

## Expected User Experience

Users will not receive none of the two popups _by default_. New versions are silently prepared for installation and installed upon next extension activation (i.e. typically with next VS Code launch).

## Proposal

Any new version of the language server will be automatically (silently) installed into a staging area on the filesystem, e.g. under a separate directory `ls-bin-staging` within the global storage path. To avoid interrupting the user during editing the new binary will be replaced at the next extension activation, i.e. the extension will check the expected staging path upon activation.

Global VS Code settings option `extensions.autoUpdate` is also reflected, i.e. updating LS is entirely turned off if the user has this set to `off`.

_(optional, depending on currently unavailable telemetry data)_

Introduce a new extension config option `lsChangelog: (true|false)` (defaulting to `false`) which would cause a popup with Changelog link to be shown once a new version is installed.

jpogran:
This uses the existing VS Code update settings to determine whether to check for updated LS and/or whether to update the LS server without prompting the user.

A user that sets `extensions.autoCheckUpdates` to true will automatically check for updates to the LS. If set to false, the extension will not check for updates to the LS.

A user that sets `extensions.autoUpdate` to true will receive new versions of the LS without a prompt.

The installation progress message is moved to the `Window` position to eliminate the final distraction but maintain notification of what is going on.

The end result is that if a user has the default settings for `extensions.autoCheckUpdates` and for `extensions.autoUpdate`, the LS will be kept up to date without prompting or distracting the user.

Closes #815

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -174,32 +174,38 @@ async function updateTerraformStatusBar(documentUri: vscode.Uri) {
 }
 
 async function updateLanguageServer(extVersion: string, clientHandler: ClientHandler, lsPath: ServerPath) {
-  console.log('Checking for language server updates...');
-  const hour = 1000 * 60 * 60;
-  languageServerUpdater.timeout(function () {
-    updateLanguageServer(extVersion, clientHandler, lsPath);
-  }, 24 * hour);
+  if (config('extensions').get<boolean>('autoCheckUpdates', true) === true) {
+    console.log('Scheduling automatic check for language server updates...');
+    const hour = 1000 * 60 * 60;
+    languageServerUpdater.timeout(function () {
+      updateLanguageServer(extVersion, clientHandler, lsPath);
+    }, 24 * hour);
+  }
 
   try {
     // skip install if a language server binary path is set
-    if (!lsPath.hasCustomBinPath()) {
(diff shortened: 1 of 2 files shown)


jpogran:
Closing in favor of https://github.com/hashicorp/vscode-terraform/issues/820

PR #834, opened 2021-11-04
Single LanguageClient

### LINKED ISSUE #816: Drop support of (old) LS without multiple folders
## Problem Statement

Users have reported bugs in https://github.com/hashicorp/vscode-terraform/issues/805 (and probably some other older issues too) which are most likely caused by the fact that we maintain a map of client/server instances where reading or writing is not thread-safe.

https://github.com/hashicorp/vscode-terraform/blob/943a57b0a2859ae93242db66f8876e126b944998/src/clientHandler.ts#L26

This can sometimes lead to client instance being read before being written. The problem is in most cases masked by the fact that the LS is initialized quickly, such that reading happens after, but this is rather happy coincidence than a guarantee.

## Expected User Experience

Most users will see no UX changes.

A small minority of users which have not upgraded to [LS 0.19.0](https://github.com/hashicorp/terraform-ls/blob/main/CHANGELOG.md#0190-8-july-2021) yet would be already affected by a different incompatibility we introduced earlier when upgrading `vscode-languageclient` to `7.0.0` in https://github.com/hashicorp/vscode-terraform/pull/780 and presented with the following error anyway

Screenshot 2021-10-18 at 14 26 44

## Proposal

Refactor `ClientHandler` such that a single client is only ever used and no situation occurs where we would attempt to access uninitialized client instance.

An error will be shown to the user whenever LS capability `workspace.workspaceFolders.supported` is set to `false` just to more explicitly communicate requirements, just in case we happen drop support in the future by accident or someone attempts to use the extension with a different LS which doesn't have that capability.

jpogran:
- Update extension tests
- Refactor to Single LanguageClient
- Extract supported commands to class property
- Simplify creating TerraformLanguageClient
- Extract resolvePathToBinary to ServerPath
- Fix ModuleProvider starting before client ready
- Enable LS in main activation flow
- Brand OutputChannel
- figure out if onDidChangeWorkspaceFolders still needed
- simplify creating language clients
- strongly type extension activation

Close #816 

Relies on test fixes in https://github.com/hashicorp/vscode-terraform/pull/828

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -1,6 +1,5 @@
 import * as fs from 'fs';
 import * as path from 'path';
-import ShortUniqueId from 'short-unique-id';
 import * as vscode from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
 import {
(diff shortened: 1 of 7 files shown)


src/clientHandler.ts
```diff
@@ -12,161 +9,99 @@ import {
...
-    }
-    if (rootModulePaths.length > 0) {
-      initializationOptions = Object.assign(initializationOptions, { rootModulePaths });
-    }
-    if (excludeModulePaths.length > 0) {
-      initializationOptions = Object.assign(initializationOptions, { excludeModulePaths });
-    }
-
+    const experimentalFeatures = config('terraform-ls').get('experimentalFeatures');
+    const initializationOptions = Object.assign(
```
dbanck, line 94: I was wondering if something like this is more readable:

```ts
const initializationOptions = {
      ...experimentalFeatures,
      ...(terraformExecPath.length > 0 && { terraformExecPath }),
      ...(terraformExecTimeout.length > 0 && { terraformExecTimeout }),
      ...
    };
```

jpogran:
Superseded by https://github.com/hashicorp/vscode-terraform/pull/845

PR #837, opened 2021-11-05
Implement opt out config option for ref count code lens

### LINKED ISSUE #713: Provide opt out config option for ref count code lens
### Use Cases

The most recent versions of the extension and LS introduced code lens for displaying reference counts above top level blocks and attributes.

While many users may find this useful, some may find it distracting or just generally may not want it (https://github.com/hashicorp/terraform-ls/pull/584#issuecomment-891190519) and it would be useful if these users have a way of opting out of this code lens.

### Proposal

Introduce a new config option `enableReferenceCountCodeLens` (defaulting to `true`) which will prevent the client from registering the relevant feature

https://github.com/hashicorp/vscode-terraform/blob/b3d812edd5ec334f5c42e7df109cef50dd32b5ac/src/clientHandler.ts#L181

and as a result prevent the relevant client capability to be sent to the server, which would in turn prevent it from sending the code lens back.

dbanck:
This PR adds a new setting called `enableReferenceCountCodeLens` which toggles displaying reference counts above top level blocks and attributes.

Closes #713.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,12 @@
             "trace.server": "off"
           }
         },
+        "terraform.enableReferenceCountCodeLens": {
+          "scope": "resource",
+          "type": "boolean",
+          "default": true,
+          "description": "Display reference counts above top level blocks and attributes."
+        },
         "terraform-ls.rootModules": {
           "scope": "resource",
           "type": "array",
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
LGTM :shipit: 

I reckon this may conflict with changes in https://github.com/hashicorp/vscode-terraform/pull/834 but hopefully shouldn't be too hard to rebase that PR 🤞🏻 

FYI @jpogran

PR #841, opened 2021-11-08
Fix ModuleProvider starting before client ready

jpogran:
Delays enabling the ModuleProvider until the client has started to avoid the case where the user has the view pane open on editor start but the language client or server has not fully initialized yet.


dbanck (approved):
LGTM

PR #844, opened 2021-11-10
Enable CodeAction tests

jpogran:
Enable the CodeAction tests that were commented out until the changes in https://github.com/hashicorp/terraform-ls/pull/680 was released.


dbanck (approved):
💯

Issue #820, opened 2021-10-19
Bundle language server with extension release


radeksimko:
## Problem Statement

Based on user's feedback reported here and also from available exception reports in AppInsights it is clear that there are users which experience various issues all caused by the fact that we attempt to:

1. download archive from a remote location over the network
2. place a new binary to the filesystem

For various reasons network can be either entirely unavailable or connection to where to store LS releases can be blocked by firewall. Filesystem can also impose some permission restrictions, making it hard to find a location that is writable and where binaries can also be executed from.

Secondly many VS Code users generally expect the extension to "just work" out of the box.

## Expected User Experience

User is able to use the extension (**with LS**) straight away after installing it, even if they have restrictive filesystem or firewall rules or no internet connection.

## Proposal

Publish vscode-terraform as a [platform specific extension](https://code.visualstudio.com/updates/v1_61). Each VSIX we publish will contain both the extension code and the platform specific terraform-ls binary. 

#### VS Code

Users will be required to have VS Code v1.61 or greater installed in order to have a platform specific extension offered through the Marketplace or through the VS Code Extension pane. Users who install via the VS Code Marketplace or through the `code` command line will receive a VSIX with the terraform-ls built for their platform.

If the user has VS Code v1.60 or lower, they will stay at the version of vscode-terraform previous to the release that introduces platform specific extension. They will not be offered any newer versions of vscode-terraform. Since the downloader logic still exists in the previous vscode-terraform version, they will continue to receive new terraform-ls versions but not any new vscode-terraform versions.

#### Platform

The VS Code Marketplace supports and distributes to a [select number of OS platforms and architectures](https://code.visualstudio.com/#alt-downloads). VS Code supports a [constrained list of OS platforms and architectures](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) to deploy platform specific extensions to. The terraform-ls binary supports [a large list of OS Platforms and architectures](https://github.com/hashicorp/terraform-ls/releases).

Taking into account all of the above lists results in the following chart, which shows that for all the platforms and architectures that VS Code supports deploying platform specific extensions to, we support for terraform-ls. 

platform | terraform-ls | extension platform | vs code editor
-- | -- | -- | --
macOS | darwin_amd64 | darwin_x64 | ✅
macOS | darwin_arm64 | darwin_arm64 | ✅
Linux | linux_amd64 | linux_x64 | ✅
Linux | linux_arm | linux_armhf | ✅
[...]

PR #845, opened 2021-11-10
Refactor to Single Language Client

### LINKED ISSUE #816: Drop support of (old) LS without multiple folders

jpogran:
## Refactor Language Client Workflow

- Refactors ClientHandler to only use one `LanguageClient` per extension activation. Removes the Map of `LanguageClients` and ensures that any attempt to use a `LanguageClient` or method happens after the onReady event finishes. The terraform-ls sever implemented multi-folder/workspace support in hashicorp/terraform-ls#502. This means the VS Code extension no longer needs multiple clients to track how many folders are added to the current workspace and which event for which file needs to be sent at certain times.
- Moves the list of supported commands to a private property of the ClientHandler class and populates it when the client and LS are initialized. Previously you could attempt to ask the server what commands were supported before the client and server had finished the initialization process. This would result in an unhandled exception and leave the client in an unknown state.
- Refactors the createTerraformClient method to reduce complexity and verbosity by using language constructs. Move variables closer to usage for future method extraction. We no longer need to track the `location` or `workspaceFolder` to pull settings or display information per folder. This greatly simplifies the code accessing settings and creating the `LanguageClient`.
- Move creation of the OutputChannel to the main flow and only create one outputChannel for the entire extension. Use 'HashiCorp Terraform' as the name the OutputChannel to brand our extension. The `name` is displayed in a drop down in the `Output` pane in the editor, so it should be reconizable and readable.
- Optimizes main execution flow by bringing language server updater to main execution flow, reducing the cases where a restart needs to occur to update, and reducing nesting constucts.
- Extracts `resolvePathToBinary` method to the ServerPath class. The `resolvedPathToBinary` resolves the LS filesystem path and uses methods inside the ServerPath class. The ClientHandler class is only responsible for LanguageClients, and so does not need this private method.

## Testing

- single folder workspace
- multiple folder workspace
- starting editor with no documents open
- starting editor with one or more documents already open
- starting editor with module view expanded

actions for each test:

1. open document, trigger intellisense and complete a resource from a provider
1. hover over a declaration or field and see hover doc info
1. navigate using any of the go to commands
1. navigate using variable reference

--- 

Closes #816

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -26,7 +26,7 @@ export interface TerraformLanguageClient {
 export class ClientHandler {
   private shortUid: ShortUniqueId;
   private tfClient: TerraformLanguageClient;
-  private commands: string[];
+  private commands: string[] = [];
 
   constructor(
     private lsPath: ServerPath,
@@ -66,7 +66,7 @@ export class ClientHandler {
 
     const initializationOptions = this.getInitializationOptions(commandPrefix);
 
-    const serverOptions: ServerOptions = this.getServerOptions();
+    const serverOptions = this.getServerOptions();
     this.outputChannel.appendLine(
       `Launching language server: ${serverOptions.run.command} ${serverOptions.run.args.join(' ')}`,
     );
@@ -107,7 +107,7 @@ export class ClientHandler {
 
   private getServerOptions() {
     const cmd = this.lsPath.resolvedPathToBinary();
-    const serverArgs: string[] = config('terraform').get('languageServer.args');
+    const serverArgs = config('terraform').get<string[]>('languageServer.args');
     const executable: Executable = {
       command: cmd,
       args: serverArgs,
@@ -121,7 +121,7 @@ export class ClientHandler {
   }
 
   private getInitializationOptions(commandPrefix: string) {
-    const rootModulePaths: string[] = config('terraform-ls').get('rootModules');
+    const rootModulePaths = config('terraform-ls').get<string[]>('rootModules', []);
     const terraformExecPath: string = config('terraform-ls').get('terraformExecPath');
     const terraformExecTimeout: string = config('terraform-ls').get('terraformExecTimeout');
     const terraformLogFilePath: string = config('terraform-ls').get('terraformLogFilePath');
@@ -162,7 +162,7 @@ export class ClientHandler {
       return Promise.resolve();
     }
 
-    if (this.tfClient.client === undefined) {
+    if (this.tfClient?.client === undefined) {
       return Promise.resolve();
     }
 
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -63,10 +63,7 @@ export class ClientHandler {
 
     const initializationOptions = this.getInitializationOptions(commandPrefix);
 
-    const serverOptions: ServerOptions = this.getServerOptions();
-    this.outputChannel.appendLine(
-      `Launching language server: ${serverOptions.run.command} ${serverOptions.run.args.join(' ')}`,
-    );
+    const serverOptions = this.getServerOptions();
 
     const documentSelector: DocumentSelector = [
       { scheme: 'file', language: 'terraform' },
@@ -102,7 +99,7 @@ export class ClientHandler {
     return { commandPrefix, client };
   }
 
-  private getServerOptions() {
+  private getServerOptions(): ServerOptions {
     const cmd = this.lsPath.resolvedPathToBinary();
     const serverArgs = config('terraform').get<string[]>('languageServer.args');
     const executable: Executable = {
@@ -114,6 +111,7 @@ export class ClientHandler {
       run: executable,
       debug: executable,
     };
+    this.outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
     return serverOptions;
   }
 
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -24,8 +24,8 @@ export interface TerraformLanguageClient {
  * based on the server's capabilities
  */
 export class ClientHandler {
-  private shortUid: ShortUniqueId;
-  private tfClient: TerraformLanguageClient;
+  private shortUid: ShortUniqueId = undefined;
+  private tfClient: TerraformLanguageClient = undefined;
   private commands: string[] = [];
 
   constructor(
@@ -34,7 +34,6 @@ export class ClientHandler {
     private reporter: TelemetryReporter,
   ) {
     this.shortUid = new ShortUniqueId();
-    this.commands = [];
     if (lsPath.hasCustomBinPath()) {
       this.reporter.sendTelemetryEvent('usePathToBinary');
     }
@@ -90,7 +89,7 @@ export class ClientHandler {
     const id = `terraform`;
     const client = new LanguageClient(id, serverOptions, clientOptions);
 
-    const codeLensReferenceCount = config('terraform').get('codelens.referenceCount');
+    const codeLensReferenceCount = config('terraform').get<boolean>('codelens.referenceCount');
     if (codeLensReferenceCount) {
       client.registerFeature(new ShowReferencesFeature(client));
     }
@@ -105,7 +104,7 @@ export class ClientHandler {
     return { commandPrefix, client };
   }
 
-  private getServerOptions() {
+  private getServerOptions(): ServerOptions {
     const cmd = this.lsPath.resolvedPathToBinary();
     const serverArgs = config('terraform').get<string[]>('languageServer.args');
     const executable: Executable = {
@@ -122,11 +121,11 @@ export class ClientHandler {
 
   private getInitializationOptions(commandPrefix: string) {
     const rootModulePaths = config('terraform-ls').get<string[]>('rootModules', []);
-    const terraformExecPath: string = config('terraform-ls').get('terraformExecPath');
-    const terraformExecTimeout: string = config('terraform-ls').get('terraformExecTimeout');
-    const terraformLogFilePath: string = config('terraform-ls').get('terraformLogFilePath');
-    const excludeModulePaths: string[] = config('terraform-ls').get('excludeRootModules');
-    const ignoreDirectoryNames: string[] = config('terraform-ls').get('ignoreDirectoryNames');
+    const terraformExecPath = config('terraform-ls').get<string>('terraformExecPath');
+    const terraformExecTimeout = config('terraform-ls').get<string>('terraformExecTimeout');
+    const terraformLogFilePath = config('terraform-ls').get<string>('terraformLogFilePath');
+    const excludeModulePaths = config('terraform-ls').get<string[]>('excludeRootModules');
+    const ignoreDirectoryNames = config('terraform-ls').get<string[]>('ignoreDirectoryNames');
 
     if (rootModulePaths.length > 0 && excludeModulePaths.length > 0) {
       throw new Error(
@@ -158,22 +157,12 @@ export class ClientHandler {
   }
 
   public async stopClient(): Promise<void> {
-    if (this.tfClient === undefined) {
-      return Promise.resolve();
-    }
-
     if (this.tfClient?.client === undefined) {
       return Promise.resolve();
     }
 
-    return this.tfClient.client
-      .stop()
-      .then(() => {
-        console.log('Client stopped');
-      })
-      .then(() => {
-        console.log('Client deleted');
-      });
+    await this.tfClient.client.stop();
+    console.log('Client stopped');
   }
 
   public getClient(): TerraformLanguageClient {
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -1,5 +1,3 @@
-import * as fs from 'fs';
-import * as path from 'path';
 import ShortUniqueId from 'short-unique-id';
 import * as vscode from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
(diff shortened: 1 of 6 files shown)


src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
- * based on the server's capabilities (whether multi-folder
- * workspaces are supported).
+ * based on the server's capabilities
  */
 export class ClientHandler {
   private shortUid: ShortUniqueId;
-  private supportsMultiFolders = true;
-
-  constructor(private lsPath: ServerPath, private reporter: TelemetryReporter) {
+  private tfClient: TerraformLanguageClient;
```
dbanck, line 28: Should we initialize this with `null` here? It might help when we enable strict null checks at some point.

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
+  private commands: string[];
```
dbanck, line 29:
```suggestion
  private commands: string[] = [];
```
and remove it in the constructor?

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
-          await this.startClients(folders);
-        }
-      });
+    console.log('Starting client');
 
-      disposables.push(tfClient.client.start());
-      await readyClient;
-      clients.set(MULTI_FOLDER_CLIENT, tfClient);
+    this.tfClient = this.createTerraformClient();
+    const readyClient = this.tfClient.client.onReady().then(async () => {
```
dbanck, line 49: Would it be possible to get rid of the `then` here? Maybe something like this?

```ts
this.tfClient = this.createTerraformClient();
disposables.push(this.tfClient.client.start());
await this.tfClient.client.onReady();

this.reporter.sendTelemetryEvent('startClient');
[...]
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
-      outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
-    }
+    const initializationOptions = this.getInitializationOptions(commandPrefix);
 
-    if (rootModulePaths.length > 0 && excludeModulePaths.length > 0) {
-      throw new Error(
-        'Only one of rootModules and excludeRootModules can be set at the same time, please remove the conflicting config and reload',
-      );
-    }
+    const serverOptions: ServerOptions = this.getServerOptions();
```
dbanck, line 69: We could get rid of the type here and add it as return type of `this.getServerOptions()`
```suggestion
    const serverOptions = this.getServerOptions();
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
+    };
+    return serverOptions;
+  }
 
-    if (this.supportsMultiFolders) {
-      promises.push(this.stopClient(MULTI_FOLDER_CLIENT));
-      return Promise.all(promises);
-    }
+  private getInitializationOptions(commandPrefix: string) {
+    const rootModulePaths: string[] = config('terraform-ls').get('rootModules');
```
dbanck, line 124: Here too and maybe add a sensible default for the following `length` checks
```suggestion
    const rootModulePaths = config('terraform-ls').get<string[]>('rootModules', []);
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
+    if (rootModulePaths.length > 0 && excludeModulePaths.length > 0) {
+      throw new Error(
+        'Only one of rootModules and excludeRootModules can be set at the same time, please remove the conflicting config and reload',
+      );
     }
 
-    for (const folder of folders) {
-      promises.push(this.stopClient(folder));
+    const experimentalFeatures = config('terraform-ls').get('experimentalFeatures');
+    let initializationOptions = { commandPrefix, experimentalFeatures };
```
dbanck, line 138: This might be easier to read and results in a better type for `initializationOptions `.

```ts
const initializationOptions = {
  commandPrefix,
  experimentalFeatures,
  ...(terraformExecPath.length > 0 && { terraformExecPath }),
  ...(terraformExecTimeout.length > 0 && { terraformExecTimeout }),
  ...
};
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
-      outputChannel: outputChannel,
+      outputChannel: this.outputChannel,
       revealOutputChannelOn: RevealOutputChannelOn.Never,
     };
 
-    const client = new LanguageClient(id, name, serverOptions, clientOptions);
+    const id = `terraform`;
+    const client = new LanguageClient(id, serverOptions, clientOptions);
 
+    const enableReferenceCountCodeLens: boolean = config('terraform').get('enableReferenceCountCodeLens');
```
dbanck, line 93: After getting rid of all the let variables, we can now pass the type to get and get rid of the cast. 

```suggestion
    const enableReferenceCountCodeLens = config('terraform').get<v>('enableReferenceCountCodeLens');
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
     });
 
     return { commandPrefix, client };
   }
 
-  public async stopClients(folders?: string[]): Promise<void[]> {
-    const promises: Promise<void>[] = [];
+  private getServerOptions() {
+    const cmd = this.lsPath.resolvedPathToBinary();
+    const serverArgs: string[] = config('terraform').get('languageServer.args');
```
dbanck, line 110: Here too
```suggestion
    const serverArgs = config('terraform').get<string[]>('languageServer.args');
```

src/clientHandler.ts
```diff
@@ -12,298 +10,177 @@ import {
...
-  private async stopClient(folder: string): Promise<void> {
-    if (!clients.has(folder)) {
-      console.log(`Attempted to stop a client for folder: ${folder} but no client exists`);
-      return;
+  public async stopClient(): Promise<void> {
+    if (this.tfClient === undefined) {
+      return Promise.resolve();
+    }
+
+    if (this.tfClient.client === undefined) {
```
dbanck, line 165: I think we can ditch the first if when we use optional chaining here.
```suggestion
    if (this.tfClient?.client === undefined) {
```

and just return without the `Promise.resolve()`

dbanck (changes requested):
Fantastic work on the refactoring! It's much easier to understand and follow what is happening on activate now. 🌟 

I took some extra time for this review because I wanted to understand how everything works. So please don't feel discouraged by the number of comments. Everything is working well for me, and most are language-related things.

src/extension.ts
```diff
@@ -69,10 +71,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
         );
       }
       languageServerUpdater.clear();
-      return clientHandler.stopClients();
+      return clientHandler.stopClient();
```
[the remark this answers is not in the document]
jpogran: I'm not 100% sure since this is a VS Code command and not a normal function. TBH I don't like stopping/starting the client inside the commands either, and I'm hoping to remove them entirely.

src/extension.ts
```diff
@@ -170,43 +180,71 @@ async function updateTerraformStatusBar(documentUri: vscode.Uri) {
...
-          await installer.install();
-        } catch (err) {
-          console.log(err); // for test failure reporting
-          reporter.sendTelemetryException(err);
-          throw err;
-        } finally {
-          await installer.cleanupZips();
-        }
-      }
+    if (lsPath.hasCustomBinPath()) {
```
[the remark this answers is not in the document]
jpogran: This gnarly set of code goes away once we accomplish https://github.com/hashicorp/vscode-terraform/issues/820, so I was trying not to change how we update the LS too much here and focus only on what was needed to change to a single client.

If we don't accomplish #820, then yes I agree this method need refactoring.

src/extension.ts
```diff
@@ -170,43 +180,71 @@ async function updateTerraformStatusBar(documentUri: vscode.Uri) {
...
+      reporter.sendTelemetryException(err);
+      throw err;
+    } finally {
+      // clean upa fter ourselves and remove zip files
+      await installer.cleanupZips();
+    }
+
+    if (scheduled) {
+      // scheduled updates still need to start client
+      return await clientHandler.startClient();
```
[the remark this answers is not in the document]
jpogran: > Do we need to return here? There is no code left to execute in this function

Good point. This place was more complicated 1 or 10 attempts ago :lolsob:

> If `updateLanguageServer` is called via the command `terraform.enableLanguageServer`. We may want to start the client, too?

Good catch. Fixed.

src/extension.ts
```diff
@@ -170,43 +180,71 @@ async function updateTerraformStatusBar(documentUri: vscode.Uri) {
...
     vscode.window.showErrorMessage(error.message);
   }
 }
 
-function execWorkspaceCommand(client: LanguageClient, params: ExecuteCommandParams): Promise<any> {
-  reporter.sendTelemetryEvent('execWorkspaceCommand', { command: params.command });
+function execWorkspaceCommand(
+  client: LanguageClient,
+  params: ExecuteCommandParams,
+  reporter: TelemetryReporter,
```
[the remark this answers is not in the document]
jpogran: Artifact of a previous attempt where I had extracted the function to a separate file. Not needed anymore.

src/extension.ts
```diff
@@ -111,23 +113,22 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
...
     }),
-    vscode.workspace.onDidChangeWorkspaceFolders(async (event: vscode.WorkspaceFoldersChangeEvent) => {
-      if (event.removed.length > 0) {
-        await clientHandler.stopClients(prunedFolderNames(event.removed));
-      }
-      if (event.added.length > 0) {
-        await clientHandler.startClients(prunedFolderNames(event.added));
-      }
-    }),
     vscode.window.onDidChangeVisibleTextEditors(async () => {
```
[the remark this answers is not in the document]
jpogran: Seems to work, good catch!

radeksimko (approved):
Looks so much cleaner! ❤️ Thank you.

--- 

I have just a few questions, but none of them truly blocking:

If I'm understanding correctly, we now no longer support single-folder LS, which now no longer works with the latest extension anyway due to other incompatibilities. For transparency, "future-proofing" and just being a good citizen in LSP - how would you feel about raising an error when the server (for any reason) communicates that it doesn't support multiple folders? I'm not sure what exact implications does that have, but I'm assuming it may not work as expected in that case - either way it seems like we shouldn't just silently ignore such a (rare) case.

> Use 'HashiCorp Terraform' as the name the OutputChannel to brand our extension.

I like the idea that we'd have a single pane to point users to... However I'm not sure we're ready for this (yet). 🤔 

Maybe if the server used structured logs, e.g. via [`window/logMessage`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_logMessage) this would be viable, but as long as it's just stream of bytes there's a real risk that the two just get randomly mixed up. Even if/when we convert all the LS logging to the LSP method there's still some logs related to startup, before the RPC server is even ready and I'm not sure what the performance implications of potentially verbose logging over LSP are.

All that said the risks are currently pretty low as we already use a single pane and the extension itself doesn't log as much - so I'm 👌🏻 with this as-is, but I feel we need to talk about our logging strategy soon before the extension starts to utilize the log pane more.

In contrast the official Go extension registers 5 panes in total, which I'm not necessarily presenting as an example we should follow, but merely for context.

Screenshot 2021-11-12 at 09 54 16

jpogran:
> If I'm understanding correctly, we now no longer support single-folder LS, which now no longer works with the latest extension anyway due to other incompatibilities. For transparency, "future-proofing" and just being a good citizen in LSP - how would you feel about raising an error when the server (for any reason) communicates that it doesn't support multiple folders? I'm not sure what exact implications does that have, but I'm assuming it may not work as expected in that case - either way it seems like we shouldn't just silently ignore such a (rare) case.

We check if LS supports multi-folder workspaces and log to the console currently, I could add a popup (that could be suppressed) notifying the user that the LS doesn't support the current workspace configuration.
 
> Use 'HashiCorp Terraform' as the name the OutputChannel to brand our extension.

The 'branding' change was to align with how most other popular extensions name the channels, i.e. with human readable names. This is the opportunity to distinguish our logs from other extensions that could be installed at the same time. To me, I look at logs when something is going awry, and at that time it is more important to me to know which log is which extension so I can identify the next steps to take. I think we're in agreement that the branding makes sense and it's more the second comment that's the question, but just writing here to make sure.

> Maybe if the server used LSP-based logging via [`window/logMessage`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_logMessage) this would be viable, but as long as it's just stream of bytes there's a real risk that the two just get randomly mixed up. Even if/when we convert all the LS logging to the LSP method there's still some logs related to startup, before the RPC server is even ready and I'm not sure what the performance implications of potentially verbose logging over LSP are.

We already log all LS to the outputchannel for that client, so not sure that there is any net new functionality here to think about throughput. The change here is that there is now only one: previously we created an outputchannel per client, since we only have one client now we only have one outputchannel. There is room for improvement in how/when we log, and as you state we'll be adding more to it to help users identify what is going on, but I think that is another PR.

radeksimko:
>  I could add a popup (that could be suppressed) notifying the user that the LS doesn't support the current workspace configuration.

I think that would be nice.

> We already log all LS to the outputchannel for that client, so not sure that there is any net new functionality here to think about throughput.

Yep, that's why I say it's totally fine in the context of this PR 😃  👍🏻  What seems to be changing is the (dev) _perception_ as previously we'd call it `terraform-ls` and extension logs would end up there more by accident, now we more explicitly acknowledge that it's a shared output pane, so it's more tempting to send more logs there.

Again - not a big deal and certainly not a blocker for this PR, just something to think about more broadly.

dbanck (approved):
Thank you for addressing and fixing all my comments. 👍 

I found just three minor things which maybe got lost.  Feel free to merge once addressed.

src/clientHandler.ts
```diff
@@ -12,298 +10,155 @@ import {
...
-          console.log(`Client deleted for ${folder}`);
-        }
-      });
-  }
+  private getInitializationOptions(commandPrefix: string) {
+    const rootModulePaths = config('terraform-ls').get<string[]>('rootModules', []);
+    const terraformExecPath = config('terraform-ls').get<string>('terraformExecPath');
+    const terraformExecTimeout = config('terraform-ls').get<string>('terraformExecTimeout');
+    const terraformLogFilePath = config('terraform-ls').get<string>('terraformLogFilePath');
+    const excludeModulePaths = config('terraform-ls').get<string[]>('excludeRootModules');
```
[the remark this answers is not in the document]
jpogran: I'd rather be explicit rather than implicit, but that's just my preference

dbanck (approved):

PR #848, opened 2021-11-10
Fix Warnings

jpogran:
Fixes warnings in extension activation by strongly typing return value. Safely ignores the rest as they rely on untyped responses.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,7 +16,12 @@ let reporter: TelemetryReporter;
 let clientHandler: ClientHandler;
 const languageServerUpdater = new SingleInstanceTimeout();
 
-export async function activate(context: vscode.ExtensionContext): Promise<any> {
+export interface TerraformExtension {
+  clientHandler: ClientHandler;
+  moduleCallers(languageClient: TerraformLanguageClient, moduleUri: string): Promise<moduleCallersResponse>;
+}
+
+export async function activate(context: vscode.ExtensionContext): Promise<TerraformExtension> {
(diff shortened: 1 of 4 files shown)


radeksimko (approved):
LGTM, thanks!

dbanck (commented):
I think we're not getting much value at this point from forcing "no explicit any" via eslint. So I would rather suggest disabling both rules globally instead of using inline comments.

```json
  "rules": {
    "@typescript-eslint/no-explicit-any": 0,
    "@typescript-eslint/explicit-module-boundary-types": 0
  }
```

We may want to look at our tsconfig next, add more restrictive compilerOptions, and then re-evaluate the eslint rules.

radeksimko:
@dbanck That's a good point - I'm also a bigger fan of global rules rather than inline comments when trying to "satisfy" a linter.

jpogran:
In general I'd agree about customizing rules to suit the project, but my intention here is the places I'm ignoring these rules are going away as we refactor the extension. Typescript gives us the ability to check types, and I don't want to encourage not having types 😁 .

If it's ok with the both of you, I'm going to create an issue off this for discussing/picking rulesets for the extension with a link to this commit to ensure the ignores/methods are removed eventually and merge this as is.

radeksimko:
@jpogran That's fair and I don't feel I have enough TypeScript experience here to suggest anything different 😄  I was looking at it purely from the perspective of extra comments that server no purpose to humans, but as you said they'll go away then that's fine with me. 🤷🏻‍♂️

PR #850, opened 2021-11-12
Add panel for module providers

### LINKED ISSUE #701: Display provider requirements/installed in the UI
## Use Cases

While providers can practically only be defined in a single place in config, it may not always be quick and easy to find that place and understand the implications.

Users would benefit from seeing what providers are required and installed for their module at a quick glance.

## Expected User Experience

The expectation is that we'd use [Tree View API](https://code.visualstudio.com/api/extension-guides/tree-view), which would look approximately like this:

Screenshot 2021-07-23 at 07 45 19

we may also consider adding the view directly under the **Explorer** menu, similar to this example from VS Code docs:

views

User would be able to see:
 - provider _requirements_ for the module(s) being edited
 - provider _versions installed_ for the module(s) being edited
 - any alternative names which each provider is available
 - any aliases under which each provider is available
 - right click to update outdated provider
 - right click to go to the docs of the provider, where applicable (Terraform Registry)

We may also consider implementing this as a WebView https://code.visualstudio.com/api/extension-guides/webview

## Proposal

 - decide UI details - i.e. what exact metadata and how we display
 - expose the relevant module data from LS via custom command
 - implement the client-side while leveraging the data from LS obtained via custom command

dbanck:
This PR adds a new tree view panel in the explorer menu with the installed providers in the current module.

It depends on the language server PR https://github.com/hashicorp/terraform-ls/pull/712

New panel, listing providers with the installed version:

CleanShot 2021-11-17 at 16 37 01@2x

No providers detected:

CleanShot 2021-11-17 at 16 35 20@2x

Providers not installed (yet):

CleanShot 2021-11-17 at 16 34 58@2x

Tooltip showing the required version:

CleanShot 2021-11-17 at 16 37 54@2x

Closes #701

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,6 +279,13 @@
     "views": {
       "explorer": [
         {
+          "id": "terraform.providers",
+          "name": "Terraform Provides",
+          "icon": "assets/icons/terraform.svg",
+          "visibility": "collapsed",
+          "when": "terraform.showProviderView"
+        },
+        {
           "id": "terraform.modules",
           "name": "Terraform Module Calls",
           "icon": "assets/icons/terraform.svg",
@@ -288,6 +295,10 @@
       ]
     },
     "viewsWelcome": [
+      {
+        "view": "terraform.providers",
+        "contents": "The active editor cannot provide information about installed providers."
+      },
       {
         "view": "terraform.modules",
         "contents": "The active editor cannot provide information about installed modules. [Learn more about modules](https://www.terraform.io/docs/language/modules/develop/index.html) You may need to run 'terraform get' or update your language server version."
--- a/package.json
+++ b/package.json
@@ -249,6 +249,11 @@
         "command": "terraform.modules.openDocumentation",
         "title": "Open Documentation",
         "icon": "$(book)"
+      },
+      {
+        "command": "terraform.providers.openDocumentation",
+        "title": "Open Documentation",
+        "icon": "$(book)"
       }
     ],
     "menus": {
@@ -259,6 +264,10 @@
         },
         {
           "command": "terraform.modules.openDocumentation",
+          "when": "false"
+        },
+        {
+          "command": "terraform.providers.openDocumentation",
           "when": "false"
         }
       ],
@@ -273,6 +282,10 @@
         {
           "command": "terraform.modules.openDocumentation",
           "when": "view == terraform.modules"
+        },
+        {
+          "command": "terraform.providers.openDocumentation",
+          "when": "view == terraform.providers"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -293,17 +293,17 @@
       "explorer": [
         {
           "id": "terraform.providers",
-          "name": "Terraform Provides",
+          "name": "Terraform Providers",
           "icon": "assets/icons/terraform.svg",
           "visibility": "collapsed",
-          "when": "terraform.showProviderView"
+          "when": "terraform.showTreeViews"
         },
         {
           "id": "terraform.modules",
           "name": "Terraform Module Calls",
           "icon": "assets/icons/terraform.svg",
           "visibility": "collapsed",
-          "when": "terraform.showModuleView"
+          "when": "terraform.showTreeViews"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -310,7 +310,7 @@
     "viewsWelcome": [
       {
         "view": "terraform.providers",
-        "contents": "The active editor cannot provide information about installed providers."
+        "contents": "The active editor cannot provide information about installed providers. [Learn more about providers](https://www.terraform.io/docs/language/providers/index.html). You may need to update your language server version."
       },
       {
         "view": "terraform.modules",
--- a/package.json
+++ b/package.json
@@ -285,7 +285,7 @@
         },
         {
           "command": "terraform.providers.openDocumentation",
-          "when": "view == terraform.providers"
+          "when": "view == terraform.providers && viewItem == moduleProviderHasDocs"
         }
       ]
     },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -128,7 +128,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<Terraf
     }),
     vscode.window.onDidChangeVisibleTextEditors(async (editors: vscode.TextEditor[]) => {
       const textEditor = editors.find((ed) => !!ed.viewColumn);
-      if (textEditor.document === undefined) {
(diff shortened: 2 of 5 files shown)


src/providers/providerProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+  public readonly onDidChangeTreeData = this.didChangeTreeData.event;
+
+  constructor(ctx: vscode.ExtensionContext, private handler: ClientHandler) {
+    ctx.subscriptions.push(
+      vscode.window.onDidChangeActiveTextEditor(async (event: vscode.TextEditor) => {
+        if (event && getActiveTextEditor()) {
+          this.refresh();
+        }
+      }),
+      vscode.commands.registerCommand('terraform.providers.openDocumentation', (module: ModuleProvider) => {
```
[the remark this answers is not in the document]
dbanck: Works, see e3b4388

src/providers/providerProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+  constructor(
+    public fullName: string,
+    public displayName: string,
+    public requiredVersion: string,
+    public installedVersion: string | undefined,
+    public docsLink: string | undefined,
+  ) {
+    super(`${displayName} ${requiredVersion}`, vscode.TreeItemCollapsibleState.None);
+
+    this.description = installedVersion ?? 'n.a.';
```
radeksimko, line 32: I would say that _if_ we are going to display anything in case of "not installed" providers, then it should be more obvious that the provider is not installed (from the text). I think that we could as well just not display anything at all?

Then perhaps once we add the capability to install providers via context menu we can adjust the UI accordingly.

package.json
```diff
@@ -284,6 +304,10 @@
       ]
     },
     "viewsWelcome": [
+      {
+        "view": "terraform.providers",
+        "contents": "The active editor cannot provide information about installed providers."
```
radeksimko, line 309:
```suggestion
        "contents": "The active editor cannot provide information about installed providers. [Learn more about providers](https://www.terraform.io/docs/language/providers/index.html). You may need to update your language server version."
```

radeksimko (approved):
Aside from my two minor in-line comments this LGTM, but James has more VS Code related experience here so he may be able to provide more relevant feedback.

src/providers/providerProvider.ts
dbanck: Should we maybe then just display the version constraints in the description (and not in the label) for now?

CleanShot 2021-11-16 at 11 46 52@2x

radeksimko: I think that installed versions are still useful, it's just the `n.a.` I was commenting on. I'd perhaps prefer empty string over that.

package.json
```diff
@@ -269,11 +278,22 @@
...
+          "command": "terraform.providers.openDocumentation",
+          "when": "view == terraform.providers && viewItem == moduleProviderHasDocs"
         }
       ]
     },
     "views": {
       "explorer": [
+        {
+          "id": "terraform.providers",
+          "name": "Terraform Provides",
```
jpogran, line 292: Is `Terraform Provides` a typo or intentional? I expected this to be `Providers`

src/extension.ts
```diff
@@ -133,6 +135,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<Terraf
       await updateLanguageServer(manifest.version, lsPath);
       await clientHandler.startClient();
       vscode.commands.executeCommand('setContext', 'terraform.showModuleView', true);
+      vscode.commands.executeCommand('setContext', 'terraform.showProviderView', true);
```
jpogran, line 138: I didn't think ahead when I made the module view and name the context name a less specific value. Do you think we should change this to `showTreeView` and make both providers key off that? The original purpose of using the context value here is to ensure that we only activate them when the LS is enabled and ready, so there really isn't a need for two values.

src/providers/providerProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+      version_constraint?: string;
+      docs_link?: string;
+    };
+  };
+  installed_providers: {
+    [provider: string]: string;
+  };
+}
+
+class ModuleProvider extends vscode.TreeItem {
```
jpogran, line 22: I think we discussed this in a different context and didn't come to a decision, but naming the treeitem `ModuleProvider` here confuses me with the `ModuleProvider` treedataprovider in `moduleProvider.ts`.

Maybe we need to append each thing with `TreeItem` and `TreeDataProvider` to disambiguate these classes since we overlap with Terraform terminology.

src/providers/providerProvider.ts
jpogran: In the Terraform Module Calls I returned an empty string if there was no version found.

If we intend to have a command to install a provider from a menu in this view, it makes sense to me for there to be nothing displayed in the description until it's installed, and in the tooltip have the constraint. To me, that would give me the visual indication that it's not installed if there is no resolved version there, and that I can right click on it or issue a command to have it installed.

src/extension.ts
dbanck: I like this idea, and I think there wouldn't be a case where we want to show just one of the views. So I'll rename the context value and remove the new one.

package.json
dbanck: Typo, good catch!

src/providers/providerProvider.ts
dbanck: Okay, I agree with that. I've moved the required version from the label into the tooltip and kept the description empty when the provider isn't installed.

CleanShot 2021-11-17 at 11 20 30@2x

radeksimko: I like that! 👍🏻 Looks quite clean.

package.json
radeksimko: I know the repetition of "Terraform" through both of our codebases is a common problem 😅  but I would argue that it may actually be relevant here as this is the title of the pane within the bigger otherwise not extension specific Explorer view.

I guess it may be obvious what kind of "providers" are these to most users, but we are sharing this space with other extensions in principle and so I'd be careful about the naming here.

src/providers/providerProvider.ts
dbanck: You're right; the naming is confusing. I've updated the class names and renamed the files. Do the new names work for you?

package.json
jpogran: I agree @radeksimko, I didn't mean to imply we remove the `Terraform` part. Meant just to call out the typo

jpogran (approved):
:shipit:

PR #851, opened 2021-11-12
Generate bug report command

jpogran:
Add a new command that records information from the current environment and generates a nicely formatted Github issue.

### DIFF
--- a/src/commands/generateBugReport.ts
+++ b/src/commands/generateBugReport.ts
@@ -1,0 +1,184 @@
+import child_process = require('child_process');
+import * as os from 'os';
+import * as vscode from 'vscode';
+
+interface TerraformInfo {
+  version: string;
+  platform: string;
+  outdated: boolean;
+}
+
+interface VSCodeExtension {
+  name: string;
+  publisher: string;
+  version: string;
+}
+
+export class GenerateBugReportCommand implements vscode.Disposable {
+  constructor(private ctx: vscode.ExtensionContext) {
+    this.ctx.subscriptions.push(
+      vscode.commands.registerCommand('terraform.generateBugReport', async () => {
+        const problemText = await vscode.window.showInputBox({
+          title: 'Generate a Bug Report',
+          prompt: 'Enter a short description of the problem or hit enter to submit now',
+          placeHolder: "For example: I'm having trouble getting autocomplete to work when I...",
+        });
+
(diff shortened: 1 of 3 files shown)


src/commands/generateBugReport.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    const spawn = child_process.spawnSync;
+    const child = spawn(terraformExe, cmdargs);
+    const response = child.stdout.toString();
+    const j = JSON.parse(response);
+    return {
+      version: j.terraform_version,
+      platform: j.platform,
+      outdated: j.terraform_outdated,
+    };
```
radeksimko, line 182: The `json` flag was actually unfortunately only added relatively recently (in 0.13), so passing that to older versions will result in error. See https://github.com/hashicorp/terraform/commit/d8e22f1804fcb99fc70dfbb4f54b9de28a92ebc2

We account for this possibility by falling back to plaintext regex parsing in tfexec https://github.com/hashicorp/terraform-exec/blob/750bc90690f17802eb236adf7336553ea947bcbe/tfexec/version.go#L57-L59
but naturally that's only available for Go, not for TypeScript.

We _could_ source Terraform version via LSP...
 - LS uses tfexec, which has a well tested canonical implementation
 - it most accurately reflects what LS can see (as e.g. `PATH` may differ between VS Code and LS?)

... however it increases the complexity in a bug reporting code path which we probably want to keep as simple as possible. Users may as well be reporting bugs with the LS - in which case we might not able to able obtain the version.

I'm fine with either option - but either way I think we should account for older Terraform versions.

If you decide to keep executing and parsing it from here directly then you can leverage the regular expressions and tests available in tfexec: https://github.com/hashicorp/terraform-exec/blob/750bc90690f17802eb236adf7336553ea947bcbe/tfexec/version.go#L110

jpogran: Great point about the older terraform cli interface, thanks! That's the kind of history I need :)

Adjusted it using your examples to try/catch and do the best it can with parsing strings. I agree, I don't think this feature is worth the effort/complexity of asking the LS, possibly in a case where we are already in an error state.

radeksimko (approved):
🚀 I have some ideas on how to improve this further, but I don't want perfect to be the enemy of good and this is already very good! 😄

PR #852, opened 2021-11-17
Use vscode.workspace.fs instead of fs package

### LINKED ISSUE #812: Use VS Code `workspace.fs` for filesystem operations
Follow up to #739 

--- 

## Problem Statement

Currently the extension uses the NodeJS "native" `fs` module to perform filesystem operations.

VS Code API provides [workspace.fs](https://code.visualstudio.com/api/references/vscode-api#FileSystem) for reading and writing files and folders from an uri.

It is possible that using the native VS Code FS API _could_ alleviate some of the FS related issues users have reported, such as https://github.com/hashicorp/vscode-terraform/issues/732

## Expected User Experience

No UX changes expected.

## Proposal

TODO

jpogran:
Replaces all usages of the fs package with the vscode.workspace.fs wrapper. This ensures safe usage of the filesystem and allows us in the future to move to remote extension use.

Fix bug where zip is not deleted after extraction. This also helps in the future when we minify the extension and del does not work.

Closes #812

### DIFF
--- a/src/languageServerInstaller.ts
+++ b/src/languageServerInstaller.ts
@@ -119,7 +119,7 @@ export class LanguageServerInstaller {
   async installPkg(release: Release): Promise<void> {
     const installDir = this.lsPath.installPath();
     const destination = path.resolve(installDir, `terraform-ls_v${release.version}.zip`);
-    vscode.workspace.fs.createDirectory(vscode.Uri.file(installDir)); // create install directory if missing
+    await vscode.workspace.fs.createDirectory(vscode.Uri.file(installDir)); // create install directory if missing
 
     const os = goOs();
     const arch = goArch();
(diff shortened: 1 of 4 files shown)


src/languageServerInstaller.ts
```diff
@@ -70,7 +68,7 @@ export class LanguageServerInstaller {
       // Most of the time, getLsVersion will produce "ENOENT: no such file or directory"
```
dbanck, line 68:
```suggestion
      // Most of the time, getLsVersion will produce "FileSystemError: FileNotFound"
```
I think `workspace.fs` is throwing `FileSystemError`s: https://code.visualstudio.com/api/references/vscode-api#FileSystemError

dbanck (approved):
Installation works well for me, and the `.zip` file is deleted, too 🎉 

I've two minor suggestions but feel free to merge anyway.

radeksimko:
> Replaces all usages of the fs package with the vscode.workspace.fs wrapper.

Just for posterity, some packages that the extension depends on still use the `fs` package, such as `js-releases`.
I'm assuming that could be addressed of as part of future work on bundling of LS though.

👍🏻 🚢

jpogran:
Updated commit message 😝 

> Replaces all usages _*inside extension*_ of the fs package with the `vscode.workspace.fs` wrapper.

PR #854, opened 2021-11-18
Fix extension ignore list for packaging

jpogran:
This fixes `testFixture` casing and updates the list with recently added files that do not need to be included in a VSIX, resulting in including 8 less files.

```
❯ npx vsce package
Executing prepublish script 'npm run vscode:prepublish'...

This extension consists of 1090 files, out of which 472 are JavaScript files. 

❯ npx vsce package
Executing prepublish script 'npm run vscode:prepublish'...

This extension consists of 1082 files, out of which 472 are JavaScript files.
```


dbanck (approved):
I'm looking forward to `esbuild`, so we can ignore `node_modules/` and `src/`.

PR #857, opened 2021-11-19
Fix terraform status bar

dbanck:
VSCode always hid the status bar because we never executed the command for getting the module callers as it was missing the `commandPrefix`.

### DIFF
--- a/src/test/integration/statusBar.test.ts
+++ b/src/test/integration/statusBar.test.ts
@@ -1,0 +1,32 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import { expect } from 'chai';
+import { terraformStatus } from '../../extension';
+import { getDocUri, open } from '../helper';
+import { sleep } from '../../utils';
+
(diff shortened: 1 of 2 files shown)


src/test/integration/statusBar.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  test('should create a status bar item', () => {
+    assert.notStrictEqual(terraformStatus, undefined);
+  });
+
+  test('should create a status bar with the root module label', async () => {
+    const documentUri = getDocUri('sample.tf');
+    await open(documentUri);
+    await sleep(500);
```
dbanck, line 20: @jpogran Do you have any idea how to avoid using `sleep` here to wait for the status bar to update?

jpogran (approved):

src/test/integration/statusBar.test.ts
jpogran: Resolved by using the function and awaiting

PR #858, opened 2021-11-22
Remove public api

dbanck:
This PR depends on #857, which needs to be merged first.

Our `activate` function returned `moduleCallers` and `clientHandler` which enabled other extensions to use them. We didn't intend this behavior and don't want to expose any public API yet.

See:
> https://code.visualstudio.com/api/references/vscode-api#extensions
> Extension writers can provide APIs to other extensions by returning their API public surface from the `activate` call.

This PR removes the return value from `activate` and refactors the `workspaces.spec.ts` test, which used this API.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -152,30 +152,33 @@ export async function deactivate(): Promise<void> {
 
 export async function updateTerraformStatusBar(documentUri: vscode.Uri): Promise<void> {
   const client = clientHandler.getClient();
-  const initSupported = clientHandler.clientSupportsCommand(`${client.commandPrefix}.terraform-ls.terraform.init`);
 
-  if (initSupported) {
-    const moduleUri = Utils.dirname(documentUri);
+  if (client) {
+    const initSupported = clientHandler.clientSupportsCommand(`${client.commandPrefix}.terraform-ls.terraform.init`);
 
-    try {
-      const response = await moduleCallers(moduleUri.toString());
+    if (initSupported) {
+      const moduleUri = Utils.dirname(documentUri);
 
-      if (response.moduleCallers.length === 0) {
-        const dirName = Utils.basename(moduleUri);
+      try {
+        const response = await moduleCallers(moduleUri.toString());
 
-        terraformStatus.text = `$(refresh) ${dirName}`;
-        terraformStatus.color = new vscode.ThemeColor('statusBar.foreground');
-        terraformStatus.tooltip = `Click to run terraform init`;
-        terraformStatus.command = 'terraform.initCurrent';
-        terraformStatus.show();
-      } else {
+        if (response.moduleCallers.length === 0) {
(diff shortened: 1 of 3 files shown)


radeksimko (approved):
Just for context - I believe the original intention behind the public API exposure was testability, but naturally it can be tested as a decoupled unit too, as you have demonstrated here. 👍🏻 

I agree with the conservative decision of not exposing any API _yet_ until we know what the consumer needs are. I reckon we'll investigate some use cases as part of https://github.com/hashicorp/vscode-terraform/issues/692

src/extension.ts
```diff
@@ -160,16 +152,19 @@ export async function deactivate(): Promise<void> {
...
-  const initSupported = clientHandler.clientSupportsCommand(`${client.commandPrefix}.terraform-ls.terraform.init`);
-  if (initSupported) {
-    const client = clientHandler.getClient();
-    const moduleUri = Utils.dirname(documentUri);
 
-    if (client) {
+  if (client) {
+    const initSupported = clientHandler.clientSupportsCommand(`${client.commandPrefix}.terraform-ls.terraform.init`);
+
+    if (initSupported) {
```
jpogran, line 159: Maybe use this chance to return early by inverting the if statement and reduce nesting?

dbanck: Yes! Good point

src/extension.ts
```diff
@@ -261,11 +257,9 @@ async function modulesCallersCommand(languageClient: TerraformLanguageClient, mo
   return execWorkspaceCommand(languageClient.client, requestParams);
 }
 
-async function moduleCallers(
-  languageClient: TerraformLanguageClient,
-  moduleUri: string,
-): Promise<moduleCallersResponse> {
-  const response = await modulesCallersCommand(languageClient, moduleUri);
+export async function moduleCallers(moduleUri: string): Promise<moduleCallersResponse> {
+  const client = clientHandler.getClient();
```
jpogran, line 261: We should check for an undefined client here

src/extension.ts
```diff
@@ -160,31 +152,35 @@ export async function deactivate(): Promise<void> {
...
+      terraformStatus.command = 'terraform.initCurrent';
+      terraformStatus.show();
+    } else {
+      terraformStatus.hide();
+      terraformStatus.text = '';
     }
+  } catch (err) {
+    vscode.window.showErrorMessage(err);
+    reporter.sendTelemetryException(err);
+    terraformStatus.hide();
```
jpogran, line 183: Nice!

jpogran (approved):
Small suggestion to check for undefined, otherwise approved!

src/extension.ts
dbanck: Good catch! I've added a check.

jpogran:
🚢

PR #859, opened 2021-11-23
Silent LS Installer

jpogran:
Testable LS Installer that does not prompt the user during detection and installation of the LS


jpogran:
Superseded by #868

PR #862, opened 2021-11-29
update mocha

jpogran:
- w
- Update mocha to 9.1.3


radeksimko (approved):
LGTM - any particular motivation for the bump or just "passing by" kind of change? 😄 

On a related note I'm wondering if it's worth tuning the dependabot versioning strategy, so we receive updates like these automatically as well...

https://github.com/hashicorp/vscode-terraform/blob/aa3ee19c3cd26b3f6cef98c048685048531be328/.github/dependabot.yml#L4

dbanck (approved):

jpogran:
@radeksimko Ran into this working on the ls installer improvements. NPM reports a vulnerability in ansi-regex, which mocha uses. Updating to a recent version enough to resolve requires bumping to a major version, thus a 'breaking' change so I thought to make it it's own PR

PR #863, opened 2021-11-30
Use more strict TypeScript configuration

dbanck:
This PR enables strict mode for TypeScript, resulting in a wide range of type checking behavior that guarantees program correctness.

I tried to do one commit per file, sometimes explaining my choices.

### DIFF
--- a/src/commands/generateBugReport.ts
+++ b/src/commands/generateBugReport.ts
@@ -169,50 +169,50 @@ Outdated:\t${info.outdated}
 
   async getRuntimeInfo(): Promise<TerraformInfo> {
     const terraformExe = 'terraform';
-
     const spawn = child_process.spawnSync;
 
-    try {
-      const child = spawn(terraformExe, ['version', '-json']);
-      const response = child.stdout.toString();
-      const j = JSON.parse(response);
+    // try to get version from a newer terraform binary
+    const resultJson = spawn(terraformExe, ['version', '-json']);
+    if (resultJson.error === undefined) {
+      try {
+        const response = resultJson.stdout.toString();
+        const j = JSON.parse(response);
 
-      return {
-        version: j.terraform_version,
-        platform: j.platform,
-        outdated: j.terraform_outdated,
-      };
-    } catch (err) {
-      if (err.status === 0) {
         return {
-          version: 'Not found',
-          platform: 'Not found',
-          outdated: false,
+          version: j.terraform_version,
+          platform: j.platform,
+          outdated: j.terraform_outdated,
         };
+      } catch {
+        // fall through
       }
     }
 
-    try {
-      // assume older version of terraform which didn't have json flag
-      const child = spawn(terraformExe, ['version']);
-      const response = child.stdout.toString() || child.stderr.toString();
-      const regex = new RegExp('v?(?<version>[0-9]+(?:.[0-9]+)*(?:-[A-Za-z0-9.]+)?)');
-      const matches = regex.exec(response);
+    // try an older binary without the json flag
+    const result = spawn(terraformExe, ['version']);
+    if (result.error === undefined) {
+      try {
+        const response = result.stdout.toString() || result.stderr.toString();
+        const regex = new RegExp('v?(?<version>[0-9]+(?:.[0-9]+)*(?:-[A-Za-z0-9.]+)?)');
+        const matches = regex.exec(response);
 
-      const version = matches[1];
-      const platform = response.split('\n')[1].replace('on ', '');
+        const version = matches && matches.length > 1 ? matches[1] : 'Not found';
+        const platform = response.split('\n')[1].replace('on ', '');
 
-      return {
-        version: version,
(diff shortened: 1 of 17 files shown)


src/commands/generateBugReport.ts
```diff
@@ -169,50 +169,50 @@ Outdated:\t${info.outdated}
...
-
-      return {
-        version: j.terraform_version,
-        platform: j.platform,
-        outdated: j.terraform_outdated,
-      };
-    } catch (err) {
-      if (err.status === 0) {
+    // try to get version from a newer terraform binary
+    const resultJson = spawn(terraformExe, ['version', '-json']);
```
jpogran, line 175: This may be out of scope of this PR because it's a refactoring and not strictly related to Typescript strict checking, but since we have to change the set of code that executes a binary in two places, I think it would make sense to extract to a single function that we can reference in both places.

src/extension.ts
```diff
@@ -287,14 +298,14 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
         response.moduleCallers.map((m) => m.uri),
         { canPickMany: false },
       );
-      selectedModule = selected[0];
+      selectedModule = selected ?? moduleUri.toString();
```
jpogran, line 301: I don't understand why we're changing this line

src/providers/moduleCalls.ts
```diff
@@ -122,7 +138,10 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
         arguments: [`uri=${documentURI}`],
       };
 
-      const response = await handler.client.sendRequest(ExecuteCommandRequest.type, params);
+      const response = await handler.client.sendRequest<ExecuteCommandParams, ModuleCallsResponse, void>(
```
jpogran, line 141: Nice!

src/showReferences.ts
```diff
@@ -25,7 +27,7 @@ export class ShowReferencesFeature implements StaticFeature {
 
   constructor(private _client: BaseLanguageClient) {}
 
-  public fillClientCapabilities(capabilities: ClientCapabilities): void {
+  public fillClientCapabilities(capabilities: ClientCapabilities & ExperimentalClientCapabilities): void {
```
jpogran, line 30: Curious how add `ExperimentalClientCapabilities` works/helps?

src/commands/generateBugReport.ts
dbanck: Thanks for your review!

I'm not sure which part of the code you're suggesting to extract here. Can you explain?

src/extension.ts
dbanck: `showQuickPick` will return a `string` instead of an array when used with the `canPickMany:  false` option. So, calling `selected[0]` will return the first letter of the uri, which seemed wrong.

`showQuickPick` can also return `undefined`. That's why I'm using `moduleUri.toString()` as a fallback here.

src/showReferences.ts
dbanck: `ClientCapabilities` does not define a structure for `experimental` and just defines it as an empty object. 

Setting the client id, will result in an error:
```
capabilities['experimental']['showReferencesCommandId'] = CLIENT_CMD_ID;
    ^^^
    Property 'showReferencesCommandId' does not exist on type '{}'.
```

Combining both interfaces will add some type information for `experimental`. This might not be the most elegant solution, but the only way to extend `ClientCapabilities` for now.

src/extension.ts
jpogran: Great catch on the behavior of `canPickMany` here!

Thinking through the process here, undefined is returned when the User does not answer the quickpick. 99% of the time that means the user canceled the operation, not that they want the operation to continue. This most often happens when you exit out of a quickpick when you don't like the choices given. I think in that case, that means we should not continue. In this case, moduleUri is the open active text document, which may not be something related to a module. 

I may be off base here.

src/commands/generateBugReport.ts
jpogran: I'm referring to the parts where we detect the version of terraform and the version of terraform-ls. Different places in the extension, but nearly the same code to be executed to try/catch around the version parameter being present and/or json being accepted.

src/extension.ts
dbanck: That's a very valid point you're bringing up. So, do we instead want to check for undefined here and return early?

src/commands/generateBugReport.ts
dbanck: You're right. It would make sense to consolidate this into a single helper function. However, since this refactoring might also affect parts of the language server installer, I suggest we do it in an extra PR after everything has been merged.

src/extension.ts
jpogran: I think we should

dbanck: I've updated the code accordingly.

jpogran (approved):

PR #868, opened 2021-12-08
silent staged installer

jpogran:
- Silent Staged Installer
- add jest
- add tests
- make mocha play nice with jest
- add tests


dbanck (commented):
Awesome work! It's nice to have the tests covering all essential aspects of this feature.

My comments are only TypeScript / Jest related right now, and I might add more as I understand how everything ties together. I'm open to discussing anything, and we can go through typing the test files together if you want.

src/test/unit/installer.test.ts
```diff
@@ -0,0 +1,49 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck ignore type checking in test files
+import { installTerraformLS } from '../../installer/installer';
+import { reporter } from '../../__mocks__/reporter';
+import { pathExists } from '../../installer/detector';
+import { window } from '../../__mocks__/vscode';
+
+jest.mock('../../installer/detector');
+describe('ls installer', () => {
+  describe('should install', () => {
```
dbanck, line 10:
```suggestion
  describe('installTerraformLS', () => {
```

src/test/unit/installer.test.ts
```diff
@@ -0,0 +1,49 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck ignore type checking in test files
+import { installTerraformLS } from '../../installer/installer';
+import { reporter } from '../../__mocks__/reporter';
+import { pathExists } from '../../installer/detector';
+import { window } from '../../__mocks__/vscode';
+
+jest.mock('../../installer/detector');
+describe('ls installer', () => {
```
dbanck, line 9: I would remove this `describe` block here, too.

src/test/unit/installer.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        download: jest.fn(),
+        verify: jest.fn(),
+        unpack: jest.fn(),
+      };
+
+      pathExists.mockImplementationOnce(() => {
+        return true;
+      });
+
+      window.withProgress.mockImplementationOnce(() => {
```
dbanck, line 36: There are a few things we need to do, to get this working.

First, since we're calling
```ts
vscode.workspace.fs.delete(vscode.Uri.file(zipfile))
```
inside the installer, I would add `delete: jest.fn(),` to our vscode workspace fs mock.

`withProgress` calls the passed `task` function with `vscode.Progress` and a cancellation token, so we need to replicate this in our test:
```ts
const report = jest.fn();
const token = {
  isCancellationRequested: false,
  onCancellationRequested: jest.fn(),
};
withProgress.mockImplementationOnce(async (_, task) => {
  task({ report }, token);
});
```

We can now use the expects, you already added and add even more:

```ts
expect(release.getBuild).toBeCalledTimes(1);
expect(withProgress).toBeCalledTimes(1);
expect(withProgress).toHaveBeenCalledWith(
  {
    cancellable: false,
    location: vscode.ProgressLocation.Window,
    title: 'Installing terraform-ls',
  },
  expect.any(Function),
);

expect(release.download).toBeCalledTimes(1);
expect(release.verify).toBeCalledTimes(1);
expect(release.unpack).toBeCalledTimes(1);
expect(vscode.workspace.fs.delete).toBeCalledTimes(1);
expect(report).toHaveBeenCalledTimes(4);
```
We could even test the arguments to all calls in greater detail, if we want.

src/test/unit/installer.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('ls installer', () => {
+  describe('should install', () => {
+    let temp;
+
+    beforeEach(() => {
+      temp = require('temp').track();
+    });
+
+    test('when valid', async () => {
+      const installPath = temp.path('foo');
```
dbanck, line 18: We're mocking everything related to `installPath` / fs, so it should be safe to use a string here and get rid of `temp`

src/installer/detector.ts
```diff
@@ -0,0 +1,66 @@
...
+): Promise<Release> {
+  const userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscodeVersion}`;
+
+  // Take the user requested version and query the hashicorp release site
+  try {
+    const release = await getRelease('terraform-ls', versionString, userAgent);
+    console.log(`Found Terraform language server version ${release.version} which satisfies range '${versionString}'`);
+    return release;
+  } catch (err) {
+    if (versionString == DEFAULT_LS_VERSION) {
```
dbanck, line 38:
```suggestion
    if (versionString === DEFAULT_LS_VERSION) {
```
We could add a eslint rule for checking this

tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "outDir": "out",
     "rootDir": "src",
     "sourceMap": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
dbanck, line 9: Why do we need `skipLibCheck`?

dbanck (commented):
I dug into the implementation, and it looks really sound! Just a few minor suggestions. Most of my feedback is related to the style and typing of the unit tests, and we can iterate on that together if you like.

The `updater.test.ts` is the last file I need to look at, but I thought we might do this in a paring session since it's the biggest of the three tests.

src/test/unit/detector.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+      const result = await getRequiredVersionRelease('10000.24.0', '2.16.0',  '1.66.0');
+
+      expect(result).toStrictEqual(expected);
+      expect(getRelease).toBeCalledTimes(2);
+    });
+  });
+
+  describe('terraform detector', () => {
+    let temp;
```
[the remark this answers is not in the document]
dbanck: After looking further into the test, I think we can get rid of `temp`.

src/extension.ts
```diff
@@ -124,7 +125,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         }
       }
     }),
-    vscode.window.onDidChangeVisibleTextEditors(async (editors: vscode.TextEditor[]) => {
+    vscode.window.onDidChangeVisibleTextEditors(async (editors: readonly vscode.TextEditor[]) => {
```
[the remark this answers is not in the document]
jpogran: The @types/vscode package was updated in package-lock.json, which specifies that that parameter is readonly. If I don't specify it, typescript throws an error

tsconfig.json
jpogran: If we do not include that, mocha throws exceptions about the included types:

```
❯ npm run test

> terraform@2.17.0 pretest C:\Users\james\src\hashicorp\vscode-terraform
> npm run compile && npm run lint


> terraform@2.17.0 compile C:\Users\james\src\hashicorp\vscode-terraform
> tsc -p ./

node_modules/@types/jest/index.d.ts:35:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'beforeEach' must be of type 'HookFunction', but here has type 'Lifecycle'.

35 declare var beforeEach: jest.Lifecycle;
               ~~~~~~~~~~

  node_modules/@types/mocha/index.d.ts:2624:13
```

jest.config.js
```diff
@@ -0,0 +1,11 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
```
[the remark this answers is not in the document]
jpogran: I trusted `npx ts-jest config:init` from https://kulshekhar.github.io/ts-jest/docs/getting-started/installation 😂.

If we're using ts-test, how does this affect things?

src/extension.ts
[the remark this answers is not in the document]
dbanck: Makes sense! I missed the update of the `@types/vscode` package.

tsconfig.json
dbanck: I see, thanks for the explanation. We now have conflicting type declarations coming from mocha and jest.

jest.config.js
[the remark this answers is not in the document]
dbanck: Thanks for the link, I didn't know `ts-jest` adds additional configuration options.

dbanck (approved):
Awesome job! Thank you for taking the time to update all the tests! 🎉 
I think the code base is now much more sound without any `ts-nocheck` files.

I've added minor style nitpicks. Feel free to merge after addressing those.

PR #870, opened 2021-12-16
Fix Detecting Terraform files in status calls

jpogran:
This modifies the moduleCalls and moduleProviders views and the updateTerraformStatusBar function to detect if the open file is an `Untitled` file or other type of file and return early instead of trying to determine Terraform module information.

Previously the editor would attempt to determine terraform module status whenever a file was opened. This worked well if a Terraform file was opened, but would throw exceptions if a file was opened that did not contain Terraform code. This includes `Untitled` files which are open files that are not saved yet and don't exist on the filesystem yet. The fix is to ignore any non Terraform file.

This contains a naive check to determine if a file is a Terraform file, which checks the file extension for `tf`. This suits for now but possibly might need to be expanded in the future if the extension supports more Terraform file types.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -126,12 +126,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     }),
     vscode.window.onDidChangeVisibleTextEditors(async (editors: vscode.TextEditor[]) => {
       const textEditor = editors.find((ed) => !!ed.viewColumn);
-
-      if (!isTerraformFile(textEditor?.document)) {
+      if (textEditor?.document === undefined) {
         return;
       }
 
-      await updateTerraformStatusBar(textEditor?.document.uri);
+      if (!isTerraformFile(textEditor.document)) {
+        return;
+      }
+
+      await updateTerraformStatusBar(textEditor.document.uri);
     }),
   );
 
@@ -156,11 +159,7 @@ export async function deactivate(): Promise<void> {
   return clientHandler.stopClient();
 }
 
-export async function updateTerraformStatusBar(documentUri?: vscode.Uri): Promise<void> {
-  if (documentUri === undefined) {
-    return;
-  }
-
+export async function updateTerraformStatusBar(documentUri: vscode.Uri): Promise<void> {
   const client = clientHandler.getClient();
   if (client === undefined) {
     return;
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,7 +10,7 @@ import { ModuleCallsDataProvider } from './providers/moduleCalls';
 import { ModuleProvidersDataProvider } from './providers/moduleProviders';
 import { ServerPath } from './serverPath';
 import { SingleInstanceTimeout } from './utils';
-import { config, getActiveTextEditor } from './vscodeUtils';
+import { config, getActiveTextEditor, isTerraformFile } from './vscodeUtils';
 
 const brand = `HashiCorp Terraform`;
 const outputChannel = vscode.window.createOutputChannel(brand);
@@ -129,6 +129,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
       if (textEditor?.document === undefined) {
         return;
       }
+
+      if (!isTerraformFile(textEditor?.document)) {
+        return;
+      }
+
       await updateTerraformStatusBar(textEditor.document.uri);
     }),
   );
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -126,15 +126,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     }),
     vscode.window.onDidChangeVisibleTextEditors(async (editors: vscode.TextEditor[]) => {
       const textEditor = editors.find((ed) => !!ed.viewColumn);
-      if (textEditor?.document === undefined) {
-        return;
-      }
 
       if (!isTerraformFile(textEditor?.document)) {
         return;
       }
 
-      await updateTerraformStatusBar(textEditor.document.uri);
+      await updateTerraformStatusBar(textEditor?.document.uri);
     }),
   );
 
@@ -159,7 +156,11 @@ export async function deactivate(): Promise<void> {
   return clientHandler.stopClient();
 }
 
-export async function updateTerraformStatusBar(documentUri: vscode.Uri): Promise<void> {
+export async function updateTerraformStatusBar(documentUri?: vscode.Uri): Promise<void> {
+  if (documentUri === undefined) {
+    return;
+  }
+
   const client = clientHandler.getClient();
   if (client === undefined) {
     return;
--- a/src/providers/moduleCalls.ts
+++ b/src/providers/moduleCalls.ts
@@ -3,7 +3,7 @@ import * as vscode from 'vscode';
 import { ExecuteCommandParams, ExecuteCommandRequest } from 'vscode-languageclient';
 import { Utils } from 'vscode-uri';
 import { ClientHandler } from '../clientHandler';
-import { getActiveTextEditor } from '../vscodeUtils';
+import { getActiveTextEditor, isTerraformFile } from '../vscodeUtils';
 
 interface ModuleCall {
   name: string;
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -129,6 +129,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
       if (textEditor?.document === undefined) {
         return;
       }
+
+      if (!isTerraformFile(textEditor?.document)) {
```
dbanck, line 133:
```suggestion
      if (!isTerraformFile(textEditor.document)) {
```

We've checked in line 129 that `textEditor` and `textEditor.document` is defined, so it's safe to access it without the `?` here.

src/vscodeUtils.ts
```diff
@@ -25,7 +25,32 @@ export function getWorkspaceFolder(folderName: string): vscode.WorkspaceFolder |
...
 export function getActiveTextEditor(): vscode.TextEditor | undefined {
-  return vscode.window.visibleTextEditors.find((textEditor) => !!textEditor.viewColumn);
+  const activeEditor = vscode.window.visibleTextEditors.find((textEditor) => !!textEditor.viewColumn);
+  return activeEditor;
+}
+
+/*
+  Detects whether this is a Terraform file we can perform operations on
+ */
+export function isTerraformFile(document: vscode.TextDocument): boolean {
```
dbanck, line 35:
```suggestion
export function isTerraformFile(document?: vscode.TextDocument): boolean {
```

I assume `document` can be `undefined` as we're checking this in line 36?

src/vscodeUtils.ts
```diff
@@ -25,7 +25,32 @@ export function getWorkspaceFolder(folderName: string): vscode.WorkspaceFolder |
 // because it also contains Output panes which are considered editors
 // see also https://github.com/microsoft/vscode/issues/58869
 export function getActiveTextEditor(): vscode.TextEditor | undefined {
-  return vscode.window.visibleTextEditors.find((textEditor) => !!textEditor.viewColumn);
+  const activeEditor = vscode.window.visibleTextEditors.find((textEditor) => !!textEditor.viewColumn);
```
dbanck, line 28: What's the reason to extract this into a variable before returning?

src/providers/moduleCalls.ts
```diff
@@ -83,7 +83,18 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
...
       vscode.window.onDidChangeActiveTextEditor(async (event: vscode.TextEditor | undefined) => {
-        if (event && getActiveTextEditor()) {
+        const activeEditor = getActiveTextEditor();
+
+        const document = activeEditor?.document;
+        if (document === undefined) {
+          return;
+        }
+
+        if (!isTerraformFile(document)) {
```
dbanck, line 93: I think we can pass  `activeEditor?.document` directly into this function since it can handle an undefined value. This allows us to remove the check for undefined above.

```ts
 if (!isTerraformFile(activeEditor?.document)) {
```

This also applied to the next check and the two in `src/providers/moduleProviders.ts`

dbanck (changes requested):
Nice work! I've tested different files with this, and it's much more robust now.

I've added a few suggestions about handling `undefined`.

src/vscodeUtils.ts
jpogran: Do we want to allow no parameter to be passed? I thought that's what `?` did if added like this. My intention is to check if undefined, not allow parameter-less invocation

src/vscodeUtils.ts
jpogran: Was part of debugging, I can revert back

src/vscodeUtils.ts
jpogran: Ah, I see. If changing usage to pass a value directly that may be undefined, would have to do this. Changing now

src/extension.ts
dbanck: Sorry, I did not mean to suggest removing the check above here. I just meant to remove the `?` in this call (like the proposed change).

We still need to make sure that `textEditor.document` isn't `undefined`, so we can access and pass `uri` to `updateTerraformStatusBar`.

So what I'm suggesting now is: keeping the `if` above as before, removing the `?` (as proposed) and undo the changes to `updateTerraformStatusBar`.

src/providers/moduleCalls.ts
```diff
@@ -109,13 +115,14 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
 
   async getModules(): Promise<ModuleCallItem[]> {
     const activeEditor = getActiveTextEditor();
-    if (activeEditor === undefined) {
-      return Promise.resolve([]);
-    }
 
-    const document = activeEditor.document;
+    const document = activeEditor?.document;
```
dbanck, line 119: We can simplify this here, too:

```ts
if (!isTerraformFile(activeEditor?.document)) {
 return [];
}
```

src/providers/moduleProviders.ts
```diff
@@ -46,7 +46,18 @@ export class ModuleProvidersDataProvider implements vscode.TreeDataProvider<Modu
   constructor(ctx: vscode.ExtensionContext, private handler: ClientHandler) {
     ctx.subscriptions.push(
       vscode.window.onDidChangeActiveTextEditor(async (event: vscode.TextEditor | undefined) => {
-        if (event && getActiveTextEditor()) {
+        const activeEditor = getActiveTextEditor();
+
+        const document = activeEditor?.document;
```
dbanck, line 51: And here:

```ts
if (!isTerraformFile(activeEditor?.document)) {
 return;
}
```

src/providers/moduleProviders.ts
```diff
@@ -82,6 +93,10 @@ export class ModuleProvidersDataProvider implements vscode.TreeDataProvider<Modu
       return [];
```
dbanck, line 93: And here:

```ts
if (!isTerraformFile(activeEditor?.document)) {
 return [];
}
```

dbanck (changes requested):
Thanks for addressing my feedback. I've marked the other three parts where the same simplification can be applied.

src/extension.ts
jpogran: Reworked as suggested

src/providers/moduleCalls.ts
jpogran: Reworked as suggested

src/providers/moduleProviders.ts
jpogran: Reworked as suggested

src/providers/moduleProviders.ts
jpogran: Reworked as suggested

dbanck (approved):
🎉

src/vscodeUtils.ts
```diff
@@ -28,6 +28,30 @@ export function getActiveTextEditor(): vscode.TextEditor | undefined {
...
+  if (document.fileName.endsWith('tf')) {
+    // For the purposes of this extension, anything with the tf file
+    // extension is a Terraform file
+    return true;
+  }
+
+  // be safe and default to false
+  return false;
+}
+
```
radeksimko, line 54: Two points/questions:

1) Isn't Untitled document also valid? The user can set the syntax (language ID) explicitly and AFAIK even these documents are normally sent to the LS and LS should be able to process them as virtual documents, even when they don't (yet) exist on the OS filesystem.
2) Is there any way we can avoid this drifting from other logic which tracks what files do we recognize, or ideally avoids the duplication entirely?
   - activation events: https://github.com/hashicorp/vscode-terraform/blob/7dc861c1b6d18d0e5416251fe0528c701b5f6631/package.json#L37-L40
   - association of language IDs with file extensions: https://github.com/hashicorp/vscode-terraform/blob/7dc861c1b6d18d0e5416251fe0528c701b5f6631/package.json#L45-L69

IMHO we should be checking the language ID here at the least (not file extensions) and relatedly check `terraform-vars` in addition to `terraform`.

radeksimko (commented):
Sorry for the delayed review.

PR #882, opened 2022-01-03
Fix creating process with environment variables

jpogran:
This ensures the child process we create for the Language Server contains the process environment variables needed to find other binaries on the host.

When creating the process for the Language Server, we relied on process.spawn to copy the environment variables to the child process. This works in most cases, but fails to contain the host environment variables in certain scenarios. For example, opening VS Code from the macOS Launcher. This leaves our process with important environment variables like PATH, and results in failures down the line attempting to use other binaries like terraform.

These errors are mostly silent, leaving the user confused why something isn't working. It's also hard to debug, because the correct variables (especially PATH) are present and work outside our app.

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -108,17 +108,62 @@ export class ClientHandler {
   private async getServerOptions(): Promise<ServerOptions> {
     const cmd = await this.lsPath.resolvedPathToBinary();
     const serverArgs = config('terraform').get<string[]>('languageServer.args', []);
+    const procEnv = this.getProcessEnvironment();
+
     const executable: Executable = {
       command: cmd,
       args: serverArgs,
-      options: {},
+      options: {
+        env: procEnv,
+      },
     };
+
     const serverOptions: ServerOptions = {
       run: executable,
       debug: executable,
     };
+
     this.outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
     return serverOptions;
+  }
+
+  private getProcessEnvironment() {
+    // deep object clone without the bother of custom traversal
+    const procEnv = JSON.parse(JSON.stringify(process.env));
+
+    // It's unlilely but possible that there is no PATH set. Object property
+    // names can be case sensitive depending on the OS, so we cannot rely on
+    // process.env.PATH working everywhere. It could be called PATH or Path or path
+    // and indexing would fail to find it.
+    if (procEnv.PATH === undefined) {
+      let envPath = '';
+      // Look through all of the environment names looking for PATH in a case
+      // insensitive way and remove the conflicting env var.
+      Object.keys(procEnv).forEach(function (keyname) {
+        if (keyname.match(/^PATH$/i)) {
+          envPath = procEnv[keyname] ?? '';
+          procEnv[keyname] = undefined;
+        }
+      });
+      procEnv.PATH = envPath;
+    }
+
+    // Child spawn environment variables can have undefined or null value
+    // In our case we can remove these elements from the Object directly and not
+    // worry about them
+    try {
+      const propNames = Object.getOwnPropertyNames(procEnv);
+      for (let i = 0; i < propNames.length; i++) {
+        const propName = propNames[i];
+        if (procEnv[propName] === null || procEnv[propName] === undefined) {
+          delete procEnv[propName];
+        }
+      }
+    } catch (err) {
+      console.log(err);
+    }
+
+    return procEnv;
   }
 
   private getInitializationOptions(commandPrefix: string) {
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -128,8 +128,8 @@ export class ClientHandler {
   }
 
   private getProcessEnvironment() {
-    // deep object clone without the bother of custom traversal
-    const procEnv = JSON.parse(JSON.stringify(process.env));
+    // shallow copy environment variables
+    const procEnv = { ...process.env };
 
     // It's unlilely but possible that there is no PATH set. Object property
     // names can be case sensitive depending on the OS, so we cannot rely on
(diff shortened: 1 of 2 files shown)


src/clientHandler.ts
```diff
@@ -108,19 +108,64 @@ export class ClientHandler {
...
+      // Look through all of the environment names looking for PATH in a case
+      // insensitive way and remove the conflicting env var.
+      Object.keys(procEnv).forEach(function (keyname) {
+        if (keyname.match(/^PATH$/i)) {
+          envPath = procEnv[keyname] ?? '';
+          procEnv[keyname] = undefined;
+        }
+      });
+      procEnv.PATH = envPath;
+    }
```
radeksimko, line 149: Is there any particular situation which you ran into that motivated you to add this?

I'm thinking that _if_ someone already has wrong casing, this must be causing various other problems, so I'm not sure we'd be helpful in hiding those problems in such a scenario?

src/clientHandler.ts
```diff
@@ -108,19 +108,64 @@ export class ClientHandler {
...
       debug: executable,
     };
+
     this.outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
     return serverOptions;
   }
 
+  private getProcessEnvironment() {
+    // deep object clone without the bother of custom traversal
+    const procEnv = JSON.parse(JSON.stringify(process.env));
```
radeksimko, line 132: This looks a bit awkward to me, but I'm not really familiar with data copying practices in JavaScript myself, so I'd defer to you and @dbanck here.

radeksimko (commented):
> When creating the process for the Language Server, we relied on process.spawn to copy the environment variables to the child process. This works in most cases, but fails to contain the host environment variables in certain scenarios. For example, opening VS Code from the macOS Launcher. This leaves our process with important environment variables like PATH, and results in failures down the line attempting to use other binaries like terraform.

Do you mind outlining the scenarios (which are being addressed by this PR) in more detail here? I don't fully understand how this would address the problem with macOS Dock/Spotlight, because it's VS Code that's being launched from there and if VS Code doesn't have the right ENV variables I don't see how any extension (which runs presumably as a child process of VS Code, hence inheriting VS Code's own environment) can fix that?

src/clientHandler.ts
dbanck: It's a useful way to deep clone an object in JavaScript, but it won't clone any function or symbol properties (which isn't necessary here). We might even consider using `Object.assign` for shallow cloning here because `process.env` should just contain strings and not any objects/arrays.

src/clientHandler.ts
jpogran: Yes, past experience with the puppet extension.

It's not 'someone' setting a 'wrong' casing. It's that the range of OS we support that each have different rules on casing and set key `PATH` casing differently. If we use `process.env.PATH` in code, but it is `process.env.Path` and access it on nix we won't get the value because the casing is different, where on windows we will get the correct value.

If we sanitize our inputs, we can at least be certain what is being passed down to the LS, which will aid debugging bug reports in the future.

Supporting documentation:

[child._process](https://github.com/nodejs/node/blob/master/doc/api/child_process.md):

> On Windows, environment variables are case-insensitive. Node.js lexicographically sorts the env keys and uses the first one that case-insensitively matches. Only first (in lexicographic order) entry will be passed to the subprocess. This might lead to issues on Windows when passing objects to the env option that have multiple variants of the same key, such as PATH and Path.

[process.env](https://nodejs.org/api/process.html#processenv)

>On Windows operating systems, environment variables are case-insensitive.
>
> import { env } from 'process';
> 
> env.TEST = 1;
> console.log(env.test);
> // => 1

src/clientHandler.ts
jpogran: I've used `Object.assign` and prototype builtins for shallow cloning objects in the past with varying success. TS strict does not like the `any` usage for prototype proprty discovery and doing a search on `Object.assign` finds [several kinds of problems](https://stackoverflow.com/questions/34504682/js-does-object-assign-create-deep-copy-or-shallow-copy) with using it and shadow copies.

For me, I chose `JSON.parse` largely from subjective rather than objective reasons. I'd rather use a built-in method here for this than custom code I wrote, or `Object.assign` which changes behavior based on what's stored in the object and what node version we run.

src/clientHandler.ts
radeksimko: Hmm, OS differences, I see. 🤔 

It looks like Go stdlib assumes more explicitly that on Windows it's always all lowercase:
https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/os/exec/lp_windows.go;l=87

Is there a reason why not follow that? It would probably simplify the code a bit and make it easier to read/understand.

src/clientHandler.ts
radeksimko: I see, that's fair! I suppose it looks easy to read at least - I have seen some other snippets of code copying data in JS which look far more complicated! 😅

radeksimko: My only worry was perhaps the performance hit with the string conversions, but I suppose most of the time there won't be as many variables to make this a big deal.

src/clientHandler.ts
radeksimko: btw. I'm somewhat surprised the NodeJS stdlib doesn't handle this automatically.

src/clientHandler.ts
dbanck: > TS strict does not like the any usage for prototype property discovery

I don't understand what you are referring to, can you explain?

---

Thanks for the hints about `Object.assign` I'm fine with avoiding it. What do you think of using `{ ...process.env }` for creating a shallow copy?

jpogran: There is no performance difference between using the `JSON` method and `{ ... process.env }` so I changed it to `{ ... proces.env } as the more important bit here is trying to solve the 'can't find terraform' part and not how we access environment variabls

src/clientHandler.ts
jpogran: I can't speak for what is done in stdlib. I'm just trying to work around documented problems in accessing the value for PATH

jpogran:
Closing for now until new reports come in for macOS users

PR #886, opened 2022-01-04
Test VS Code Version Matrix

jpogran:
Change default `npm test` behavior to use the version of VS Code specified in the `VSCODE_VER` environment variable, instead of testing hardcoded versions inside `runtest.ts`.

`VSCODE_VER` supports latest, insiders and version numbers lik 1.55.0.

If the `VSCODE_VER` environment variable is not specified, then the latest stable version of VS Code is automatically used.

### DIFF
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
-import { runTests } from 'vscode-test';
-import { TestOptions } from 'vscode-test/out/runTest';
+import { runTests } from '@vscode/test-electron';
+import { TestOptions } from '@vscode/test-electron/out/runTest';
 import { exec } from '../utils';
 
 async function terraformInit() {
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -32,6 +32,7 @@ async function main(): Promise<void> {
 
   // common options for all runners
   const options: TestOptions = {
+    version: process.env['VSCODE_VERSION'] ?? 'stable',
     extensionDevelopmentPath,
     extensionTestsPath,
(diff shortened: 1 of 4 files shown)


jpogran:
Good news, we now have separate runs for each platform and each vs code version we currently test against. 

Bad news, have uncovered an actual problem with vs code 1.55.0

jpogran:
1.58.2 works

jpogran:
VS Code 1.55 was released in March 2021. 1.58 was release in June 2021. A small subset of our users are still on 1.55 or lower.

The test that is erroring is an autocompletion expectation. The rest of the extension tests out ok.

As a result of all of the above, I pinned to 1.58.2 as there are no user reports from 1.55 users.

src/test/runTest.ts
```diff
@@ -41,15 +41,24 @@ async function main(): Promise<void> {
...
     // start in the fixtures folder to prevent the language server from walking all the
     // project root folders, like node_modules
-    console.log('_______________LATEST_____________________');
-    await runTests(options);
-
-    console.log('_______________INSIDERS___________________');
-    options.version = 'insiders';
-    await runTests(options);
+    const vscodeVersion = process.env['VSCODE_VERSION'];
+    switch (vscodeVersion) {
```
dbanck, line 45: What do you think of removing this `switch` statement entirely? With the new matrix, it's clear which version is being tested. We could then add `version` to the `TestOptions` object in line 34.

If we want to keep some output, we could do something like this:
```ts
console.log(`_______________${options.version ?? 'latest'}_____________________`);
```

dbanck (approved):
LGTM! Just added a suggestion to get rid of some code, but I'm also fine if you want to keep it.

dbanck (approved):
Nice work!

PR #890, opened 2022-01-05
Document automatic format

jpogran:
This adds documentation to the README on how to configure VS Code to automatically format Terraform files.


README.md
```diff
@@ -59,20 +59,34 @@ If you want to automatically search root modules in your workspace and exclude s
...
+}
 ```
 
-It is also recommended to set a default formatter if you have other extensions installed which also claim a capability to format Terraform files:
+It is recommended to set `editor.defaultFormatter` to ensure that VS Code knows which extension to use to format your files. It is possible to have more than one extension installed which claim a capability to format Terraform files.
+
+When using the `editor.formatOnSaveMode` setting, only `file` is currently supported. The `modifications` or `modificationsIfAvailable` settings [use the currently configured SCM](https://code.visualstudio.com/updates/v1_49#_only-format-modified-text) to detect file line ranges that have changed and send those ranges to the formatter. The `file` setting works because `terraform fmt` was written to format an entire file, not ranges. If you don't have a SCM enabled for the files you are editing, `modifications` won't work at all. The `modificationsIfAvailable` setting will fall back to `file` if there is no SCM and will appear to work sometimes.
+
+If you want to use `editor.codeActionsOnSave` with `editor.FormatOnSave` to automatically format Terraform files, use the following configuration:
```
dbanck, line 73:
```suggestion
If you want to use `editor.codeActionsOnSave` with `editor.formatOnSave` to automatically format Terraform files, use the following configuration:
```

README.md
```diff
@@ -59,20 +59,34 @@ If you want to automatically search root modules in your workspace and exclude s
...
+  "editor.formatOnSave": false,
+  "editor.codeActionsOnSave": {
+    "source.formatAll.terraform": true
+  },
+},
 ```
 
+This will keep the global `editor.formatOnSave` for other languages you use, and configure the Terraform extension to only format during a `codeAction` sweep.
+
+> **Note**: Ensure that the terraform binary is present in the environment PATH variable. If the terraform binary cannot be found, formatting will silently fail.
```
dbanck, line 88:
```suggestion
> **Note**: Ensure that the terraform binary is present in the environment `PATH` variable. If the terraform binary cannot be found, formatting will silently fail.
```

dbanck (approved):
Nice work! Just two small suggestions

README.md
```diff
@@ -59,20 +59,34 @@ If you want to automatically search root modules in your workspace and exclude s
...
+  "editor.defaultFormatter": "hashicorp.terraform",
+  "editor.formatOnSave": true,
+  "editor.formatOnSaveMode": "file"
+}
 ```
 
-It is also recommended to set a default formatter if you have other extensions installed which also claim a capability to format Terraform files:
+It is recommended to set `editor.defaultFormatter` to ensure that VS Code knows which extension to use to format your files. It is possible to have more than one extension installed which claim a capability to format Terraform files.
+
+When using the `editor.formatOnSaveMode` setting, only `file` is currently supported. The `modifications` or `modificationsIfAvailable` settings [use the currently configured SCM](https://code.visualstudio.com/updates/v1_49#_only-format-modified-text) to detect file line ranges that have changed and send those ranges to the formatter. The `file` setting works because `terraform fmt` was written to format an entire file, not ranges. If you don't have a SCM enabled for the files you are editing, `modifications` won't work at all. The `modificationsIfAvailable` setting will fall back to `file` if there is no SCM and will appear to work sometimes.
```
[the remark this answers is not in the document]
radeksimko: > The `file` setting works because `terraform fmt` was written to format an entire file, not ranges.

You're not wrong here in the sense that it was _designed for_ whole files but there's no strong reason why we can't implement [`textDocument/rangeFormatting`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_rangeFormatting) and pass just part of the config to stdin to `terraform fmt -`.

I guess this will expose some edge cases, like formatting range which slices through an attribute or block - and I'm unsure how the underlying implementation deals with that or how it should.

Anyway - I'm happy with your conclusion here, just wanted to say that we can address this on LS side if there's interest.

README.md
```diff
@@ -59,20 +59,34 @@ If you want to automatically search root modules in your workspace and exclude s
...
+
+If you want to use `editor.codeActionsOnSave` with `editor.formatOnSave` to automatically format Terraform files, use the following configuration:
 
 ```
-    "[terraform]": {
-        "editor.defaultFormatter": "hashicorp.terraform",
-        "editor.formatOnSave": true
-    }
+"editor.formatOnSave": true,
+"[terraform]": {
```
radeksimko, line 77: Does this represent language ID? If so, then do we need a similar block for `[terraform-vars]`?

radeksimko (approved):

README.md
jpogran: Yes, that's the language ID

image

Does terraform format tfvars files?

README.md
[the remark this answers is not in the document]
jpogran: Thanks for the added context. The purpose of this text is to instruct the user how to use it as currently implemented so I did not include the possibility of passing ranges of text to terraform to format. Since we both thought of this, this sounds like I should create a ticket for this idea.

README.md
radeksimko: It does.

```
Usage: terraform [global options] fmt [options] [DIR]

	Rewrites all Terraform configuration files to a canonical format. Both
	configuration files (.tf) and variables files (.tfvars) are updated.
	JSON files (.tf.json or .tfvars.json) are not modified.
```
> Both configuration files (.tf) and **variables files (.tfvars)** are updated.

radeksimko: It's not a big deal I suppose. If we want it to be complete then I suppose we should include it.

Feel free to merge as is though.

jpogran: Might as well since we're here! Added

PR #891, opened 2022-01-05
Exclude terraform cache from file watcher

### LINKED ISSUE #405: Exclude files in .terraform from watching
I'm not sure which part of the extension exactly decides which files to watch and send as part of `didChangeWatchedFiles`, but I noticed that it's attempting to report files in the Terraform cache to the server:
https://gist.github.com/radeksimko/12e7d68005a95b87c2b4fd813354f888

I believe that `.terraform` should be excluded from these events (from being watched), because:

 - Editing cached modules is not a practice we should be encouraging, so the server does not intend to support _editing_ these files
 - Even if there is some server logic which relies on `.terraform` contents, the server will likely always prefer watching and interpreting it directly from the disk, because it cannot rely on every server to watch hidden directories for essential functions. (some LSP plugins leave this as user-configurable, which leaves a lot of room for error)
 - It can generate a _lot_ of requests when `terraform init` runs

This is somewhat related to https://github.com/hashicorp/terraform-ls/issues/137
The server doesn't support `workspace/didChangeWatchedFiles` yet and so it can generate more errors than expected as a result of the above.

--- 

We may consider offloading watching to the client by dynamic registration of paths per https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeWatchedFiles but I think that `.terraform` should be excluded _by default_.

--- 

Related: https://github.com/hashicorp/terraform-ls/issues/867

jpogran:
This adds the `.terraform` cache directory to the file watcher exclude list by default. A user can override this by configuring `files.watcherExclude` in their user or workspace settings file.

Closes #405

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,11 @@
         "path": "./snippets/terraform.json"
       }
     ],
+    "configurationDefaults": {
+      "files.watcherExclude": {
+        "**/.terraform/*/**": true
+      }
+    },
     "configuration": {
       "type": "object",
       "title": "Terraform",


dbanck:
Do you know if `configurationDefaults` can override any default setting? The extension documentation isn't really clear on this: https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults

And will it override the default value of the object? https://code.visualstudio.com/docs/getstarted/settings#_default-settings

```
  // Configure paths or glob patterns to exclude from file watching.
  "files.watcherExclude": {
    "**/.git/objects/**": true,
    "**/.git/subtree-cache/**": true,
    "**/node_modules/*/**": true,
    "**/.hg/store/**": true
  },
```

jpogran:
If user specifies, then the default is overridden. After testing this with no user config, user provided config, workspace provided config, etc, there are too many variables at this moment to document. I am going to close this in favor of stepping back and researching this ticket better.

PR #895, opened 2022-01-07
Ignore jest config when packaging

jpogran:
This ignores the jest config file when packaging the extension.


dbanck:
There is a stray `foo/` directory in this PR

jpogran:
Oops

dbanck (approved):

Issue #849, opened 2021-11-11
Set ESLint and Typescript project rules


jpogran:
I think we're not getting much value at this point from forcing "no explicit any" via eslint. So I would rather suggest disabling both rules globally instead of using inline comments.

```json
  "rules": {
    "@typescript-eslint/no-explicit-any": 0,
    "@typescript-eslint/explicit-module-boundary-types": 0
  }
```

We may want to look at our tsconfig next, add more restrictive compilerOptions, and then re-evaluate the eslint rules.

# Proposal

* Add `"strict": true` (https://www.typescriptlang.org/tsconfig#strict) to our `tsconfig.json` and fix all compiler errors
* Re-evaluate eslint rules to match strict flag

_Originally posted by @dbanck in https://github.com/hashicorp/vscode-terraform/pull/848#pullrequestreview-803636835_

PR #896, opened 2022-01-07
Update eslint and minimal ruleset

jpogran:
This updates eslint and related plugins to latest package versions. It also brings the eslint config into line with default config published by the VS Code Extension yo template.

Most of the code is compliant with the eslint rulsets, with the exception of naming convention for fields and variables. Where we could update the naming, we did. Where we could not eslint disable lines were added.


jpogran:
This partially address #849 in that it adds some default ruleset. It does not go as far as we want to yet.

dbanck (approved):
Thank you for updating the packages and adding more linting rules.

`@typescript-eslint/naming-convention` got me thinking. I don't particularly appreciate having `eslint-disable` comments in the code and wondered why we need that much. The LSP spec is using camelcase for all attributes on the RPC calls, but we missed doing this for our custom commands :/
For future commands, we might want to use camelcase for the JSON attributes in the go code and revisit our existing commands at some point.

jpogran:
Yep. We missed the chance before, and changing now would require a tandem client/ls release. Which is not easy now but will be when platform specific extension lands.

PR #897, opened 2022-01-10
Update actions/setup-node to v2

dbanck:
I’ve seen multiple canceled test runs on macos-latest in the setup-node step over the last couple of days. Upgrading to `actions/setup-node@v2` should make things more reliable.

Additionally, https://github.com/actions/setup-node/pull/338 added support for fetching the node version from a `.npmrc` file, so I was able to simplify most workflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,21 +12,13 @@ jobs:
       - name: checkout repo
         uses: actions/checkout@v2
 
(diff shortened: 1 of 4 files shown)


radeksimko (approved):
Nice cleanup! 👍🏻

jpogran (approved):
😍

PR #905, opened 2022-01-14
Publish Platform Specific Extension

jpogran:
This converts the extension to use a bundled terraform-ls binary and packages the VSIX as a Platform Specific Extension. A matrix of all supported platforms and architectures run the commands necessary to bundle terraform-ls and the extension into a VSIX. After packaging, it publishes all stored VSIXs to the Marketplace. This results in a VSIX that is ready to run at install time, instead of having to use the network at runtime to download a terraform-ls binary.

This also adds minification as a final step to the interactive development and packaging processes. This now produces a single file extension, compared to previously shipping several thousand files. This results in a much faster installation experience and reduces the deployment size of the extension.


build/downloader.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface ExtensionInfo {
+  extensionVersion: string;
+  languageServerVersion: string;
+}
+
+export function getExtensionInfo(): ExtensionInfo {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const pjson = require('../package.json');
+  // return `${pjson.publisher}.${pjson.name}`;
```
dbanck, line 35: Can be removed?

src/extension.ts
```diff
@@ -64,7 +51,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
           vscode.ConfigurationTarget.Global,
         );
       }
-      await updateLanguageServer(manifest.version, lsPath);
+      await clientHandler.startClient();
       return clientHandler.startClient();
```
dbanck, line 55: When starting the client via the 'Enable Language Server' command here, we aren't catching any errors and are not running `vscode.commands.executeCommand('setContext', 'terraform.showTreeViews', true);` afterwards.

This is resulting in two different error messages:
_here_
CleanShot 2022-01-17 at 14 35 50@2x

_line 135_
CleanShot 2022-01-17 at 14 36 11@2x

As you just added the error dialog in line 135, do you think we should do this here, too? Or should I create an issue for that?

dbanck (commented):
Awesome work! I'm amazed that esbuild doesn't need a bulky config file. It looks so lean with your `package.json` scripts.

I haven't looked a the GitHub workflows yet; as you mentioned, you want to test them first.

I wondered if we could avoid downloading the language server every time we run 'Launch client' in the local development workflow?

src/extension.ts
jpogran: I extracted two helper methods to wrap calling start/stop with error handling to improve the enable/disable commands. I don't like them being there in the first place, but that is another set of work entirely.

Speaking of new issues, we need to add error handling with discovering the binary to avoid errors like you showed. This generic error handler won't precisely tell the user what to do and we can do better than that.

jpogran:
> I wondered if we could avoid downloading the language server every time we run 'Launch client' in the local development workflow?

It won't download every time for local development because it uses the `npm watch` task. On the second run, it's already running so it won't rerun `npm watch`. If the terminal that's running watch is terminated, then it will redownload. I'm fine with having it exit early if the folder is present.

package.json
```diff
@@ -320,53 +323,53 @@
     ]
   },
   "scripts": {
-    "vscode:prepublish": "npm run compile",
-    "compile": "tsc -p ./",
-    "watch": "tsc -watch -p ./",
-    "lint": "eslint src --ext ts",
-    "pretest": "npm run compile && npm run lint",
+    "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild": "npm run esbuild-base -- --sourcemap",
```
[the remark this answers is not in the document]
dbanck: I assume you're referring to [this](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#run-esbuild) docs? From my understanding they are not running `esbuild` when publishing, but `esbuild-base -- --minify` instead.

I think we need source maps for development and can skip shipping them.

.github/workflows/preview.yml
```diff
@@ -3,43 +3,91 @@ name: Publish Preview release
...
-  publish:
-    name: Publish Preview to VS Code Marketplace
-    runs-on: ubuntu-latest
+  build:
+    name: Package
+    strategy:
+      matrix:
+        include:
+          - os: windows-latest
+            platform: win32
```
dbanck, line 15: What are `platform` and `arch` doing? I didn't see any references to them, and they shouldn't affect the [runner platform](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on) as `runs-on` only uses `target`.

.github/workflows/preview.yml
```diff
@@ -3,43 +3,91 @@ name: Publish Preview release
 on:
   schedule:
     # run every day at 00:15 UTC to avoid high load times at the start of every hour
-    - cron: '15 0 * * *'
+    - cron: "15 0 * * *"
```
dbanck, line 6: Do we want to use single or double quotes in yaml files? Prettier is formatting the file with single quotes for me.

DEVELOPMENT.md
```diff
@@ -1,32 +1,108 @@
...
+
+>It is *required* that `terraform` is available on `$PATH` to run the tests.
+
+To run the `unit tests` from the command-line run:
+
+```bash
+> `npm test:unit`
+```
+
+To run the `integration tests` from the command-line without any configuration, close any open VSCode windows and run `npm test`. By default, `npm test` will test against VS Code Stable. If you want to test against a different VS Code version, or if you want to have VS Code remain open, use an environment variable to indicate which version of VS Code to test against:
```
dbanck, line 50: With the updated version of `@vscode/test-electron` VSCode can remain open, even if you're running tests against stable 🎉

.github/workflows/preview.yml
```diff
@@ -3,43 +3,91 @@ name: Publish Preview release
...
-
+          node-version-file: ".nvmrc"
       - name: Install dependencies
         run: npm ci
-
-      - name: vsce package
-        run: npx -- vsce package --out=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
-
+        env:
+          npm_config_arch: ${{ matrix.npm_config_arch }}
```
dbanck, line 64: What is `npm_config_arch` doing? I wasn't able to find anything in the `npm ci` docs

build/downloader.ts
```diff
@@ -0,0 +1,110 @@
...
+// ls_target=linux-x64 npm run package -- --target=linux-x64
+const lsTarget = process.env.target;
+if (lsTarget !== undefined) {
+  const tgt = lsTarget.split('-');
+  os = tgt[0];
+  arch = tgt[1];
+}
+
+// npm run download:ls --target=darwin-x64
+const target = process.env.npm_config_target;
```
dbanck, line 103: I'm unsure what this is doing. How is `npm_config_target` passed?

build/preview.ts
```diff
@@ -0,0 +1,40 @@
...
+async function run() {
+  const cwd = path.resolve(__dirname);
+
+  const buildDir = path.basename(cwd);
+  const repoDir = cwd.replace(buildDir, '');
+
+  // Patch README
+  const data = fs.readFileSync(path.join(repoDir, 'README.md'), { encoding: 'utf8' });
+  const insert = fs.readFileSync(path.join(buildDir, 'nightly.md'), { encoding: 'utf8' });
+  fs.writeFileSync(path.join(repoDir, 'README.md'), `${insert}\n${data}`);
```
dbanck, line 13: This won't replace the `# Terraform Visual Studio Code Extension` headline in the `README.md` anymore, but I think that's fine and looks better now.

There is an old typo in `nightly.md` : `Preivew`. Do you mind fixing it?

build/preview.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const version = `${String(date.getFullYear())}.${date.getMonth() + 1}.${date.getDate()}${String(
+    date.getHours(),
+  ).padStart(2, '0')}`;
+
+  json = JSON.stringify(
+    {
+      ...json,
+      preview: true,
+      // name: `${json.name}-preview`,
```
dbanck, line 28: I think we need the `-preview` suffix to the name?

build/preview.ts
```diff
@@ -0,0 +1,40 @@
...
+  const data = fs.readFileSync(path.join(repoDir, 'README.md'), { encoding: 'utf8' });
+  const insert = fs.readFileSync(path.join(buildDir, 'nightly.md'), { encoding: 'utf8' });
+  fs.writeFileSync(path.join(repoDir, 'README.md'), `${insert}\n${data}`);
+
+  // Patch package.json
+  const date = new Date(new Date().toLocaleString('en-US', { timeZone: 'America/New_York' }));
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  let json = require('../package.json');
+
+  const version = `${String(date.getFullYear())}.${date.getMonth() + 1}.${date.getDate()}${String(
```
dbanck, line 20: If we are using a date here instead of the date of the last git commit, we'll publish a new version every night, even if nothing was changed. Is this intended?

dbanck (commented):
Thanks for taking the time to update the development docs; this will be very helpful in the future!

Most of the code looks good to me! I've added mainly question to gain a better understanding.

.github/workflows/preview.yml
```diff
@@ -3,43 +3,91 @@ name: Publish Preview release
...
+        run: npm run preview
+      - name: Package VSIX
+        run: npm run package -- --target=${{ matrix.target }}
+        env:
+          target: ${{ matrix.target }}
+      - name: Upload vsix as artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.target }}
+          path: "*.vsix"
```
radeksimko, line 76: Is there a particular reason for adding the vsix, when it can be downloaded from the Marketplace?

.github/workflows/preview.yml
```diff
@@ -3,43 +3,91 @@ name: Publish Preview release
...
           export EXTENSION_ID=`cat package.json | jq -r '.publisher + "." + .name'`
           npx -- vsce show --json $EXTENSION_ID | jq '.versions[0]'
 
-      - name: vsce publish
-        run: npx -- vsce publish --no-git-tag-version --packagePath=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
+  publish:
+    name: Publish All
+    runs-on: ubuntu-latest
+    needs: build
+    if: success() && startsWith( github.ref, 'refs/tags/')
```
radeksimko, line 87:
```suggestion
    if: success() && startsWith( github.ref, 'refs/tags/v')
```
Totally nicpicky suggestion - I guess it won't make a difference but I'm thinking it would more accurately reflect intentions - of publishing on _version_ tags.

.github/workflows/publish.yml
```diff
@@ -6,14 +6,58 @@ on:
...
+          - os: windows-latest
+            platform: win32
+            arch: ia32
+            npm_config_arch: ia32
+            target: win32-ia32
+          # - os: windows-latest
+          #   platform: win32
+          #   arch: arm64
+          #   npm_config_arch: arm
+          #   target: win32-arm64
```
radeksimko, line 28: I remember you mentioning some issues with windows/arm64, what's the latest on this topic?

.github/workflows/publish.yml
```diff
@@ -29,18 +73,26 @@ jobs:
...
-      - name: vsce publish
-        run: npx -- vsce publish --no-git-tag-version --packagePath=${{ runner.temp }}/terraform-${{ steps.ext-version.outputs.content }}.vsix
+          name: ${{ matrix.target }}
+          path: "*.vsix"
+
+  publish:
+    name: Publish All
+    runs-on: ubuntu-latest
+    needs: build
+    if: success() && startsWith( github.ref, 'refs/tags/')
```
radeksimko, line 92:
```suggestion
    if: success() && startsWith( github.ref, 'refs/tags/v')
```

DEVELOPMENT.md
```diff
@@ -1,32 +1,108 @@
...
-Requirements:
+## Building
 
-- Node
-- npm
-- VSCode
+The extension makes use of the [VSCode Language Server](https://github.com/Microsoft/vscode-languageserver-node) client package to integrate with [terraform-ls](https://github.com/hashicorp/terraform-ls) for [language features](https://code.visualstudio.com/api/language-extensions/programmatic-language-features). The directions below cover how to build and package the extension; please see the [`terraform-ls`](https://github.com/hashicorp/terraform-ls) documentation for how to build the language server executable.
 
-After cloning the repo, run `npm i` to install dependencies. There's an included build task to compile the TypeScript files to JavaScript; you can run it directly with `npm run compile`.
+### Requirements:
```
radeksimko, line 9:
```suggestion
### Requirements
```

DEVELOPMENT.md
```diff
@@ -1,32 +1,108 @@
...
-The extension can be run in a development mode (useful for testing syntax highlighting changes, for example) via the debug task called `Launch Client`. This will open a new window with the extension loaded, and from there you can open any files or folders you want to check against. This extension development window can also be used to run commands, and use the language server installer.
+### Getting the code
 
-To package your local development copy for testing, use the [vsce](https://www.npmjs.com/package/vsce) tool with `vcse package`.
+```
+git clone https://github.com/hashicorp/vscode-terraform
+```
 
-# Development
+### Dependencies:
```
radeksimko, line 21:
```suggestion
### Dependencies
```

DEVELOPMENT.md
```diff
@@ -1,32 +1,108 @@
...
+
+The extension can be run in a development mode via the launch task called `Launch Client`. This will open a new VS Code window with the extension loaded, and from there you can open any files or folders you want to check against. This extension development window can also be used to run commands or any other feature the extension provides.
+
+> New to VS Code development? You can get started [here](https://code.visualstudio.com/api/get-started/your-first-extension)
+
+## Tests
+
+Automated `unit` and `integration` tests can be written using [mocha](https://mochajs.org) and live inside `./src/test` with file pattern `*.test.ts`.
+
+>It is *required* that `terraform` is available on `$PATH` to run the tests.
```
radeksimko, line 42:
```suggestion
> It is *required* that `terraform` is available on `$PATH` to run the tests.
```

build/downloader.ts
```diff
@@ -0,0 +1,110 @@
...
+    console.log('terraform-ls path exists. Exiting');
+    return;
+  }
+
+  fs.mkdirSync(installPath);
+
+  const extInfo = getExtensionInfo();
+  console.log(extInfo);
+
+  const userAgent = `Terraform-VSCode/${extInfo.extensionVersion}`;
```
radeksimko, line 68: I see the new User-Agent omits VS Code version, I'm guessing this was intentional? I suppose that the download stats won't really be as useful after this goes live anyway - i.e. I suppose we don't care as much about downloads from CI 😄  - so I have no strong opinions here - just checking your intentions.

radeksimko (approved):
Thank you for all the work, and esp. for the updated docs!

.github/workflows/preview.yml
jpogran: Keeping them is not for providing downloadable packages, its for ensuring a build is published only if all platforms succeed.

We have a matrix of platform/arch that build concurrently, using the default GA behavior of fail all if one fails. If we did not separate the publish and package steps, one matrix set could succeed and another fails, meaning that we publish for one platform/arch and not others.

build/downloader.ts
jpogran: It was intentional in the sense that at the time I was more interested in getting the script working than reporting which vscode version we're using. Since this is now outside of vscode, there isn't an easy variable to pull to get the vscode version, I would have to write something extra to find that out.

As you point out, this is less relevant as we're building in CI and that there isn't a 'VS Code' there to track, however it might be a good idea to put something there so we can filter CI runs from the data so that we can see those coming from other editors. Ideas for a name/version to use here?

.github/workflows/preview.yml
jpogran: I use the YAML extension, not prettier, to format yaml and that uses double quote by default. If we already have an existing formatting convention for single quotes I can change the settings to use single instead.

build/preview.ts
jpogran: During publish testing I couldn't have that there. Nice catch!

.github/workflows/preview.yml
jpogran: This approach was taken from the [platform specific extensions CI example](https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml#L61). Things like `platform` and `arch` were supposed to be combined inside the shell commands to make environment variables like `"target=${{ matrix.platform }}-${{ matrix.arch }}`.

However, if I did that I would have to write out different environment variable statements for different shells, since each OS has a different way to do that. If I used powershell, it could be one command for all, but we don't do that here.

I chose instead to be more verbose in the matrix and add `target` which has the correctly combination of platform and arch. We technically can remove `platform` and `arch`, since they're not used at the moment.

.github/workflows/preview.yml
jpogran: The [`npm_config`](https://docs.npmjs.com/cli/v7/using-npm/config) part is a special set of environment variables that NPM automatically populates for each `npm run` command, with whatever after the last `_` being accessible as an environment variable.

While I don't think we'll need it for our stuff, I saw this as setting the `arch` used to compile npm dependencies that had platform specific code inside it. So, not our package per se, but packages we depend on that would need to know we're packaging an `arm` build vs ` x64` build.

.github/workflows/publish.yml
jpogran: I did not attempt to run this again on the test repo. Was too busy trying to get the whole thing working. Will circle back on this

build/downloader.ts
jpogran: The [`npm_config`](https://docs.npmjs.com/cli/v7/using-npm/config) system is hard to explain in github comments. I think it's better if we do a screen share on it.

The intention here was to eventually get to the point where you could on Linux run `npm package -- --target=win32-x64` and get a Windows build, or vice versa. The actual command name doesn't matter, I wanted an ability to run some kind of command to get a 'test' VSIX for a given platform/architecture combo no matter which OS you were sitting on currently. This would allow any of us to test locally whaterver combo we needed, as well as run the build in CI on one platform but package for all.

This gets complicated using the `npm run` script commands, as parameters don't flow down to second level `npm run` script commands. So if I call `npm run foo --bar=baz`, which calls `npm run wakka`, the script `wakka` doesn't see `bar` as either a parameter or an environment variable. As shown in the `downloader.ts` file, I eventually got it to the point where if you used the right combo of environment variables and parameters you could get a correct build, but the UX didn't feel right and felt kludgy and there was no way to 'tell' the user how to do this without reading docs or the source code.

The obvious solution would be to start writing some kind of shell script or use a build task runner framework but that quickly goes beyond scope of this ticket. I ultimately gave up trying to get this to work and put it off for a second pass later.

build/preview.ts
jpogran: Nice catch, didn't notice the line not being replaced in my frustration at `git log` date formatting not working across platforms.

build/preview.ts
jpogran: Good point. So, we should we prevent the run from executing in Github instead of relying on git log behavior to prevent this.

build/downloader.ts
jpogran: Updated with modified user agent

.github/workflows/publish.yml
jpogran: With the new matrix approach, this builds an arm distributable

build/preview.ts
jpogran: With the new matrix approach, we are able to still use the bash scripts

build/preview.ts
jpogran: With the new matrix approach we no longer need this file

build/preview.ts
jpogran: With the new matrix approach, we no longer need this file

build/downloader.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  fs.mkdirSync(installPath);
+
+  const extInfo = getExtensionInfo();
+  console.log(extInfo);
+
+  // userAgent = `Terraform-VSCode/${extensionVersion} VSCode/${vscodeVersion}`;
+  const ciBuild = process.env.CI;
+  const runnerLocation = ciBuild ? `CLI-Downloader GitHub-Actions` : `CLI-Downloader`;
+  const userAgent = `Terraform-VSCode/${extInfo.extensionVersion} ${runnerLocation}/(${platform}; ${architecture})`;
```
radeksimko, line 71:
```suggestion
  const userAgent = `Terraform-VSCode/${extInfo.extensionVersion} ${runnerLocation} (${platform}; ${architecture})`;
```
Version would typically follow `/` in User-Agent, and metadata is more often separated by space in braces.

radeksimko (approved):

dbanck (approved):

PR #909, opened 2022-01-26
Standardize Changelog

jpogran:
The current changelog contains a list of all versions back to the inception of the extension and is an amalgam of different processes for classifying and sorting different issues or pull requests added to the log.

This PR moves all releases 1.4.0 and older to a new Previous Releases section, to reduce the length of the log while maintaining references to those versions. For each release header and classification, this adds dates and normalizes the capitalization, naming conventions, and ordering. Lastly it adds links to the tag comparison for each release, so a view of changes for each release can be seen.

This results in a human readable and accessible format that works both in GitHub and in the VS Code Marketplace.


dbanck (approved):
LGTM! Thanks for going over so many lines of changelog!

dbanck:
You could add a new unreleased section for the now merged PSE :)

jpogran:
Thanks for the reminder about the Unreleased section. I added a link in the bottom to the compare URL, but forgot to add the header in the beginning. Now we can add a date header underneath `# Unreleased` instead of remembering to add it each time

radeksimko (approved):
One nitpick - would you mind also making the headers 2nd level? i.e. replace `^# ` with `## `? I'm guessing that shouldn't upset VS Code Marketplace and aligns us with other HashiCorp changelogs (which we also already follow in LS).

PR #911, opened 2022-01-27
Manual Preview release workflow dispatch

jpogran:
This adds a workflow_dispatch event to the preview workflow so that we can publish the extension manually.


dbanck (approved):
⌛

PR #914, opened 2022-02-01
Add Terraform language tmgrammar snapshosts

jpogran:
This adds a snapshot file for every language construct from https://www.terraform.io/language that makes sense to test syntax grammar for. Some constructs are left out because it does not make sense to test grammar for it now.

A minimal listing of numeric and string functions are added to expressions_functions.tf, which is not the complete listing of functions. The existing grammar uses a regex with a list of function names, so this small list is enough to test at the moment. If we decide to change the regex for functions in the future, we should add more test cases to this file.


tests/snapshot/terraform/expressions_operators.tf
```diff
@@ -0,0 +1,25 @@
...
+5 % 6
+
+-thing
+
+# equality
+
+thing1 == thing2
+thing1 != thing2
+
+# comaprison
```
radeksimko, line 14:
```suggestion
# comparison
```

radeksimko (approved):
I'm assuming you know this and it may not matter for the static grammar - that some lines represent an expression which is only valid either within a template (i.e. "${ HERE }") or on the RHS from attribute name.

jpogran:
Correct, however we can't always check for things like that with the grammar how it is. For example, if we are inside a template or if we are in a tfvars file.

There is some refactoring we can do to separate out those aspects, but these snapshots are an attempt to get a baseline of what we have now and what it lights up.

dbanck:
Sorry for the delayed review.

I've noticed that only the first snapshot test is being executed.

CleanShot 2022-02-03 at 11 42 15

I've raised https://github.com/hashicorp/vscode-terraform/pull/917 to address this.

jpogran:
Apparently that's a platform specific glob problem, as it worked locally 😂

PR #917, opened 2022-02-03
Run syntax tests on all snapshot files

dbanck:
This PR modifies the command that runs the tmgrammar tests to pick up all snapshot tests instead of just the first one.


package.json
```diff
@@ -336,8 +336,8 @@
     "test": "node ./out/test/runTest.js",
     "test:unit": "jest",
     "test:syntax": "npm run test:grammar && npm run test:snap",
-    "test:grammar": "npx vscode-tmgrammar-test -s scope.terraform -g syntaxes/terraform.tmGrammar.json -t tests/unit/**/*.tf",
-    "test:snap": "npx vscode-tmgrammar-snap -s scope.terraform -g syntaxes/terraform.tmGrammar.json -t tests/snapshot/**/*.tf",
+    "test:grammar": "npx vscode-tmgrammar-test -s scope.terraform -g syntaxes/terraform.tmGrammar.json -t 'tests/unit/**/*.tf'",
```
jpogran, line 339: Since we're here, would it be better to use `\"` instead as we'll have to eventually do `{tf|tfvars}`, etc? Or wait until the need arises?

dbanck: I think there is no need to use `\"` here. Multiple file extensions should work with single quotes, too: `'tests/snapshot/**/*.+(tf|tfvars)'`.

jpogran (approved):

PR #922, opened 2022-02-04
Run syntax tests when grammar changes

jpogran:
This runs the syntax tests only if the grammar files change in the syntaxes directory.

As we start to modify the syntax and grammar more, we'll be running tests more frequently. We don't need to incur the cost or time of running the entire suite when we are testing grammar changes.

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -6,11 +6,13 @@ on:
       - main
     paths:
       - 'syntaxes/**'
+      - 'tests/**'
   pull_request:
     branches:
       - main
     paths:
       - 'syntaxes/**'
+      - 'tests/**'
 
 jobs:
   syntax:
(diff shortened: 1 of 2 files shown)


.github/workflows/syntax.yml
```diff
@@ -0,0 +1,28 @@
+name: Run syntax tests
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'syntaxes/**'
```
dbanck, line 8: We might also want to run this when the syntax test files change: `'tests/**'`

dbanck (changes requested):
To be sure: this will still run the main `test.yml` workflow on each syntax change, as long as we don't exclude the syntax paths in `test.yml`, right?

jpogran:
Good point. This didn't go far enough and prevent `test.yml` from running when there are just syntax changes

dbanck (approved):
LGTM!

Issue #63, opened 2018-03-25
Support #region #endregion for code folding


jcrben:
These are supported in core vscode for some languages (https://stackoverflow.com/a/46597272/4200039 and https://github.com/Microsoft/vscode/issues/12146) - I suggest adding `#region` support. 

Also if the autoformatter did not force a space before the trailing `#endregion` comment I'd prefer that. 

I tried adding something like https://github.com/Microsoft/vscode/blob/6c421d85528d4ddb37dcb167b7a442e348f51378/extensions/javascript/javascript-language-configuration.json to https://github.com/mauve/vscode-terraform/blob/master/syntaxes/terraform.json and https://github.com/mauve/vscode-terraform/blob/master/terraform.configuration.json but it didn't work.

mauve:
just back from vacation will see if I can find some time this or next week

unacceptable:
Mention was accidental.  Please disregard.

PR #924, opened 2022-02-05
Restore region markers support

kinday:
This PR restores changes from 3ae5eea0b2689d2839db75b388d9baadd81d0f55 to detect region markers (`#region` and `#endregion`) used for [folding blocks of code](https://code.visualstudio.com/docs/editor/codebasics#_folding).

## Demo

https://user-images.githubusercontent.com/1751980/153667692-0578a05e-5135-4602-8092-78a99a8d6297.mov

### DIFF
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -71,5 +71,11 @@
 			"'",
 			"'"
 		]
-	]
+	],
+	"folding": {
+		"markers": {
+			"start": "^\\s*#region",
+			"end": "^\\s*#endregion"
+		}
+	}
 }


kinday:
Hey @jpogran, this is not specific to Terraform, it is usually implemented either by editor or by some plugin. I am not sure if there’s any specification, trying to find it out.

What I know so far is that in most cases it contains `#region` and `#endregion` even if `#` is not used to denote a comment in particular language. I will research it a bit more and come back with findings if any.

As for multiline comments, I can exclude them from pattern. My thinking was to support every comment type but if using multiline comments for single-line comments is discouraged, it is fair not to accept them as region markers.

Maybe it is sensible to remove `// #region` and `// #endregion` support too as documentation implies `#` is primary way to write a comment and it would result in most concise marker notation:

> The Terraform language supports three different syntaxes for comments:
>
> - `#` begins a single-line comment, ending at the end of the line.
> - `//` also begins a single-line comment, as an alternative to `#`.

kinday:
<details>
  <summary><code>#region</code>/<code>#endregion</code> markers origin</summary>

I could not find any specification for it, could not even trace which language/version introduced it first, but it could be found in code examples in books from early 2000s.

Seems like it originated in either C, C# or VB and also seems to be Microsoft’s invention which then spread into other languages mostly unaltered (in Java there’s alternative convention to use `<editor-fold>`/`</editor-fold>`).

</details>

I concur that `/* #region */` does not make much sense, region markers should be one-line comments. I would like to know if `// #region` should be supported or the most concise option should be enforced (`#region`/`#endregion`).

It also seems that this was implemented before (3ae5eea0b2689d2839db75b388d9baadd81d0f55) in response to #63 and then marker handling was delegated to `FoldingProvider` (e39d8dbee29562fc09e2c24b1cdeb49d8e7da340) and is supposed to be working. For some reason it does not in my case.

I’ll look into my own setup and either close the pull request or submit a fix if there’s an issue. Till then.

radeksimko:
@kinday Do you mind expanding on what your main motivation is - i.e. what snippets of code would you typically want to fold using this feature?

Relatedly - would a pre-built (albeit more generic, schema-driven) folding support provided by language server help here? See https://github.com/hashicorp/terraform-ls/issues/105 - I think we could implement that without having to commit to a particular comment format (i.e. without having to "create new standard/convention").

kinday:
@radeksimko right now I have pretty small infrastructure that I would keep in one file and group resources in a pretty similar way I’ve done in my video above: domains, DNS records, servers, etc.

Additionally would have Terraform configuration, variables and outputs in their own groups. This would still be a case for projects that are split into multiple files.

If I were to summarise: I’d like to use folding regions as a tool to have overview and navigation in the file(s).

I’m not that familiar with language server concept and I feel like I might misunderstand your question. Do you mean introduction of new _block type_ for grouping (and therefore folding)? Something like:

```hcl
group "group_name" {
  resource "type" "name1" {}
  resource "type" "name2" {}
  resource "type" "name2" {}
}
```

radeksimko:
@kinday I was assuming that with LS we could potentially have foldable regions based on the config (schema) alone, rather than user-defined comments, such that the user could fold _all output blocks_, _all variable blocks_,  _all resources of a particular type `digitalocean_record`_ etc. regardless of where they're in the config and what comments are around it.

Is that something that would help you?

kinday:
Thanks for clarification. I don’t think it will. If we take larger projects with `module` blocks` as example it’s impossible to derive groups of `module` blocks automatically.

Also there’s “Fold all regions...” command in VS Code that affects specifically folding markers. If I run it in a file without any `#region`/`#endregion` blocks nothing gets folded. While it is most likely possible to inject such markers programmatically I see it as a hack as it has clear connection with the `#region`/`#endregion` convention.

kinday:
I’ve restored the version that existed prior to `2.x` release.

jpogran (approved):

jpogran:
I'm going to merge this with just the single line `#region` and `//region`. This enables your workflow but does not change how the Terraform files are parsed and does not affect any other part of the extension, so should be a safe add even if it is not used by the language itself.

Thanks for the contribution @kinday! I appreciate you working through the implementation with us.

kinday:
Fantastic! Thanks for the guidance. 💖

PR #933, opened 2022-02-10
Fix highlighting for attribute access with a dash

### LINKED ISSUE #927: Wrong highlighting for identifiers with dash

dbanck:
This PR fixes the syntax highlighting for attribute access containing a dash (`-`) and the splat and indices detection.

### Variables with dashes
Before
CleanShot 2022-02-10 at 12 06 49

After
CleanShot 2022-02-10 at 12 07 02

### Splat & Indices
Before
CleanShot 2022-02-10 at 12 08 22

After
CleanShot 2022-02-10 at 12 08 02

Fixes #927

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -638,7 +638,7 @@
 					"name": "keyword.operator.accessor.terraform"
 				}
 			},
-			"end": "\\w+(\\[[0-9\\*]+\\])?",
+			"end": "[[:alpha:]][[:alnum:]_-]*(\\[[0-9\\*]+\\])?",
 			"endCaptures": {
 				"0": {
 					"patterns": [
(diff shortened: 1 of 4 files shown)


syntaxes/terraform.tmGrammar.json
```diff
@@ -638,19 +638,15 @@
 					"name": "keyword.operator.accessor.terraform"
 				}
 			},
-			"end": "\\w+(\\[[0-9\\*]+\\])?",
+			"end": "[[:alpha:]][[:alnum:]_-]*(\\[[0-9\\*]+\\])?",
```
jpogran, line 641: Can you add a `"comment"` line here explaining the regex? For future us so when we come back to this we know what this is :)

dbanck: Good idea, I've updated the comments in the attribute access section

jpogran (approved):

PR #934, opened 2022-02-10
Add new scope for block type and name 

### LINKED ISSUE #710: [Feature request] Color syntax: add scope for resource_type and resource_name
Using VSCode Version: 1.58.2, OS: Linux x64 5.8.0-63-generic snap (Ubuntu 20.04.2)

Currently a resource type and resource identification are recognized as string. As such the color syntax is not very readable. Can you please add new editor scopes for:

* Terraform resource_type
* Terraform resource_name

To allow to configure the corresponding element with a different color syntax. The attached screenshot shows the desired effects using IntelliJ IDEA v2021.2

VS Code Color theme:
- [Customizing a Color Theme](https://code.visualstudio.com/docs/getstarted/themes#_customizing-a-color-theme)
- [Theme Color Reference](https://code.visualstudio.com/api/references/theme-color)

VSCode_Terraform_colorSyntax_resource

jpogran:
This changes the existing block matcher to differentiate the type and name of a given block.

Previously the type and name of a block were scoped as `string.quoted.double`. This change scopes them to an `entity.name.tag`, which is commonly used to indicate a tag or attribute.

This will match Terraform blocks like `resource "aws_instance" "web" {` or `module "foo" {`

image

This will match Terraform blocks like `module "foo" {` or `variable`.

image

This will leave alone blocks without types or names:

image

> **Note:** The LS needs to be disabled for you to see this until the semantic tokens are updated in the terraform-ls project.

Fixes #710

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -112,6 +112,7 @@
 		},
 		"block": {
 			"name": "meta.block.terraform",
+			"comment": "This will match Terraform blocks like `resource \"aws_instance\" \"web\" {` or `module {`",
 			"begin": "(\\w+)(?:([\\s\\\"\\-[:word:]]*)(\\{))",
 			"beginCaptures": {
 				"1": {
@@ -130,7 +131,19 @@
 				"2": {
 					"patterns": [
 						{
-							"include": "#string_literals"
+							"name":"entity.name.tag.terraform",
+							"begin": "\"",
+							"beginCaptures": {
+								"0": {
+									"name": "entity.name.tag.begin.terraform"
+								}
+							},
+							"end": "\"",
+							"endCaptures": {
+								"0": {
+									"name": "entity.name.tag.end.terraform"
+								}
+							}
 						}
 					]
 				},
--- a/tests/snapshot/terraform/basic.tf.snap
+++ b/tests/snapshot/terraform/basic.tf.snap
@@ -69,9 +69,9 @@
 >provider "azurerm" {
 #^^^^^^^^ source.terraform meta.block.terraform entity.name.type.terraform
 #        ^ source.terraform meta.block.terraform
-#         ^ source.terraform meta.block.terraform string.quoted.double.terraform punctuation.definition.string.begin.terraform
-#          ^^^^^^^ source.terraform meta.block.terraform string.quoted.double.terraform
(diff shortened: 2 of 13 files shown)


radeksimko (commented):
FWIW Terraform/HCL does support unquoted labels too, e.g. `resource aws_instance test { }`. It's not considered "best practice" (at least you would not find it in any snippets of code on terraform.io) but it is supported and will likely remain supported in the foreseeable future as there are compatibility promises that were already made.

That's not to say the grammar _has_ to account for that - I think it's a relative minority using that syntax, but given this fact I'm wondering if we should be reporting both enclosing quotes as part of the "label", effectively highlighting them the same colour as the characters within it?

radeksimko:
> Do you know if it's allowed to use string interpolation in labels? Do we need to account for this here?

Not that I'm aware of - at least not the interpolation we think of in HCL traditionally - i.e. `hcl.Expression`.

Packer does use labels which seem to refer to other blocks of code (similar to standard references), but under the hood HCL AFAIK still treats this as string and I'm relatively sure they have to deal with that string internally within Packer somehow.
https://www.packer.io/docs/templates/hcl_templates/contextual-variables#source-variables

https://pkg.go.dev/github.com/hashicorp/hcl/v2#Block

```go
type Block struct {
	Type   string
	Labels []string
// ...
}
```

jpogran:
Our existing regex doesn't account for blocks without quotes, the LS saves us by recognizing it:

image

I don't think we should expand scope of this by trying to address legacy behavior. We can ticket an investigation of whether we can have two behaviors inside one regex in another ticket.

I'm partial to the differently highlighted quotes, I think it helps identify different parts easier, but can see where it would conflict with the way interpolated strings look. Let's ask a few others for input and see what feedback we get

jpogran:
Team agreed to keep the string quotes and content the same color/token. PR updated and images adjusted

radeksimko (approved):
LGTM 👍🏻 

_Non-blocking question:_
I understand that not everything in TM may perfectly apply to every language and, but was there a particular reason to choosing `entity.name.tag` over `entity.name.section` or (TM undocumented) `entity.name.label`? They all seem to be somewhat applicable.

https://macromates.com/manual/en/language_grammars
>  - `tag` — a tag name.
>  - `section` — the name is the name of a section/heading.

PR #935, opened 2022-02-11
Improve comment detection

jpogran:
This disambiguates hash line comments, double slash comments, and block comments by adding unique identifiers. This will allow different highlighting for different comment types, as well as help with future code folding and keyboard shortcuts.


dbanck (approved):
Good idea!

PR #939, opened 2022-02-15
Update to @vscode/extension-telemetry

jpogran:
The telemetry package was renamed to @vscode/extension-telemetry


radeksimko (approved):
Looks like this should bring in a number of fixes and improvements 👍🏻 
https://github.com/microsoft/vscode-extension-telemetry/releases

dbanck (approved):

Issue #802, opened 2021-10-05
Error "Unable to launch language server: ENOENT: no such file or directory" with version 2.15.0


cezarsa:
## Versions

This bug is reproducible in:
 - [X] the **latest** version of the extension (below)

### Extension

```
HashiCorp Terraform
v2.15.0
```

### VS Code

```
Version: 1.60.2
Commit: 7f6ab5485bbc008386c4386d08766667e155244e
Date: 2021-09-22T11:59:08.826Z (1 wk ago)
```

### Operating System

```
Mac OS Big Sur 11.6 on ARM M1 macbook
```

### Terraform Version

```
Terraform v1.0.8
```

### Steps To Reproduce
Right after installing the extension the following error message is reported:

```
Unable to launch language server: ENOENT: no such file or directory, access '/Users/cezarsa/.vscode/extensions/hashicorp.terraform-2.15.0/lsp/terraform-ls'undefined
```
<img width="497" alt="Screen Shot 2021-10-05 at 14 44 27" src="https://user-images.githubusercontent.com/11041/136076110-de0878fc-e833-4a4c-93b2-ba79e7e47285.png">

The same error message shows up if I try to run `Terraform: Enable Language Server` command.

### Expected Behavior
The language server executable should be present and the language server should start.

### Actual Behavior
Only the error message above is reported.
 
### Additional context
There's no `lsp` directory in the extension installation path:
```
$ ls /Users/cezarsa/.vscode/extensions/hashicorp.terraform-2.15.0/
CHANGELOG.md                 README.md                    node_modules/                package.json                 syntaxes/                    terraform.png
LICENSE.txt                  language-configuration.json  out/                         snippets/                    terraform-banner.png         testFixture/
```

jpogran:
I'm glad you go it working with a custom install. In https://github.com/hashicorp/vscode-terraform/issues/812 we implemented a different way of installing the terraform-ls binary to avoid file permission issues, which may work for you. We continue to work on bundling the terraform-ls binary inside the extension in https://github.com/hashicorp/vscode-terraform/issues/820. 

https://github.com/hashicorp/vscode-terraform/pull/868 might be of interest to you, as we configured the extension to silently install the terraform-ls binary without prompting the user. This is currently released in the [preview extension](https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform-preview) if you would like to try this out now.

jpogran:
[...]

PR #943, opened 2022-02-23
syntax: Highlight block label as "enumMember" & highlight unquoted labels

radeksimko:
This is to complement https://github.com/hashicorp/terraform-ls/pull/802 by aligning the scope of block labels in the grammar with semantic token type assigned by the language server.

It may seem counter-intuitive or wrong from semantic perspective to use `variable.other.` alongside `entity.name.` (used for known block type) for labels but as explained in the commit log - this is mostly a pragmatic decision made to provide better UX which otherwise would be hard/impossible to achieve due to the scopes used by default VSCode themes.

As mentioned in #802 we explored a number of other token types which VSCode mostly ends up just mapping down to available scope names per https://github.com/microsoft/vscode/blob/de9ab35ce9da7cb3c850a1b570876263a472174b/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L540-L570

So these are the limitations we have to work within. Our future hope is possibly addition of some more token types into LSP as described in https://github.com/microsoft/language-server-protocol/issues/1067

## Before
Screenshot 2022-02-23 at 20 30 18 Screenshot 2022-02-23 at 20 30 35

## After
Screenshot 2022-02-23 at 20 35 34 Screenshot 2022-02-23 at 20 35 21

--- 

## Future Enhancements

@dbanck rightly pointed out that the LS/extension alignment may do a dis-service to the user in case of "invalid" labels. This is tracked under https://github.com/hashicorp/terraform-ls/issues/804

There was also point made in https://github.com/hashicorp/vscode-terraform/issues/710 about distinguishing "dependent labels" from "freeform-text labels". This is tracked under https://github.com/hashicorp/terraform-ls/issues/803

### DIFF
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -131,7 +131,7 @@
 				"2": {
 					"patterns": [
 						{
-							"name": "entity.name.tag.terraform",
+							"name": "variable.other.enummember.terraform",
 							"match": "[\\\"\\-\\w]+"
 						}
 					]
--- a/syntaxes/terraform.tmGrammar.json
+++ b/syntaxes/terraform.tmGrammar.json
@@ -113,12 +113,12 @@
 		"block": {
 			"name": "meta.block.terraform",
 			"comment": "This will match Terraform blocks like `resource \"aws_instance\" \"web\" {` or `module {`",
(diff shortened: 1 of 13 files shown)


dbanck (approved):
LGTM 👍

PR #954, opened 2022-03-01
use terraform ls preview

jpogran:
- Download terraform-ls preview on build
- Set terraform-ls preview version in package.json
- Set preview workflow to use terraform-ls preview


radeksimko (approved):
Code LGTM - testing-wise I guess it's best to test it in a real release.

PR #958, opened 2022-03-03
WIP - custom semantic tokens

radeksimko:
## TODO

 - language server side implementation
 - screenshots demoing a custom theme taking advantage of these token types & modifiers

--- 

This is POC demonstrating how we could support custom token types for highlighting in a way, which:

1. doesn't hamper UX with default themes
2. allows custom themes to be more specific
3. allows us to leverage language server

The assumption here is that `superType` styles are applied if theme doesn't support the custom type identified via `id`. Language server will then be responsible for associating all the custom token types and modifiers to the relevant pieces of code.

Theoretically we could use these in our static grammar too, but I have not explored that possibility too deeply.
 - One clear benefit would be that themes would be applied immediately (before LS starts)
 - The grammar may not be as accurate as semantic highlighting
 - Use cases involving _both_ custom themes _and_ environment without language server are likely to be very few, if any
 - There is maintenance cost associated with this

so given the above I'm leaning more towards keeping this mostly driven by the language server.

The idea is that theme creators could take advantage of these per [available docs](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-coloring-in-color-themes), e.g.

```json
{
  "name": "Red Theme",
  "tokenColors": [
    {
      "scope": "hcl-BlockType",
      "settings": {
        "foreground": "#dd0000",
        "fontStyle": "italic"
      }
    }
  ],
  "semanticHighlighting": true,
  "semanticTokenColors": {
    "hcl-BlockType.terraform-block-type-resource": "#ff0011",
    "hcl-BlockType.terraform-data-type-resource": "#ff0011"
  }
}
```


package.json
```diff
@@ -89,6 +89,58 @@
...
+          "hcl-bool": ["keyword.control"],
+          "hcl-string": ["string"],
+          "hcl-number": ["constant.numeric"],
+          "hcl-objectKey": ["variable.parameter"],
+          "hcl-mapKey": ["variable.parameter"],
+          "hcl-keyword": ["keyword.control"],
+          "hcl-traversalStep": ["variable.other.readwrite"],
+          "hcl-typeCapsule": ["keyword.control"],
+          "hcl-typePrimitive": ["keyword.control"]
+        }
```
radeksimko, line 123: The scopes here just follow VSCode's defaults per https://github.com/microsoft/vscode/blob/02e8bd149ca3fb68e7f26ebec33f920ba0feef96/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L541-L582

radeksimko: As per conversation in https://github.com/microsoft/vscode/issues/145312 we could make these more specific, e.g. by appending `.terraform` to all scopes, but I'm not sure what is the best long-term strategy for token/scope "parity", so I just mostly mimicked the existing behaviours for semantic tokens for now.

src/extension.ts
```diff
@@ -25,6 +25,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 
   const lsPath = new ServerPath(context);
   clientHandler = new ClientHandler(lsPath, outputChannel, reporter);
+  clientHandler.extSemanticTokenTypes = tokenTypesFromExtManifest(context.extension);
```
dbanck, line 28: We could pass `manifest` (which we declare in line 21) to both functions.

dbanck (approved):
LGTM and works as expected; so much room for possibilities now!

Just a minor style suggestion, but feel free to ignore and merge anyway.

src/extension.ts
radeksimko: I did consider that actually but I felt that we'd loose the type-safety for the argument of both functions as we'd have to make it `any` (at least I couldn't find a way of getting the JSON as a special type) - which makes it more error-prone IMO, but I may be overly sensitive to this. Maybe I'm trying to bring too much Go-ism into TypeScript? 🤷🏻  😄

dbanck: I see your point and think vscode doesn't ship types for the `package.json` content.

We _could_ define a custom interface that only contains the fields we access in the function body, but we would still pass `manifest,` which is `any` to the functions.

```ts
interface PartialManifest {
 contributes: {
   semanticTokenTypes?: ObjectWithId[];
   semanticTokenModifiers?: ObjectWithId[];
 };
}

function tokenTypesFromExtManifest(manifest: PartialManifest): string[] {
``` 

I don't have a clear preference here and would love to have types for the `package.json` content. ([some](https://github.com/sindresorhus/type-fest/blob/main/source/package-json.d.ts) are available, but they lack the vscode specific `contributes` object)

radeksimko: Good idea!

PR #961, opened 2022-03-04
github: declare explicit permissions for workflows

radeksimko:
This allows us to be a little bit more defensive by not giving any workflows more permissions than they really need and use this safer default settings for the repo

Screenshot 2022-03-04 at 09 16 36

### DIFF
--- a/.github/workflows/syntax.yml
+++ b/.github/workflows/syntax.yml
@@ -1,5 +1,8 @@
 name: Run syntax tests
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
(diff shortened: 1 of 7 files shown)


.github/workflows/syntax.yml
```diff
@@ -1,5 +1,8 @@
 name: Run syntax tests
 
+permissions:
+  contents: read
```
dbanck, line 4: Why do we need `read` here? For cloning?

dbanck (approved):

.github/workflows/syntax.yml
radeksimko: That is my understanding, but let me try removing it and run it without, to confirm that theory.

radeksimko: Looks like that is not needed, so I'll also remove it from the test workflow and review the same permission in LS.

Good catch - thank you!

dbanck (approved):

PR #968, opened 2022-03-04
github: pin Windows to 2019 to prevent timeouts

radeksimko:
As other people have reported in https://github.com/actions/virtual-environments/issues/4856 there are some issues with windows-2022 (which we consume through `windows-latest`) causing our windows builds to occasionally time out through no fault of our own.

This is in order to address that temporarily but I am hoping we can use 2022 soon, once the bugs are fixed by GitHub.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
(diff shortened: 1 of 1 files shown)


dbanck (approved):
Fingers crossed that this will improve things

Issue #768, opened 2021-09-14
Make Terraform extension available in "browser mode"


radeksimko:
## Problem Statement

GitHub employees and other users use VS Code in the browser for various reasons.
This extension currently isn't available for them.

Example https://github.dev/hashicorp/terraform-guides

<img width="809" alt="Screenshot 2021-09-14 at 10 56 15" src="https://user-images.githubusercontent.com/287584/133237039-37c9068f-a6d0-4755-bd54-76f5c9b8ebc2.png">

## Expected User Experience

Extension is available to assist users in editing Terraform code in their browser, the same way as in desktop VS Code.

For example at https://vscode.dev/ or https://github.dev/

## Proposal

https://code.visualstudio.com/docs/remote/codespaces#_why-is-an-extension-not-installable-in-the-browser

There is some documentation here: https://code.visualstudio.com/api/extension-guides/web-extensions

We could initially provide just a minimal extension for the web - e.g. with syntax highlighting only.

 - [x] Investigate impact on telemetry logic
 - [x] Investigate replacing of esbuild with webpack
 - [x] Create browser entrypoint and link it from `package.json` per docs, e.g. `"browser": "./dist/web/extension.js",`

Longer-term we'd need to find out if there's any way or benefit from running LS in the browser - either via WASM or GopherJS.

andymac4182:
This would be awesome! I have been doing lots of PR reviews in the browser with github.dev

Satak:
Please make this available for github codespaces.

jpogran:
Hey @Satak, apologies for the delay in response. You should be able to use the extension in Codespaces:

![image](https://user-images.githubusercontent.com/272569/157525838-a74ab014-b4f9-4479-9a4d-53a7fc6b1755.png)

Intellisense and other advanced language features work:

![image](https://user-images.githubusercontent.com/272569/157526021-d3008de2-c562-470f-ad87-8a88b3da594b.png)

Satak:
But it is not available for Codespaces in web browser:

![image](https://user-images.githubusercontent.com/10957012/157600413-9e568c7d-0474-47c3-b6b8-a4abdeeef8f5.png)

jpogran:
Yep! I misunderstood what you were referring to because the two are called different things: https://docs.github.com/en/codespaces/the-githubdev-web-based-editor

We currently work in the full Codespaces environment. We don't yet work in the web-based editor.

jpogran:
[...]

PR #989, opened 2022-03-09
Advertise proper exectution location

jpogran:
Specify the workspace execution location so that VS Code prefers to install this extension in the remote host when using Remote Extensions like the Remote Container, SSH, or WSL extensions. VS Code will still install this extension locally when working on the local machine.

For more inforamtion see:

- https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location
- https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location

Install to local:

image

Install to WSL:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,9 @@
   "license": "MPL-2.0",
   "preview": false,
   "private": true,
+  "extensionKind": [
+    "workspace"
+  ],
   "engines": {
     "npm": "~8.X",
     "node": "~16.X",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "preview": false,
   "private": true,
   "extensionKind": [
-    "workspace"
+    "workspace",
+    "ui"
   ],
   "engines": {
     "npm": "~8.X",


radeksimko (approved):
https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location

👍🏻  thank you for identifying the problem and for the fix

Do you reckon this would affect https://github.com/hashicorp/vscode-terraform/issues/768 in any way?

jpogran:
> Do you reckon this would affect #768 in any way?

In the short term, no. Longer term we have to modify the manifest and extension to support web browser only deployment, so this change doesn't prevent that.

The `extensionKind` setting determines where the extension is loaded. The `workspace` value indicates:

> "extensionKind": ["workspace"] — Indicates the extension requires access to workspace contents and therefore needs to run where the workspace is located. That can be on the local machine or on the remote machine or Codespace. Most extensions fall into this category.

The extension relies on terraform-ls to provide most of the advanced language features and terraform-ls needs to run where the files are physically located to build a context for all features it provides. This means the extension needs to run where the workspace is located.

We could specify `"extensionKind": ["workspace","ui"]` so that the `ui` is a fallback, but I didn't test that combination. This might be a good idea to pursue before we merge this.

As for the web extension, in order for us to run there we have to either WASM the terraform-ls binary or modify the extension manifest to [declare its a web extension](https://code.visualstudio.com/api/extension-guides/web-extensions#web-extension-anatomy), which will be pretty involved for us. I don't think that is a concern for us until we start to work on that.

dbanck (approved):
Nice catch!

jpogran:
Updated the PR with `ui` and tested it out locally and in WSL. I don't have a SSH host up to check ATM and don't yet know how to test Codespaces.

PR #1004, opened 2022-03-21
Download syntax

jpogran:


build/downloader.ts
```diff
@@ -36,22 +37,26 @@ function getArch(arch: string) {
...
+    name: pjson.name,
     extensionVersion: pjson.version,
     languageServerVersion: pjson.langServer.version,
+    syntaxVersion: pjson.syntax.version,
     preview: pjson.preview,
   };
 }
 
-async function run(platform: string, architecture: string) {
+async function downloadLanguageServer(platform: string, architecture: string, extInfo: ExtensionInfo) {
```
radeksimko, line 59: Thank you for that rename 👍🏻

build/downloader.ts
```diff
@@ -94,6 +96,36 @@ async function run(platform: string, architecture: string) {
   });
 }
 
+async function downloadSyntax(info: ExtensionInfo) {
+  const release = `v${info.syntaxVersion}`;
+
+  const fileName = `${info.name}.tmGrammar.json`;
```
radeksimko, line 102: Would this work for preview?

radeksimko (commented):
Just one pending question in-line, otherwise LGTM.

package.json
```diff
@@ -342,18 +345,15 @@
     "pretest": "npm run download:ls && npm run test-compile && npm run lint",
     "test": "node ./out/test/runTest.js",
     "test:unit": "jest",
-    "test:syntax": "npm run test:grammar && npm run test:snap",
```
dbanck, line 345: Not directly related to this line, but I'm not able to comment above. Do we want to rename the npm script `download:ls` into something else, since it's now downloading multiple things?

dbanck (changes requested):
Thanks for the cleanup!

We need to update the `scopeName` in `package.json` from `source.terraform` to `source.hcl.terraform` to make the new grammar work.

package.json
jpogran: Went back and forth on this. Just rename to 'artifacts' or something else? Or do we want separate arguments and then:

```
    "download:artifacts": "ts-node ./build/downloader.ts",
    "download:ls": "ts-node ./build/downloader.ts -thing ls",
    "download:syntax": "ts-node ./build/downloader.ts -thing syntax",
```

Separate files altogether?

radeksimko: Assuming we change the reference in `vscode:prepublish` I'd just rename it to `download:artifacts` or anything similar for now, just to avoid confusion - we can address the separation later.

dbanck: `download:artifacts` sounds good to me, too.

radeksimko (approved):
:shipit:

PR #1019, opened 2022-03-29
Format semantic token settings

jpogran:
This adds a Github Action to check if there are any formatting changes to any PR submitted and ignores the package.json file when running prettier.

This is to prevent churn when adding new options to the package.json manifest.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,8 @@ jobs:
         run: npm install
       - name: lint
         run: npm run lint
+      - name: format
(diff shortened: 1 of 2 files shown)


dbanck (approved):
Good catch!

Sometimes I dislike the formatting done by prettier, because it's less readable now

jpogran:
Same ☹️

We could try to adjust rules for json files now that we've moved syntax out...

dbanck:
Are you suggesting special rules for json files? Or have prettier ignore them?

PR #1024, opened 2022-03-31
Organize extension settings into Sections

jpogran:
This groups and orders the existing settings into VS Code Setting sections while preserving the exact keys for each setting.

The extension previously utilized different prefixes to indicate different areas like terraform-ls.rootModules or terraform-ls.terraformExecTimeout. This leaves settings in a seemingly random order in the Settings UI and makes it hard to autocomplete inside the json file.

We solve this by organizing the extension settings into sections in the `contribute` part of the package.json manifest. These sections provide a logical order to extension settings, especially if the extension has many settings or many settings that are nested.

image

This also allows us to separate out configuration options, like a separate section for Experimental Options:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,129 +150,144 @@
         "path": "./snippets/terraform.json"
       }
     ],
-    "configuration": {
-      "type": "object",
-      "title": "Terraform",
-      "properties": {
-        "terraform.languageServer": {
-          "type": "object",
-          "description": "Language Server settings",
-          "properties": {
-            "external": {
-              "type": "boolean",
-              "default": true,
-              "description": "Whether an external language server binary should be launched."
+    "configuration": [
+      {
+        "title": "General",
+        "order": 0,
+        "properties": {
+          "terraform.codelens.referenceCount": {
+            "scope": "resource",
+            "type": "boolean",
+            "default": false,
+            "description": "Display reference counts above top level blocks and attributes."
+          }
+        }
+      },
+      {
+        "title": "Language Server",
+        "order": 1,
+        "properties": {
+          "terraform.languageServer": {
+            "order": "0",
+            "type": "object",
+            "description": "Language Server settings",
+            "properties": {
+              "external": {
+                "type": "boolean",
+                "default": true,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -150,129 +150,126 @@
...
+          "terraform-ls.ignoreDirectoryNames": {
+            "scope": "resource",
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "default": [],
+            "description": "Per-workspace list of directory names for the language server to ignore when indexing files"
+          },
+          "terraform-ls.experimentalFeatures": {
```
radeksimko, line 252: Am I right that there isn't much we can do about making this one editable (without breaking stuff)?
Screenshot 2022-04-05 at 16 02 32

radeksimko (approved):
LGTM - thank you for the UX improvement! 🙇🏻‍♂️  🚀

package.json
```diff
@@ -150,129 +150,126 @@
...
+            "order": 1,
+            "scope": "resource",
+            "type": "string",
+            "default": "",
+            "description": "Path to language server binary (optional)"
+          },
+          "terraform.languageServer.args": {
+            "order": 2,
+            "scope": "resource",
+            "type": "array",
```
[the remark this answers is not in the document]
jpogran: Good catch. I wasn't trying for that deep of an improvement yet but this is easy to add.

package.json
jpogran: I have some ideas. Let me test and get back to you.

jpogran:
> Actually I just opened the JSON file and this came up

It is supposed to be greyed out but still read the settings. It would be greyed out because it's technically not a setting anymore, but still read by VS Code because of the way JSON works.

While this worked in testing, it's too much magic to put in a PR that's purpose is to not affect the user. I have walked this back to just organizing sections and we will work in the `terraform.languageServer` extraction in a later PR.

dbanck (approved):
👍

radeksimko (approved):

PR #1031, opened 2022-04-06
Add terraform.languageServer.ignoreSingleFileWarning setting

jpogran:
Adds the terraform-ls setting that controls whether warnings are shown when editing single Terraform files.

Works with https://github.com/hashicorp/terraform-ls/pull/843

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,6 +220,11 @@
               "trace.server": "off"
             }
           },
+          "terraform.languageServer.ignoreSingleFileWarning": {
+            "scope": "resource",
+            "type": "boolean",
+            "description": "Path to the Terraform binary"
+          },
           "terraform-ls.rootModules": {
             "scope": "resource",
             "type": "array",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -220,6 +220,11 @@
               "trace.server": "off"
             }
           },
+          "terraform.languageServer.ignoreSingleFileWarning": {
+            "scope": "resource",
+            "type": "boolean",
+            "description": "Path to the Terraform binary"
+          },
```
radeksimko, line 227: Is there any benefit to setting `default: false` here as well?

jpogran: Good thing you asked, I missed adding a description here.

As for the setting, the `const ignoreSingleFileWarning = config('terraform').get<boolean>('languageServer.ignoreSingleFileWarning', false);` line in `clientHandler.ts` will ensure we get false if nothing is specified, but it's a good idea to be explicit in both places so I added it.

PR #1032, opened 2022-04-06
Prevent preview from activating when stable is enabled

jpogran:
This PR prevents the preview extension from activating when the stable extension is enabled, or vice versa.

image

To test:

> Note: You cannot test this inside the Extension host.

- Package hashicorp.terraform
- Package hashicorp.terraform-preview
- Install both in a new VS Code Install (or ensure your ~/.vscode/extensions directory does not have any existing installs)

On installing the second VSIX, you should get an error message popup. Disabling one and reloading should show a properly activated extension. Enabling the other one and disabling the first should produce a error message pop up.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -23,6 +23,17 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
   reporter = new TelemetryReporter(context.extension.id, manifest.version, manifest.appInsightsKey);
   context.subscriptions.push(reporter);
 
+  if (context.extension.id === 'hashicorp.terraform-preview') {
+    const stable = vscode.extensions.getExtension('hashicorp.terraform');
+    if (stable !== null) {
+      vscode.window.showErrorMessage(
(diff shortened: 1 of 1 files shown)


dbanck (approved):
👍 works for me!

radeksimko (approved):
Works well - as far as I can test it before a release 😄  :shipit:

src/extension.ts
```diff
@@ -23,6 +23,25 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
   reporter = new TelemetryReporter(context.extension.id, manifest.version, manifest.appInsightsKey);
   context.subscriptions.push(reporter);
 
+  const stable = vscode.extensions.getExtension('hashicorp.terraform');
+  const preview = vscode.extensions.getExtension('hashicorp.terraform-preview');
+
+  if (context.extension.id === 'hashicorp.terraform-preview') {
+    if (stable !== undefined) {
```
radeksimko, line 30: Is it worth adding a comment here to say that `getExtension` returns `undefined` when the extension ID is either not installed or disabled? Just to clarify...

PR #1036, opened 2022-04-07
Disable naming convention warning for Code Action identifier

jpogran:
The Typescript naming convention does not apply to Code Action identifiers, but is being triggerd here because it's being used as a object property key to update the VS Code settings in this test. This can be safely ignored.


dbanck (approved):
🚀

PR #1048, opened 2022-04-13
Deprecate terraform.languageServer.trace.server

jpogran:


radeksimko (approved):
Given that this is already no-op (and AFAIK always has been - at least throughout v2 versions) I'd be also okay with just removing it, but I'm unsure of the UX implications for the end users who have this declared in their configs.

jpogran:
> Given that this is already no-op (and AFAIK always has been - at least throughout v2 versions) I'd be also okay with just removing it, but I'm unsure of the UX implications for the end users who have this declared in their configs.

Removing it entirely does not break anything, it will just be grayed out in the JSON file and won't show at all in the UI. If a user hovers over the setting in the JSON file, they'll be given a general message that the setting is no longer used.

radeksimko:
> I'd rather remove entirely than wait 2 release cycles for something that does nothing.

👍🏻  Same, let's remove it then

radeksimko (approved):

PR #1051, opened 2022-04-14
Github Release Notes Generator file

jpogran:
This adds a configuration file to the repo for the Github Release Notes
API to read. This does not create a release or any other change to the
repo, it just generates a changelog that is very close to how we
currently format the changelog. This aids in reconciling changes before
release time.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,32 @@
+changelog:
+  categories:
+    - title: BREAKING CHANGES
+      labels:
+        - breaking
+    - title: NOTES
+      labels:
+        - note
+        - documentation
+    - title: FEATURES
+      labels:
+        - feature
+    - title: ENHANCEMENTS
+      labels:
+        - enhancement
+    - title: BUG FIXES
+      labels:
+        - bug
+    - title: INTERNAL
+      labels:
+        - dependencies
+        - ci
+        - "*"
+  exclude:
+    labels:
+      - duplicate
+      - wontfix
+      - question
+      - invalid
+      - needs-reproduction
+      - needs-research
+      - stale


.github/release.yml
```diff
@@ -0,0 +1,32 @@
+changelog:
+  categories:
+    - title: BREAKING CHANGES
+      labels:
+        - breaking
```
radeksimko, line 5: Now I understand why you wanted to label BCs 😄

.github/release.yml
```diff
@@ -0,0 +1,32 @@
...
+        - "*"
+  exclude:
+    labels:
+      - duplicate
+      - wontfix
+      - question
+      - invalid
+      - needs-reproduction
+      - needs-research
+      - stale
```
radeksimko, line 32: I think we don't actually have most of these - is it worth revising the list to reflect the reality?

.github/release.yml
```diff
@@ -0,0 +1,32 @@
...
+    - title: BREAKING CHANGES
+      labels:
+        - breaking
+    - title: NOTES
+      labels:
+        - note
+        - documentation
+    - title: FEATURES
+      labels:
+        - feature
```
radeksimko, line 12: We also don't track features separately from enhancements (i.e. there's no such label currently) - unsure if we should and where do we even draw the line between the two

.github/release.yml
jpogran: It does no harm to have them here, it's a catch if there is something that did get labeled as such since the labels exist. I can remove the labels from the repo and here if we think that's best, or can merge as is and we can revisit the labels in general.

.github/release.yml
jpogran: I'm find with collapsing to Enhancements, as it can cover both an improvement or an addition while Feature means different things to different peopl

.github/release.yml
jpogran: Removed nonexistant

.github/release.yml
jpogran: Removed as there isn't a features label anymore

radeksimko (approved):
Overall I appreciate the effort that goes into automation! 😄 

I'm sure we can discuss this in more detail synchronously, but I feel I could share some related thoughts around Changelog management here.

 - PR titles are IMO often written from dev perspective, not so often from end-user one and so 1-1 mapping between PR titles and changelog entries can be tricky. We'd always have to ensure the PR title is modified accordingly, or fix a generated changelog later.
 - Sometimes there are PRs which warrant more than 1 changelog entry (especially when it comes to breaking changes or bigger PRs) as they have multiple different end-user implications.
 - Crafting changelog at release-time (as opposed to merge-time) means we have to remember the context for each PR that is being released, which can be hard esp. if the person cutting the release is not the PR author and esp. if the PR was merged weeks ago.

That said this PR isn't setting anything in stone - as you mentioned it's rather a convenience helper. 👍🏻

PR #1063, opened 2022-04-19
Remove terraform-ls.rootModules and terraform-ls.excludeRootModules

jpogran:
These settings are no longer used by terraform-ls.


radeksimko (changes requested):
These ones are still in use actually and will remain in use at least until https://github.com/hashicorp/terraform-ls/issues/704 is addressed and more specifically https://github.com/hashicorp/terraform-ls/issues/724

PR #1064, opened 2022-04-20
Add OpenVSX publishing

### LINKED ISSUE #379: Publish extension on open-vsx.org
Hi!

Congrats to the 2.0 release! I'm very much looking forward to using it often and hope that I will no longer have bogus syntax validation errors :)

Could you also [publish](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions#how-to-publish-an-extension) your Terraform VS Code Extension on https://open-vsx.org/ ?

This would make it much easier to use the awesome extension in project such as
- [gitpod.io](https://gitpod.io)
- [theia-ide.org](https://theia-ide.org)
- [vscodium.com](https://github.com/VSCodium/vscodium/issues/372)
- [code-server](https://github.com/cdr/code-server)

because the [ToS](https://cdn.vsassets.io/v/M146_20190123.39/_content/Microsoft-Visual-Studio-Marketplace-Terms-of-Use.pdf) doesn't allow those projects to obtain the extension from  Microsoft's marketplace:
> Marketplace Offerings are intended for use only with Visual Studio Products and Services and you may only install and use Marketplace Offerings with Visual Studio Products and Services.

filiptronicek:
This is a succession to https://github.com/hashicorp/vscode-terraform/pull/388, which adapts publishing to OpenVSX embedded in the publishing workflow in GitHub Actions.

Fixes #379 

Before using this it is necessary to set up an `OVSX_PAT` GitHub Action secret with the value from [open-vsx.org/user-settings/tokens](https://open-vsx.org/user-settings/tokens). A claim of the `hashicorp` namespace will be also required (this can be done by submitting [an issue](https://github.com/EclipseFdn/open-vsx.org/issues/new?assignees=&labels=namespace&template=claim-namespace-ownership.md&title=Claiming+namespace+%5Bname%5D) in the OpenVSX repo).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -91,6 +91,6 @@ jobs:
     steps:
       - uses: actions/download-artifact@v3
       - name: Publish Extension
-        run: npx ovsx publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
+        run: npx ovsx publish --packagePath $(find . -iname *.vsix)
         env:
           OVSX_PAT: ${{ secrets.OVSX_PAT }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,18 +71,26 @@ jobs:
           name: ${{ matrix.vsce_target }}
           path: "*.vsix"
 
-  publish:
-    name: Publish All
+  publish-vsce:
+    name: Publish All (Microsoft Marketplace)
     runs-on: ubuntu-latest
     needs: build
     if: success() && startsWith( github.ref, 'refs/tags/v')
     steps:
       - uses: actions/download-artifact@v3
-      - name: Publish Extension (Microsoft Marketplace)
+      - name: Publish Extension
         run: npx vsce publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
-      - name: Publish Extension (OpenVSX)
+
+  publish-ovsx:
+    name: Publish All (OpenVSX)
+    runs-on: ubuntu-latest
+    needs: build
+    if: success() && startsWith( github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/download-artifact@v3
+      - name: Publish Extension
         run: npx ovsx publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
         env:
           OVSX_PAT: ${{ secrets.OVSX_PAT }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -78,7 +78,11 @@ jobs:
     if: success() && startsWith( github.ref, 'refs/tags/v')
     steps:
       - uses: actions/download-artifact@v3
-      - name: Publish Extension
+      - name: Publish Extension (Microsoft Marketplace)
         run: npx vsce publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
+      - name: Publish Extension (OpenVSX)
+        run: npx ovsx publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
+        env:
+          OVSX_PAT: ${{ secrets.OVSX_PAT }}


filiptronicek:
Hey @radeksimko, thanks for the quick review!

> Would you mind signing the CLA?

Working on that :)

> Would you mind decoupling the OpenVSX into a separate job?

I would not at all, but maybe we could implement something a bit different which would wouldn't require its own step. We could essentially have the setup I made, but we would use `if: always()` on the OpenVSX step so that these two registries publish independently without having a successful publish to the other one.

> the two publishing jobs can run in parallel

This is a fair point that would not work with the approach I'm proposing, but I think the ~30 seconds it takes the job to publish all of the extension versions could be tolerated because the GitHub Actions runners often have different boot times (hence this approach could be faster sometimes, and sometimes slower).

radeksimko:
So the main benefit of splitting it into its own job is that GitHub allows us to rerun a failed job, but we cannot rerun individual steps unfortunately.

Screenshot 2022-04-20 at 16 25 30

So let's say that OpenVSX publishing fails and Marketplace succeeds - if we had both as steps - we'd have to retry both, or have some extra logic to ensure that repeated publishing to the Marketplace is harmless. More likely we just wouldn't retry.

> but we would use if: always() on the OpenVSX step so that these two registries publish independently without having a successful publish to the other one.

Right, I wasn't entirely clear but this is the main benefit I had in mind when talking about parallel jobs, more than the time saved. I think the two jobs can still be independent of each other and both use something like this?

```yaml
    needs: build
    if: success() && startsWith( github.ref, 'refs/tags/v')
```

filiptronicek:
> I think the two jobs can still be independent of each other and both use something like this?

I think so as well.

All other comments sound fair, I will add it as a separate job just like the MS publish. My main concern here was just duplicate workflow code, but that could perhaps be addressed in a future PR with moving some steps into their own `.yml` files.

radeksimko (commented):
This LGTM, thank you for the edits and for signing the CLA.

I'm just in the process of clarifying one small part of the Terms with the Eclipse Foundation, as asked by our Legal and once done I will go ahead with the namespace transfer and PAT creation, and _then_ we can merge this.

Thanks for your patience.

.github/workflows/publish.yml
```diff
@@ -82,3 +82,15 @@ jobs:
...
+
+  publish-ovsx:
+    name: Publish All (OpenVSX)
+    runs-on: ubuntu-latest
+    needs: build
+    if: success() && startsWith( github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/download-artifact@v3
+      - name: Publish Extension
+        run: npx ovsx publish --no-git-tag-version --packagePath $(find . -iname *.vsix)
```
radeksimko, line 94:
```suggestion
        run: npx ovsx publish --packagePath $(find . -iname *.vsix)
```

This flag is actually not supported in ovsx from what I can tell.

radeksimko (commented):
@filiptronicek We have claimed the namespace and have a PAT ready, but there's still an issue with publishing as the latest version of ovsx with the target support isn't available on npm - see https://github.com/eclipse/openvsx/issues/450

Not realizing I used an old version (0.3.0) I managed to accidentally already publish two versions as universal (0.20.0 & 0.20.1) - is there any way of either retracting them or republishing as platform specific?

.github/workflows/publish.yml
filiptronicek: I am not sure myself, but I suppose if removing it doesn't break anything, should be fine.

.github/workflows/publish.yml
```diff
@@ -82,3 +82,15 @@ jobs:
...
+
+  publish-ovsx:
+    name: Publish All (OpenVSX)
+    runs-on: ubuntu-latest
+    needs: build
+    if: success() && startsWith( github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/download-artifact@v3
+      - name: Publish Extension
+        run: npx ovsx publish --packagePath $(find . -iname *.vsix)
```
dbanck, line 94:
```suggestion
        run: npx ovsx publish --packagePath $(find . -iname '*.vsix' ! -name '*web*')
```

We now publish a web version (https://github.com/hashicorp/vscode-terraform/pull/1210) of the extension as well, which — _I think_ — we need to filter here. This change should ensure we only upload the platform-specific packages while skipping the web one.

dbanck (changes requested):
All issues regarding platform-specific extensions on the OpenVSX side seem to have been resolved. I was able to upload all the missing versions manually. :tada:

After one more minor change, we can finally merge this PR. Thank you for being so patient.

.github/workflows/publish.yml
filiptronicek: Happy to commit this, @dbanck, and thanks for taking care of the publishing. 

I'm unsure if this is a necessary change, though: if the extension is offered in `web` form as well, it will work properly for VS Code Web, otherwise, it shouldn't be labelled as compatible.

From the VS Code Docs: [[1](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions)]

> If you want a platform-specific extension to also support running in the browser as a [web extension](https://code.visualstudio.com/api/extension-guides/web-extensions), it must target the web platform when publishing. The web platform will respect the browser entry point in the package.json. The extension capabilites that are not supported in the web should use when clauses in the package.json to make sure that they are disabled. We do not recommend that extensions ship a different package.json for the web platform or to remove parts of the VSIX that do not work in the web.

dbanck: We offer a web version of the extension, which works in VS Code Web. We added an additional target to our publishing pipeline and skipped bundling the language server (https://github.com/hashicorp/vscode-terraform/pull/1262).

We weren't sure if OpenVSX also supports `web` as a target when publishing, so I proposed filtering the package. Fantastic if that's not required and there is `web` target support! Do you know if that's the case? I did not try to publish the web `vsix` because I didn't want to break anything.

dbanck: You were right, this change isn't necessary, and OpenVSX supports `web` targets.

CleanShot 2022-11-23 at 19 55 19@2x

dbanck (approved):

radeksimko (approved):

PR #1072, opened 2022-04-25
dependabot: Reduce interval for dev dependencies to weekly

radeksimko:
The motivation is to reduce the noise coming mainly from frequently updated `@types/*` packages which are not critical for us, along with most dev dependencies.

`@hashicorp/js-releases` is one we maintain ourselves and don't release as frequently, so that shouldn't present the same challenge as the others.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,19 @@ updates:
     schedule:
       interval: "daily"
     labels: ["dependencies"]
+    allow:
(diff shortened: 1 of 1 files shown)


dbanck (approved):
👍

PR #1073, opened 2022-04-25
Organize Static Features

jpogran:
This moves all custom Static Feature classes to a central folder. It also consolidates the Semantic Token manifest parsing helper methods and moves them to the CustomSemanticTokens class.


src/clientHandler.ts
```diff
@@ -37,6 +37,7 @@ export class ClientHandler {
     private lsPath: ServerPath,
     private outputChannel: vscode.OutputChannel,
     private reporter: TelemetryReporter,
+    private manifest: PartialManifest,
```
radeksimko, line 40: Since this is now used outside the scope of semantic tokens I wonder if it's worth calling it e.g. `ManifestWithSemanticTokens`? _or_ making a copy of that interface here which can be expanded and used elsewhere later - presumably one that is compatible with `PartialManifest` within `semanticTokens`.

I suppose there are other places where we access the manifest for some other properties.

radeksimko (approved):
Thanks for the cleanup! 👍🏻 

Just one question - but not blocking.

src/clientHandler.ts
jpogran: I would agree if we didn't have more work to break up the dependencies inside ClientHandler. 

ClientHandler only needs to know about this because of the StaticHandler registrations. Building up a StaticHandler should be done outside ClientHandler and passed in rather than passing in all information required to build several StaticHandlers.

I could do this here now, or do that work with the next planned PR for refactoring ClientHandler. I prefer to make those changes in the next PR.

PR #1075, opened 2022-04-25
Remove command prefix

jpogran:
This removes generating a command prefix when registering commands for the language client connecting to terraform-ls.

Previously the extension would create a LanguageClient/terraform-ls pair per 'folder' in a workspace. This means there was a terraform-ls process per folder to keep track of. The https://github.com/hashicorp/terraform-ls/pull/502 PR enabled a single terraform-ls process to handle multi-folder workspaces, which meant we could refactor the extension In https://github.com/hashicorp/vscode-terraform/pull/845 to use a single LanguageClient instance. A leftover from the original design was a prefix for commands registered between terraform-ls and the client. This ensured unique command IDs when multiple clients were present, but is no longer necessary with a single language client.

Acceptance tests:

- Open a single folder and use any LS based commands
- Open a single folder, then use `Add Folder to workspace` to open a multi-folder workspace
- Open a multi-folder workspace directly using a workspace file like `something.code-workspace`
- In any example, open one of the Module Views, wait for items to show up, then close editor (do not close views). Then re-open editor. The Module views should show a progress bar until the LS is ready and responds with data, without error.

In each case, the extension should start with no errors. When adding additional folders to the workspace or when opening a mult-folder workspace directly, there should be no errors regarding duplicate commands.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MPL-2.0",
       "dependencies": {
         "@vscode/extension-telemetry": "^0.4.9",
-        "short-unique-id": "^3.2.3",
         "vscode-languageclient": "^7.0.0",
         "vscode-uri": "^3.0.2",
         "which": "^2.0.2"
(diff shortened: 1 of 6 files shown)


dbanck (approved):
LGTM 👍 

I wasn't able to reproduce any error cases we encountered in the past :)

PR #1079, opened 2022-04-26
Optimize main entry point execution path

jpogran:
This extracts functions, re-orders execution and otherwise cleans up the main entry point for the extension.

Migrating settings is extracted to a function to prepare for the work to be done renaming settings later on.

The command registration calls are reorganized to prioritize registering first those that do not require the terraform-ls to function. Starting the terraform-ls binary is then called directly before registering the rest of the commands and providers.

The end result is registering and activating the parts that need the LS only if it is enabled.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -68,25 +68,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     vscode.commands.registerCommand('terraform.apply', async () => {
       await terraformCommand('apply', false);
     }),
-    vscode.commands.registerCommand('terraform.init', async () => {
-      const workspaceFolders = vscode.workspace.workspaceFolders;
-      const selected = await vscode.window.showOpenDialog({
-        canSelectFiles: false,
-        canSelectFolders: true,
-        canSelectMany: false,
-        defaultUri: workspaceFolders ? workspaceFolders[0]?.uri : undefined,
-        openLabel: 'Initialize',
-      });
-      const client = clientHandler.getClient();
-      if (selected && client) {
-        const moduleUri = selected[0];
-        const requestParams: ExecuteCommandParams = {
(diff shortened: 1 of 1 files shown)


dbanck (changes requested):
LGTM, but there are some more commands which require the LS and can be moved to the bottom subscription block:
* `'terraform.validate'`
* `'terraform.initCurrent'`
* `'terraform.init'`

jpogran:
> LGTM, but there are some more commands which require the LS and can be moved to the bottom subscription block:
> 
> * `'terraform.validate'`
> * `'terraform.initCurrent'`
> * `'terraform.init'`

They do *use* the LS, but if you look at `terraformCommand` in [`extension.ts#287`](https://github.com/hashicorp/vscode-terraform/blob/main/src/extension.ts#L287) it will use the terminal if there is no LS enabled. I went back and forth which section they 'belong' in because of this.

dbanck:
> They do _use_ the LS, but if you look at `terraformCommand` in [`extension.ts#287`](https://github.com/hashicorp/vscode-terraform/blob/main/src/extension.ts#L287) it will use the terminal if there is no LS enabled. I went back and forth which section they 'belong' in because of this.

Oh, sorry I missed that. I was fooled by the `languageServerExec = true` parameter.

I think we still need to move `'terraform.init'`, because it's using `const client = clientHandler.getClient();` and `clientHandler` will be `undefined`.

jpogran:
Ugh, these commands really need a review. If you look at `terraform.init` it checks if client is undefined and returns silently if it is undefined. So, if the LS is not available/started/whatever, the command does not run and the user won't know why.

Still moving it because it needs the LS, just noting this for when we return to these commands.

dbanck (approved):
🚢

Issue #791, opened 2021-09-23
Review Extension Settings and Plan Future direction


jpogran:
## Problem Statement

The current settings for the Terraform VS Code Extension are an amalgam of historical settings from the time before HashiCorp took ownership of the extension and settings we have added since taking it over.

Some settings are present that are not used in either vscode-terraform or in terraform-ls. Some settings are present in vscode-terraform, but not used by terraform-ls. Some settings are confusingly named and would benefit from more precise naming. Other settings do not follow common VS Code setting practices.

### Flatten `terraform.languageServer` to single level settings

- `terraform.languageServer.external`
- `terraform.languageServer.pathToBinary`
- `terraform.languageServer.args`

Assign `order` values to each so these are presented first in the settings list.

### Flatten `terraform.experimentalFeatures` to single level settings

- `terraform.experimentalFeatures.validateOnSave`
- `terraform.experimentalFeatures.prefillRequiredFields`

### Rename settings

Existing | New
-- | --
`terraform.languageServer.external` | `terraform.languageServer.enabled`
`terraform-ls.terraformExecPath` | `terraform.languageServer.terraform.execPath`
`terraform-ls.terraformExecTimeout` | `terraform.languageServer.terraform.execTimeout`
`terraform-ls.terraformLogFilePath` | `terraform.languageServer.terraform.logFilePath`

Add steps to automatically rename these client side in [extension.ts#L51-L62](https://github.com/hashicorp/vscode-terraform/blob/main/src/extension.ts#L51-L62)

### Remove Settings

- `terraform.languageServer.trace.server`
- `terraform.languageServer.maxNumberOfProblems`

### Deprecate Settings

- `terraform.languageServer.excludeRootModules`
- `terraform.languageServer.rootModules`

### Final settings:

- `terraform.codelens.referenceCount`
- `terraform.languageServer.enabled`
- `terraform.languageServer.pathToBinary`
- `terraform.languageServer.args`
- `terraform.languageServer.ignoreSingleFileWarning`
- `terraform.languageServer.experimentalFeatures`
- `terraform.languageServer.trace.server`
- `terraform.languageServer.terraform.execPath`
- `terraform.languageServer.terraform.execTimeout`
- `terraform.languageServer.terraform.logFilePath`
- `terraform.languageServer.rootModules`
- `terraform.experimentalFeatures.validateOnSave`
- `terraform.experimentalFeatures.prefillRequiredFields`

PR #1082, opened 2022-04-28
Extract LanguageClient from ClientHandler

jpogran:
This extracts the LanguageClient creation from ClientHandler and puts it directly inside the activation point.

In order to implement https://github.com/hashicorp/terraform-ls/issues/722 we need to add a new StaticFeature to register a client-side command for terraform-ls to call when it knows the `terraform.providers` view needs to be refreshed.

StaticFeatures are registered using the LanguageClient.registerFeature method, which means the ClientHandler class needs to create the StaticFeature. The StaticFeature needs both the LanguageClient and `terraform.providers` view created before it can be initialized. The LanguageClient is created by the ClientHandler class. This all results in a cyclic dependency.

We could resolve this cyclic dependency by adding more responsibility to ClientHandler, but this introduces more complexity to the class. This will become increasingly hard to deal with as more aspects like StaticFeature are added.

We resolve this by extracting the creation of LanguageClient and moving dependent features like the StaticFeatures to the main activation method. This puts all the parts that rely on each other in the same place where the data needed is located to make decisions about whether they are activated or not.

This builds on work done in:

- https://github.com/hashicorp/vscode-terraform/pull/1073
- https://github.com/hashicorp/vscode-terraform/pull/1074
- https://github.com/hashicorp/vscode-terraform/pull/1075
- https://github.com/hashicorp/vscode-terraform/pull/1079

### DIFF
--- a/src/clientHandler.ts
+++ b/src/clientHandler.ts
@@ -1,168 +1,53 @@
-import * as vscode from 'vscode';
-import TelemetryReporter from '@vscode/extension-telemetry';
-import {
-  DocumentSelector,
-  Executable,
-  LanguageClient,
-  LanguageClientOptions,
-  RevealOutputChannelOn,
-  ServerOptions,
-  State,
-} from 'vscode-languageclient/node';
+import { Executable, ServerOptions } from 'vscode-languageclient/node';
 import { config } from './utils/vscode';
 import { ServerPath } from './utils/serverPath';
-import { PartialManifest, CustomSemanticTokens } from './features/semanticTokens';
-import { ShowReferencesFeature } from './features/showReferences';
-import { TelemetryFeature } from './features/telemetry';
 
-/**
- * ClientHandler maintains lifecycles of language clients
- * based on the server's capabilities
- */
-export class ClientHandler {
-  private client: LanguageClient | undefined;
-  private commands: string[] = [];
+export async function getServerOptions(lsPath: ServerPath): Promise<ServerOptions> {
+  const cmd = await lsPath.resolvedPathToBinary();
+  const serverArgs = config('terraform').get<string[]>('languageServer.args', []);
 
-  public extSemanticTokenTypes: string[] = [];
-  public extSemanticTokenModifiers: string[] = [];
+  const executable: Executable = {
+    command: cmd,
+    args: serverArgs,
+    options: {},
+  };
+  const serverOptions: ServerOptions = {
+    run: executable,
+    debug: executable,
+  };
 
-  constructor(
-    private lsPath: ServerPath,
-    private outputChannel: vscode.OutputChannel,
-    private reporter: TelemetryReporter,
-    private manifest: PartialManifest,
-  ) {
-    if (lsPath.hasCustomBinPath()) {
-      this.reporter.sendTelemetryEvent('usePathToBinary');
-    }
+  // this.outputChannel.appendLine(`Launching language server: ${cmd} ${serverArgs.join(' ')}`);
+
+  return serverOptions;
+}
+
+export function getInitializationOptions() {
+  const rootModulePaths = config('terraform-ls').get<string[]>('rootModules', []);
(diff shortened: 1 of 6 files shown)


src/clientHandler.ts
```diff
@@ -1,168 +1,53 @@
...
-  public getClient(): LanguageClient | undefined {
-    return this.client;
-  }
+  return serverOptions;
+}
 
-  public clientSupportsCommand(cmdName: string): boolean {
-    return this.commands.includes(cmdName);
-  }
+export function getInitializationOptions() {
```
jpogran, line 24: This is basically a set of settings masquerading as a class method. My intention here is to make room for this to be added to a configuration builder when we tackle #791

src/clientHandler.ts
```diff
@@ -1,168 +1,53 @@
...
 
-/**
- * ClientHandler maintains lifecycles of language clients
- * based on the server's capabilities
- */
-export class ClientHandler {
-  private client: LanguageClient | undefined;
-  private commands: string[] = [];
+export async function getServerOptions(lsPath: ServerPath): Promise<ServerOptions> {
+  const cmd = await lsPath.resolvedPathToBinary();
```
jpogran, line 6: This function has little reason to exist outside of handling LS paths and args. When we address #791 we can fold this into the main client activation

src/providers/moduleCalls.ts
```diff
@@ -131,13 +131,16 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
...
     }
-    await handler.onReady();
+    await this.handler.onReady();
 
-    const commandSupported = this.handler.clientSupportsCommand(`terraform-ls.module.calls`);
+    // clientSupportsCommand(`terraform-ls.module.calls`)
+    const commandSupported =
+      this.handler.initializeResult?.capabilities.executeCommandProvider?.commands.includes(
+        'terraform-ls.module.calls',
+      );
```
jpogran, line 143: We lose the neat `clientSupportsCommand` function by removing ClientHandler so we have to use this long property accessor here to check if the command is supported.

Since we ship the LS with the extension now and the versions are in lockstep, we don't technically need to check anymore if it's supported, but I'm leaving this for now as it does not harm anything to check here.

dbanck: Maybe we can add a small function to `clientHelpers.ts` and keep the check just to be safe?

```ts
export function clientSupportsCommand(initializeResult: InitializeResult | undefined, cmdName: string): boolean {
  if (!initializeResult) {
    return false;
  }

  return initializeResult.capabilities.executeCommandProvider?.commands.includes(cmdName) ?? false;
}
```

And call it everywhere, where we check for commands:

```ts
const commandSupported = clientSupportsCommand(this.client.initializeResult, 'terraform-ls.module.calls');
```

src/extension.ts
```diff
@@ -85,9 +101,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+
+  client = new LanguageClient(id, serverOptions, clientOptions);
+  client.onDidChangeState((event) => {
+    console.log(`Client: ${State[event.oldState]} --> ${State[event.newState]}`);
+    if (event.newState === State.Stopped) {
+      reporter.sendTelemetryEvent('stopClient');
+    }
+  });
+
+  const features: StaticFeature[] = [new CustomSemanticTokens(client, manifest)];
```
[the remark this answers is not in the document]
dbanck: I think the `if/else` here is fine, since we're avoiding creating an instance of the static feature if it's not required.

dbanck (approved):
Really nice refactoring! I think the code is much easier to navigate now.

Aside from the suggested `clientSupportsCommand` helper, everything LGTM 👍

src/providers/moduleCalls.ts
jpogran: Done!

PR #1084, opened 2022-04-28
Automatically refresh Terraform Providers View when providers change in open document

jpogran:
This introduces a StaticFeature which registers a handler for the `client.refreshModuleProviders` request which terraform-ls can call to refresh the data shown in the Terraform Providers view. The terraform-ls server was modified to send `client.refreshModuleProviders` requests in https://github.com/hashicorp/terraform-ls/pull/902


radeksimko (approved):
LGTM - works great! :shipit:

PR #1089, opened 2022-04-29
Update CHANGELOG

jpogran:


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+## [2.23.0] (Unreleased)
+
+BREAKING CHANGES:
+
+ - Remove `terraform.languageServer.maxNumberOfProblems` ([1062](https://github.com/hashicorp/vscode-terraform/pull/1062))
```
radeksimko, line 5: I'd add some context here so it doesn't look as scary. We are unlikely to break anything since the setting was not used since `v2.0.0`.

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+
+ENHANCEMENTS:
+
+ - Add Module Providers view refresh button ([1065](https://github.com/hashicorp/vscode-terraform/pull/1065))
+
+INTERNAL:
+
+ - dependabot: Reduce interval for dev dependencies to weekly ([1072](https://github.com/hashicorp/vscode-terraform/pull/1072))
+ - Organize Static Features ([1073](https://github.com/hashicorp/vscode-terraform/pull/1073))
```
radeksimko, line 14: I'd argue this is a user-visible enhancement, so we'd be down-selling it here under `INTERNAL`? 😄

radeksimko (approved):
Just a few suggestions left in-line. btw. the format LS at least follows is that links to PRs are prepended with `#`.

Otherwise LGTM.

CHANGELOG.md
jpogran: Moving classes around is user-visible? Are you referring to #1082?

radeksimko: My bad - I was referring to https://github.com/hashicorp/vscode-terraform/pull/1024 which was already shipped :fades-away:

radeksimko (approved):
:shipit:

PR #1094, opened 2022-05-03
Remove vscode-tmgrammar-test

radeksimko:
This reflects the new reality of syntax files living under `hashicorp/syntax` along with tests, so we no longer need the testing library here.


package.json
```diff
@@ -93,56 +93,170 @@
...
-	    {"id": "hcl-string", "superType": "string", "description": "String"},
-	    {"id": "hcl-number", "superType": "number", "description": "Number"},
-	    {"id": "hcl-objectKey", "superType": "parameter", "description": "Object key"},
-	    {"id": "hcl-mapKey", "superType": "parameter", "description": "Map key"},
-	    {"id": "hcl-keyword", "superType": "variable", "description": "Keyword"},
-	    {"id": "hcl-traversalStep", "superType": "variable", "description": "Traversal step"},
-	    {"id": "hcl-typeCapsule", "superType": "keyword", "description": "Type (complex)"},
-	    {"id": "hcl-typePrimitive", "superType": "keyword", "description": "Type (primitive)"}
+      {
+        "id": "hcl-attrName",
```
jpogran, line 97: Did you mean to format the semantic token blocks?

radeksimko: To be honest I don't know what the preferred format is - if there is any. `npm install` seemed to have done the reformatting, so I just went ahead with it. 🤷🏻‍♂️

jpogran: Ah that explains it. Daniel and I disabled formatting it in #1019 because of the verbosity of these entries. Maybe we should reconsider.

radeksimko: I don't disagree with you on the verbosity, but I'm not sure there's an easy way of avoiding it in `npm` directly 🤔

jpogran (approved):
We'll figure out formatting in a separate PR.

PR #1113, opened 2022-05-20
github: Update issue templates

radeksimko:
This proposes two changes:

1. Removes the tickboxes - it doesn't seem to bring much value as anecdotally many people just leave it out
2. LS is now bundled with the extension, so we no longer need to ask for the version as it's implied from extension version.


jpogran (approved):
👍🏼

PR #1131, opened 2022-06-01
Fix enabling or disabling language server commands

jpogran:
This updates the method the commands to enable or disable the language server use to modify the extension configuration. Previously this relied on untyped calls which are not allowed anymore for this method.

This will allow https://github.com/hashicorp/vscode-terraform/pull/1117 to be merged.


src/extension.ts
```diff
@@ -54,26 +54,38 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+        return stopLanguageServer();
       }
-      return stopLanguageServer();
+
+      const current: LanguageServerSettings = config('teraform').get<LanguageServerSettings>('languageServer', {
+        external: true,
+        pathToBinary: '',
+        args: ['serve'],
+        ignoreSingleFileWarning: false,
+      });
```
radeksimko, line 83: I'm not as familiar with this syntax here - do you mind explaining the implications here for me? Is the second `get()` argument basically just default value used in case that `languageServer` section is not found? If so - that's fine 👍🏻 

Just making sure we don't override some other settings like custom `args` just by enabling or disabling the server. 😅

radeksimko (approved):
Good catch!

Just one question and two minor suggestions, otherwise LGTM!

## Before
Screenshot 2022-06-06 at 12 03 50

## After
Screenshot 2022-06-06 at 12 05 17

--- 

Completely unrelated to this PR, but it would be amazing if we could enable and disable the LS without reload. 😄

jpogran:
> Completely unrelated to this PR, but it would be amazing if we could enable and disable the LS without reload. 😄

This is completely doable from a technical standpoint:

`vscode.commands.executeCommand("workbench.action.reloadWindow")`

This is the same as what we do in the `onDidChangeConfiguration` block.

src/extension.ts
jpogran: The `vscode.WorkspaceConfiguration.get()` method returns an object *or* undefined if no setting is found that matches the string your provide. So to get the default value, there has to be no setting found that matches `terraform.languageServer`.

If we instead change it to allow undefined as a value, we have to check if the value returned is undefined:

```javascript
const current: LanguageServerSettings | undefined = config('teraform').get<LanguageServerSettings>('languageServer');
if (current === undefined) {
  current = {
    external: true,
    pathToBinary: '',
    args: ['serve'],
    ignoreSingleFileWarning: false,
  };
}
```

Which is what the original method is already doing.

Why do we have to do this?

The root of this problem is that we made an untyped JSON object out of `languageServer` setting, instead of having separate fields. This has caused/is causing problems everywhere we try to access and change these setting values. Since it's treated like one object we can't set one value (`external`), we have to set all the values at the same time. In short, the `vscode.WorkspaceConfiguration.update()` method operates on `terraform.languageServer` not `terraform.languageServer.external`. To complicate it further, these settings can be different depending on scope (user, workspace, folder, etc).

To handle this previously we pulled the existing setting values (albeit incompletely) and used `Object.assign` to map the new value onto the existing values. This now does not work because `Object.assign` does not accept `any` as an object to work on anymore. We have to provide a typed object. 

To fix this we we provide the small `LanguageServerSettings` interface to the `vscode.WorkspaceConfiguration.get()` method which now returns typed object that we can set values to, then pass to `vscode.WorkspaceConfiguration.update()`  and have it work.

One last thing to consider is scope. We read the value several times before we assign it. VS Code automatically figures out the value from the scopes when we call get, so we don't have to specify scope. We do have to specify scope when updating the settings. We get scope by using the `vscode.WorkspaceConfiguration.inspect()` method, which gives us the information on which scope to update.

This will be all avoided when we do #791, which will allow us to only update one value at a time.

jpogran:
Stopping does not strictly require a reload, so I can see the possibility there.

Starting would possibly require a reload in order to let VS Code know there is an LS there to do something with, but unsure about that without testing. Agreed, this is another PR of work

src/extension.ts
radeksimko: 👍🏻 Thanks for the explanation

PR #1133, opened 2022-06-03
Remove package.json from .prettierignore

jpogran:
This removes package.json and syntaxes from the .prettierignore file.

We previously wanted package.json to not be formatted because of the addition of the semantic token data, but keeping it formatted is more important that the extra lines that adds.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,7 +1,6 @@
 snippets/
 syntaxes/
 out/
-package.json
 package-lock.json
 node_modules/
 .vscode-test/


radeksimko (commented):
Doesn't this leave us in a situation where whenever we run `npm install`, `npm version` etc we end up with a differently formatted file than when editing `package.json` manually (because prettier is more likely to be installed and triggered on save in IDE)? I'm okay with either formatting, but if we prefer one, it should be obvious which one it is - e.g. could we run prettier in the CI?

Alternatively/additionally - is there any way we could align prettier with `npm` at all?

jpogran:
After discussing with @radeksimko, we've concluded that `npm version --no-git-tag-version <ver>` is the thing that is changing formatting for him.

In https://github.com/hashicorp/vscode-terraform/commit/6e391cd8f3c25de448dd9337956b5fdd614dcfd6, we added `package.json` to the `.prettierignore` file because of hand written information introduced in https://github.com/hashicorp/vscode-terraform/commit/036593c993feb716a452ff9575829c49fbb042f4. We thought the hand written stanzas were visually better than the formatted version. Unfortunately `npm version` will rewrite the package.json file in order to update the package version. This was run in https://github.com/hashicorp/vscode-terraform/commit/0ac44cd22e86365f8352a6ad768cf61e383548aa, which undid the custom formatting.

Going forward we'll keep prettier formatting package.json as it maintains a format that npm version recognizes as correct.

radeksimko:
> is there any way we could align prettier with npm at all?

And to answer my own question - anecdotally it seems that they actually agree on the formatting.

radeksimko (approved):

PR #1156, opened 2022-06-22
Refactor Extension Settings

jpogran:
Migrate all settings to conform to common VS Code setting naming conventions. This improves readability in the VS Code Settings UI as well as making the settings easier to discover when typing in the JSON Settings view.

The `terraform.languageServer` setting block has been extracted out to individual settings. In addition, `terraform.languageServer.external` has been renamed to `terraform.languageServer.enable`

<table>
<tr><td>Old</td><td>New</td></tr>
<tr>
<td>

```json
"terraform.languageServer": {
  "external": true,
  "pathToBinary": "",
  "args": [ "serve" ],
  "ignoreSingleFileWarning": false
},
```
</td>
<td>

```json
"terraform.languageServer.enable": true,
"terraform.languageServer.pathToBinary": "",
"terraform.languageServer.args": [ "serve" ],
"terraform.languageServer.ignoreSingleFileWarning": false
```

</td>
</tr>
</table>

The `terraform-ls.terraformExec` settings have been moved to the `terraform` section and have been renamed:

<table>
<tr><td>Old</td><td>New</td></tr>
<tr>
<td>

```json
"terraform-ls.terraformExecPath": "",
"terraform-ls.terraformExecTimeout": "",
"terraform-ls.terraformExecLogFilePath": ""
```

</td>
<td>

```json
"terraform.languageServer.terraform.path": "",
"terraform.languageServer.terraform.timeout": "",
"terraform.languageServer.terraform.logFilePath": ""
```

</td>
</tr>
</table>

All settings previously under the `terraform-ls` section have been moved to the `terraform` section:

<table>
<tr><td>Old</td><td>New</td></tr>
<tr>
<td>

```json
"terraform-ls.rootModules": [],
"terraform-ls.excludeRootModules": [],
"terraform-ls.ignoreDirectoryNames": []
```

</td>
<td>

```json
"terraform.languageServer.rootModules": [],
"terraform.languageServer.excludeRootModules": [],
"terraform.languageServer.ignoreDirectoryNames": []
```

</td>
</tr>
</table>
The `terraform.experimentalFeatures` setting block has been extracted out to individual settings:

<table>
<tr><td>Old</td><td>New</td></tr>
<tr>
<td>

```json
"terraform.experimentalFeatures": {
  "validateOnSave": false,
  "prefillRequiredFields": false,
},
```
</td>
<td>
    
```json
"terraform.experimentalFeatures.validateOnSave": false,
"terraform.experimentalFeatures.prefillRequiredFields": false,
```

</td>
</tr>
</table>

Miscellaneous Setting Changes:

- Simplified enable/disable language server commands due to extracted `terraform.languageServer.enable` setting. No longer need to read the current value of the entire `terraform.languageServer` block.
- Updated `terraform.languageServer.terraform.logFilePath` to use `markdownDescription` to increase readability.
- Updated `terraform.languageServer.ignoreSingleFileWarning` description for better readability in Settings UI
- Added descriptions to `terraform.experimentalFeatures.prefillRequiredFields` and `terraform.experimentalFeatures.validateOnSave` since they are now first level settings.
- Added item type to `terraform.languageServer.args` for better UX in Settings UI.

Final UI:

image

image


package.json
```diff
@@ -285,46 +285,40 @@
...
-                "scope": "resource",
-                "type": "array",
-                "default": [
-                  "serve"
-                ],
-                "description": "Arguments to pass to language server binary"
-              }
+            "type": "boolean",
+            "default": true,
+            "description": "Enable/disable Terraform Language Server"
```
radeksimko, line 293: Is there a particular reason behind the "disable" in the description? I mean if it's a tickbox/boolean I'd expect it to express what the ticked/`true` value means.

Screenshot 2022-06-23 at 12 35 53
Screenshot 2022-06-23 at 12 37 08

src/extension.ts
```diff
@@ -383,22 +367,70 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+  }
+
+  switch (choice.title) {
+    case "Don't remind me again":
+      ctx.globalState.update('terraform.disableSettingsMigration', true);
+      return;
+    case 'Migrate':
+    // migrate below
+  }
+
```
radeksimko, line 414: The migration UI looks neat, but it didn't seem to do anything for me once confirmed

Screenshot 2022-06-23 at 12 33 42

Screenshot 2022-06-23 at 14 59 00

Not sure if that has to do anything with the dev mode - do I need to package a vsix and install it properly for this to work?

src/extension.ts
```diff
@@ -383,22 +367,70 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+  await migrate('terraform', 'languageServer.args', 'languageServer.args');
+  await migrate('terraform', 'languageServer.ignoreSingleFileWarning', 'languageServer.ignoreSingleFileWarning');
+
+  await migrate('terraform-ls', 'rootModules', 'languageServer.rootModules');
+  await migrate('terraform-ls', 'excludeRootModules', 'languageServer.excludeRootModules');
+  await migrate('terraform-ls', 'ignoreDirectoryNames', 'languageServer.ignoreDirectoryNames');
+
+  await migrate('terraform-ls', 'terraformExecPath', 'languageServer.terraform.executable.path');
+  await migrate('terraform-ls', 'terraformExecTimeout', 'languageServer.terraform.executable.timeout');
+  await migrate('terraform-ls', 'terraformLogFilePath', 'languageServer.terraform.executable.logFilePath');
```
radeksimko, line 426: Maybe I'm misunderstand the code, but doesn't this migration keep the options under `terraform-ls`, which we wanted to get rid of?

src/extension.ts
```diff
@@ -383,22 +367,70 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+
+  await migrate('terraform-ls', 'experimentalFeatures.validateOnSave', 'experimentalFeatures.validateOnSave');
+  await migrate(
+    'terraform-ls',
+    'experimentalFeatures.prefillRequiredFields',
+    'experimentalFeatures.prefillRequiredFields',
+  );
```
radeksimko, line 433: Similar question to above - isn't this no-op?

src/utils/clientHelpers.ts
```diff
@@ -21,22 +21,22 @@ export function getInitializationOptions() {
...
-  const ignoreDirectoryNames = config('terraform-ls').get<string[]>('ignoreDirectoryNames', []);
-
+  const terraformExecPath = config('terraform').get<string>('languageServer.terraform.executable.path', '');
+  const terraformExecTimeout = config('terraform').get<string>('languageServer.terraform.executable.timeout', '');
+  const terraformLogFilePath = config('terraform').get<string>('languageServer.terraform.executable.logFilePath', '');
+  const ignoreDirectoryNames = config('terraform').get<string[]>('languageServer.ignoreDirectoryNames', []);
   const ignoreSingleFileWarning = config('terraform').get<boolean>('languageServer.ignoreSingleFileWarning', false);
+  const experimentalFeatures = config('terraform').get('experimentalFeatures');
 
+  // deprecated
```
radeksimko, line 31: I think we have not deprecated these options _yet_?

src/utils/vscode.ts
```diff
@@ -35,18 +28,61 @@ export function getScope(section: string, settingName: string): vscode.Configura
   return scoppe;
 }
 
-export async function updateConfig<T>(
+export async function warnIfMigrate(settings: string[]): Promise<boolean> {
+  for (let index = 0; index < settings.length; index++) {
+    const setting = settings[index];
+    const section = setting.split('.')[0];
+    const settingName = setting.replace(section + '.', '');
```
radeksimko, line 35: Nitpick: I know this code will likely have a relatively short timespan, but is it worth making this a little bit cleaner? e.g. by making the argument something like `settings: OptionAddress[]` where `OptionAddress` is

```ts
interface ConfigOption {
  section: string;
  name: string;
}
```

radeksimko (commented):
What is intended/expected to happen when the user refuses the settings migration? Would it just stop working? A quick look at the implementation would suggest it stops working, but I may be misreading the code? 🤔

README.md
```diff
@@ -292,19 +288,19 @@ If you want to automatically ignore certain directories when terraform-ls indexe
...
 ```json
-"terraform-ls.terraformExecTimeout": "30"
+"terraform.languageServer.terraform.executable.timeout": "30"
 ```
 
 You can set the path Terraform logs (`TF_LOG_PATH`) by configuring this setting:
 
 ```json
-"terraform-ls.terraformLogFilePath": "C:/some/folder/path/log-{{varname}}.log"
+"terraform.languageServer.terraform.executable.logFilePath": "C:/some/folder/path/log-{{varname}}.log"
```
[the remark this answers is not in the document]
jpogran: We went back and forth with lots of variations and left it with this naming to be explicit that this was how to configure how the CLI behaves and where it was physically located. 

Removing `executable` results in `terraform.languageServer.terraform.path` and `Language Server > Terraform > Path`. My concern is that `terraform.languageServer.terraform.path` and `terraform.languageServer.terraform.logFilePath` do not have much visually to disambiguate each other, but I may be over thinking it.

However, if we can conceivably add more options here, then it would make sense to remove `executable` as I think this reads nicely:

```
terraform.languageServer.terraform.path
terraform.languageServer.terraform.timeout
terraform.languageServer.terraform.plan.parallelism
terraform.languageServer.terraform.apply.autoApprove
```

This allows segments for each subcommand and room for growth without squishing everything together like `terraform.languageServer.terraform.planParallelism`.

package.json
jpogran: I don't know what other wording to use here.

src/extension.ts
```diff
@@ -383,22 +367,70 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+      ctx.globalState.update('terraform.disableSettingsMigration', true);
+      return;
+    case 'Migrate':
+    // migrate below
+  }
+
+  await migrate('terraform', 'languageServer.external', 'languageServer.enable');
+  await migrate('terraform', 'languageServer.pathToBinary', 'languageServer.path');
+  await migrate('terraform', 'languageServer.args', 'languageServer.args');
+  await migrate('terraform', 'languageServer.ignoreSingleFileWarning', 'languageServer.ignoreSingleFileWarning');
```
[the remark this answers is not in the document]
jpogran: The `migrate` function detects if the setting is present at any of the scopes we support, then writes it out with the current value. If it's not defined, then it does nothing and returns.

Previously `terraform.languageServer` was written as a JSON object. We are extracting this out into separate settings. If a user had the block written out with any or all of the settings defined, we have to migrate it otherwise the user is left with different settings in different formats. The older object based notation will be grayed out and inactivate, but still present with the new format side by side.

For example, a user used autocomplete in the JSON file and had the following populated thanks to our `default` stanza on line 313:

```
"terraform.languageServer": {
  "external": true,
  "pathToBinary": "",
  "args": [
    "serve"
  ],
  "ignoreSingleFileWarning": false
}
```

They then edited the file like so (with made up argument values):

```
"terraform.languageServer": {
  "external": true,
  "args": ["serve", "--foo"],
}
```

If we did not account for `args`, ran the migration, then the settings would look like:

```
"terraform.languageServer": {
  "external": true,
  "args": ["serve", "--foo"],
}
"terraform.languageServer.enable": true,
```

Our extension would pick up `terraform.languageServer.enable` correctly, but miss `terraform.languageServer.args` and the user would not know why. The terraform-ls binary would be started with 'server' and miss '--foo'.

src/extension.ts
jpogran: The `migrate()` function has `const targetSection = section === 'terraform-ls' ? 'terraform' : section;` inside it which will purposely transform `terraform-ls` to `terraform` if it's provided.

This allows me to do `await migrate('terraform-ls', 'rootModules', 'languageServer.rootModules');` and have it automatically move to `terraform` without having to specify `terraform` in the function call.

I could have made the function declaration `migrate('terraform-ls','terraform', 'rootModules', 'languageServer.rootModules')`, but I felt that that was a lot of untyped strings with no checked order to get right. I could have made it with typed arguments like `migrate(ExtensionSection.TerraformLs,ExtensionSection.Terraform, 'rootModules', 'languageServer.rootModules')` or an object or something, but I thought that was too much for a small temporary part of our extension.

src/extension.ts
jpogran: Same answer as above

src/extension.ts
jpogran: No, you should be able to do this in the extension host in dev mode. Was that example in user, folder or workspace settings?

src/utils/clientHelpers.ts
jpogran: They are listed as deprecated in package.json

src/utils/vscode.ts
jpogran: Let me be lazy! 😛 

Adjusted as suggested

src/utils/clientHelpers.ts
radeksimko: Oh right, just have not done that yet on the LS side then. 😅

src/extension.ts
radeksimko: Oh I see, the `migrate()` function assumes the target section. I did not realise that 😄 

I think that the more explicit (yet verbose) interface would be nicer, but I'm also fine with this.

src/extension.ts
radeksimko: This is user, i.e. JSON file in `~/Library/Application Support/Code/User/settings.json`
I'm happy to jump on Zoom to look into this further - it may be something on my side.

src/extension.ts
```diff
@@ -383,22 +367,90 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+  }
+
+  switch (choice.title) {
+    case "Don't remind me again":
+      ctx.globalState.update('terraform.disableSettingsMigration', true);
+      return;
+    case 'More Info':
+      await vscode.commands.executeCommand(
+        'vscode.open',
+        vscode.Uri.parse('https://www.github.com/hashicorp/vscode-terraform#settings-migration'),
```
radeksimko, line 424: This is very useful! The only downside I noticed is that given that it's a button in multiple-choice dialog, it means that the dialog disappears, so if - after reading the details - user decides to do the migration, they have to reload the IDE to bring the dialog back.

Is there any way we could keep the dialog in place in this context? e.g. could the question in the dialog be markdown with that link? 🤔

README.md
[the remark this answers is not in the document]
jpogran: Removed `executable`

package.json
jpogran: Dropped disable

src/extension.ts
jpogran: > This is very useful! The only downside I noticed is that given that it's a button in multiple-choice dialog, it means that the dialog disappears, so if - after reading the details - user decides to do the migration, they have to reload the IDE to bring the dialog back.

Unfortunately, yes, the modal window disappears when a button is clicked. However the window will come back every time the user starts VS Code until all settings we define in lines 392-402 are changed. We could implement a timer like we had with the LS updater to prompt the user, but that seems like it would upset the user more than help.

> Is there any way we could keep the dialog in place in this context? e.g. could the question in the dialog be markdown with that link? 🤔

With a dialog, not that I know of. These message windows have to be cross platform, so there are limited things you can do.

README.md
```diff
@@ -340,6 +334,109 @@ The configuration has changed from 1.4.0 to v2.X. If you are having issues with
 - Contributions are always welcome! Please see our [contributing guide](https://github.com/hashicorp/vscode-terraform/issues/new?assignees=&labels=enhancement&template=feature_request.md) for more details
 - If you're interested in the development of the extension, you can read about our [development process](DEVELOPMENT.md)
 
+### Settings Migration
+
+In v0.24.0 we migrated all settings to conform to common VS Code setting naming conventions. This improves readability in the VS Code Settings UI as well as making the settings easier to discover when typing in the JSON Settings view.
+
+After v0.24.0, if there are old settings left in the User or Workspace scopes, this extension will show a prompt asking to migrate the settings. The user can either automatically migrate, ignore future warnings, open the Settings UI to manually migrate, or click a link for more information.
```
radeksimko, line 341:
```suggestion
In v2.24.0 we migrated all settings to conform to common VS Code setting naming conventions. This improves readability in the VS Code Settings UI as well as making the settings easier to discover when typing in the JSON Settings view.

After v2.24.0, if there are old settings left in the User or Workspace scopes, this extension will show a prompt asking to migrate the settings. The user can either automatically migrate, ignore future warnings, open the Settings UI to manually migrate, or click a link for more information.
```

src/extension.ts
```diff
@@ -383,22 +375,99 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
-        'languageServer',
-        { enabled: undefined, external: true },
-        vscode.ConfigurationTarget.Global,
+async function migrateLegacySettings(ctx: vscode.ExtensionContext) {
+  // User has asked not to check if settings need to be migrated, so return
+  if (ctx.globalState.get('terraform.disableSettingsMigration', false)) {
+    return;
+  }
+
+  // TODO: Short circuit here so we don't check for every activation forever
```
radeksimko, line 387: 👀 I think this was already implemented above, wasn't it?

src/extension.ts
```diff
@@ -383,22 +375,99 @@ async function terraformCommand(command: string, languageServerExec = true): Pro
...
+  const warnMigration = warnIfMigrate([
+    { section: 'terraform', name: 'languageServer.external' },
+    { section: 'terraform', name: 'languageServer.pathToBinary' },
+    { section: 'terraform-ls', name: 'rootModules' },
+    { section: 'terraform-ls', name: 'excludeRootModules' },
+    { section: 'terraform-ls', name: 'ignoreDirectoryNames' },
+    { section: 'terraform-ls', name: 'terraformExecPath' },
+    { section: 'terraform-ls', name: 'terraformExecTimeout' },
+    { section: 'terraform-ls', name: 'terraformLogFilePath' },
+    { section: 'terraform-ls', name: 'experimentalFeatures' },
```
radeksimko, line 401: This looks so much cleaner & easier to read with the interface! 😍

radeksimko (commented):
This looks pretty good. I left some comments/questions in-line, but most are minor.

One question which came to my mind while testing this: I see that old settings will be ignored, which is fine from _functional_ perspective (i.e. I agree we should be ignoring them), but I think we should make it clear that we do that in the migration docs.

On a related note, I wonder if it's worth keeping the old setting options in place as deprecated, so that if user ignores them, they see the warnings in both the UI and JSON? The UI doesn't show deprecated fields there unless they're set in the config - so it wouldn't even upset the UI of users who migrated.

For example:

Screenshot 2022-06-28 at 11 56 16

Screenshot 2022-06-28 at 11 56 09
Screenshot 2022-06-28 at 11 56 09

README.md
```diff
@@ -340,6 +334,109 @@ The configuration has changed from 1.4.0 to v2.X. If you are having issues with
 - Contributions are always welcome! Please see our [contributing guide](https://github.com/hashicorp/vscode-terraform/issues/new?assignees=&labels=enhancement&template=feature_request.md) for more details
 - If you're interested in the development of the extension, you can read about our [development process](DEVELOPMENT.md)
 
+### Settings Migration
```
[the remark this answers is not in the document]
jpogran: Moved!

src/extension.ts
radeksimko: Oh I see - I misunderstood your intentions. Agreed that it's not worth caching though. So I'd just remove the comment then.

README.md
```diff
@@ -340,6 +334,10 @@ The configuration has changed from 1.4.0 to v2.X. If you are having issues with
...
+
+
```
radeksimko, line 339:
```suggestion
Read more about [changes in settings options introduced in v2.24.0](./docs/settings-migration.md).
```

radeksimko (approved):
LGTM, migration works well for me. The UX looks decent with the right bottom message window, and I like the ordering of buttons, as discussed. 👍🏻 
shipit

package.json
```diff
@@ -334,20 +328,52 @@
...
+          },
+          "terraform-ls.rootModules": {
+            "scope": "resource",
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "default": [],
+            "description": "Per-workspace list of module directories for the language server to read",
+            "deprecationMessage": "Moved: Use `#terraform.languageServer.rootModules#` instead"
```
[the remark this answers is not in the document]
jpogran: It would show a link to click if i had used the markdown version of it. Fixing!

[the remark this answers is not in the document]
radeksimko: It still looks the same, but it may be just my local environment - I couldn't get the markdown to work before when toying with deprecation messages at all.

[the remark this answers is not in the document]
jpogran: What version are you on?

Insiders:

image

Stable:

image

PR #1164, opened 2022-06-29
Set proper scope for machine based settings

jpogran:
This sets the following settings to `machine-overridable` scope. This ensures these settings will be read in the correct location whether the user is using the extension locally or in a remote context like the Remote Extensions or Codespaces.

It is set to `machine-overridable` instead of just `machine`, to still allow users to have these settings set inside a Workspace file or in the Workspace folder. If we set it to just `machine`, the settings would not be modifiable in the Workspace file or in the Workspace folder scopes.

```
terraform.languageServer.path
terraform.languageServer.args
terraform.languageServer.terraform.path
terraform.languageServer.terraform.timeout
terraform.languageServer.terraform.logFilePath
terraform.languageServer.rootModules
terraform.languageServer.excludeRootModules
terraform.languageServer.ignoreDirectoryNames
```

For all other settings, this sets the scope to the default `window`. This removes the error messages being logged in the extension host about accessing a resource scoped setting using a non resource scoped method:

> [2022-06-29 12:40:41.683] [exthost] [warning] [hashicorp.terraform] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'terraform.languageServer.enable', provide the URI of a resource or 'null' for any resource.

### DIFF
--- a/package.json
+++ b/package.json
@@ -274,7 +274,7 @@
         "order": 0,
         "properties": {
           "terraform.codelens.referenceCount": {
-            "scope": "resource",
+            "scope": "window",
             "type": "boolean",
             "default": false,
             "description": "Display reference counts above top level blocks and attributes."
@@ -286,7 +286,7 @@
         "order": 1,
         "properties": {
           "terraform.languageServer.enable": {
-            "scope": "resource",
+            "scope": "window",
             "order": "0",
             "type": "boolean",
             "default": true,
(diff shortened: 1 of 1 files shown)


dbanck (approved):
Looks like a sensible choice of scopes to me. We could remove `"scope": "window", since it's the default. But I'm also okay with being explicit here.

PR #1171, opened 2022-07-05
Terraform View container

jpogran:
Create a new Terraform [View Container](https://code.visualstudio.com/api/ux-guidelines/views#view-containers) that will house the existing `terraform.providers` and `terraform.modules` views. These views will no longer be inside the Explorer View Container.

image

This Terraform View Container will support all the built-in mechanisms users expect out of views, most notable of which is the ability to move it to other places in the editor like the new SideBar:

image

Or be able to be hidden altogether:

image

This enables us to have a single place to land new views as well as allow the user to customize as they see fit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -552,20 +552,23 @@
       ]
     },
     "views": {
-      "explorer": [
+      "terraform": [
         {
           "id": "terraform.providers",
-          "name": "Terraform Providers",
-          "icon": "assets/icons/terraform.svg",
-          "visibility": "collapsed",
-          "when": "terraform.showTreeViews"
+          "name": "Terraform Providers"
         },
         {
           "id": "terraform.modules",
-          "name": "Terraform Module Calls",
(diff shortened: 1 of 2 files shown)


dbanck (approved):
LGTM 👍 

This introduces a small UX change. We now always show the panes, even if the user has the language server disabled. Should we evaluate hiding the whole view container if there is no LS?

radeksimko (approved):
Thanks for the update. This gives us (and the user) more flexibility for now and the future.

Agreed with @dbanck that it would be neat if we could hide the container entirely dynamically, like before, but I'm not sure if that's even possible with the VS Code API.

That said as everything is now more subtly hidden under the icon, it may be less of a problem as user would have to click the icon first and even when they do, we already provide some sensible messages:
Screenshot 2022-07-08 at 12 00 25

dbanck:
Thanks for the explanation. Let's keep it that way

:shipit:

PR #1185, opened 2022-07-18
Refactor Terraform Execution API

jpogran:
Adds new button to the providers view which calls `terraform init current `:

image

Introduces terraform-ls error response handling:

image

Introduce progress reporting to all Terraform commands:

image

- Remove terraform init status bar
- Move terraform-ls dependent command registration
- Refactor terraform execution
- Extract Terraform commands
- Consolidate all terraform executions
- Add progress notification
- Rename commands for fluent api
- Register commands before starting LS
- Improve open dialog title


src/providers/moduleProviders.ts
```diff
@@ -1,25 +1,11 @@
...
 import * as vscode from 'vscode';
 import { Utils } from 'vscode-uri';
-import { ExecuteCommandParams, ExecuteCommandRequest } from 'vscode-languageclient';
-
 import { getActiveTextEditor, isTerraformFile } from '../utils/vscode';
 import { LanguageClient } from 'vscode-languageclient/node';
-import { clientSupportsCommand } from '../utils/clientHelpers';
+import TelemetryReporter from '@vscode/extension-telemetry';
 
 /* eslint-disable @typescript-eslint/naming-convention */
```
dbanck, line 8:
```suggestion
```

src/providers/moduleProviders.ts
```diff
@@ -1,25 +1,11 @@
...
-      display_name: string;
-      version_constraint?: string;
-      docs_link?: string;
-    };
-  };
-  installed_providers: {
-    [provider: string]: string;
-  };
-}
 /* eslint-enable @typescript-eslint/naming-convention */
```
dbanck, line 9:
```suggestion
```

src/terraform.ts
```diff
@@ -0,0 +1,228 @@
...
+import { clientSupportsCommand } from './utils/clientHelpers';
+
+/* eslint-disable @typescript-eslint/naming-convention */
+export interface ModuleCaller {
+  uri: string;
+}
+
+export interface ModuleCallersResponse {
+  version: number;
+  moduleCallers: ModuleCaller[];
```
dbanck, line 15: While working on the mocha stuff yesterday, I noticed that the fields in the response might have the wrong names (also wrong in previous versions). Shouldn't they be called `v` and `callers` or am I missing something?

https://github.com/hashicorp/terraform-ls/blob/4e96ebe79a350c56774cb1ee1f18924ee6563f1a/internal/langserver/handlers/command/module_callers.go#L15-L17

src/terraform.ts
```diff
@@ -0,0 +1,228 @@
...
+      return;
+    }
+
+    const moduleUri = selected[0];
+    const command = `terraform-ls.terraform.init`;
+
+    return execWorkspaceLSCommand<void>(command, moduleUri.toString(), client, reporter);
+  } catch (error) {
+    if (error instanceof Error) {
+      vscode.window.showErrorMessage(error instanceof Error ? error.message : error);
```
dbanck, line 104:
```suggestion
      vscode.window.showErrorMessage(error.message);
```

We checked if `error` is instance of `Error` in the if clause, so it's safe to access `.message` without the extra check here.

src/terraform.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+}
+
+export async function initCommand(client: LanguageClient, reporter: TelemetryReporter) {
+  try {
+    await terraformCommand('init', client, reporter);
+  } catch (error) {
+    if (error instanceof Error) {
+      vscode.window.showErrorMessage(error instanceof Error ? error.message : error);
```
dbanck, line 116:
```suggestion
      vscode.window.showErrorMessage(error.message);
```

src/terraform.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+}
+
+export async function command(command: string, client: LanguageClient, reporter: TelemetryReporter, useShell = false) {
+  try {
+    await terraformCommand(command, client, reporter, useShell);
+  } catch (error) {
+    if (error instanceof Error) {
+      vscode.window.showErrorMessage(error instanceof Error ? error.message : error);
```
dbanck, line 128:
```suggestion
      vscode.window.showErrorMessage(error.message);
```

dbanck (changes requested):
Great work! Looks much cleaner now!

I left a few minor suggestions.

src/terraform.ts
jpogran: I think you're right. I updated the interface

dbanck (approved):
Thanks for the updates! LGTM 👍 

We could add some unit tests for `terraform.ts` or do this later in a separate PR.

PR #1187, opened 2022-07-22
Set machine scope to path settings

jpogran:
This scopes all path related and dependent settings to machine scope. This restricts the settings to only be specified in the User or Remote settings, not workspace or folder settings.

This allows a setting to have different values depending on where VS Code is being run in. For example, if a user runs this extension in Windows locally and in an Ubuntu WSL instance they need to specify two different paths to terraform-ls (if they use a custom install location). They configure a Windows path in their User settings and configure a Linux path in their Remote WSL session. The correct path will be resolved in either session.

### DIFF
--- a/docs/settings-migration.md
+++ b/docs/settings-migration.md
@@ -121,3 +121,23 @@ The `terraform.experimentalFeatures` setting block has been extracted out to ind
 </table>
 
 For an easy to view table of settings, see the `Contributions` tab when viewing the Terraform Extension in the VS Code Extension Pane.
+
+## Setting Scope
+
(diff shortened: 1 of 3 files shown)


docs/settings-migration.md
```diff
@@ -121,3 +121,23 @@ The `terraform.experimentalFeatures` setting block has been extracted out to ind
 </table>
 
 For an easy to view table of settings, see the `Contributions` tab when viewing the Terraform Extension in the VS Code Extension Pane.
+
+## Setting Scope
```
radeksimko, line 125: Nitpick: Just for extra clarity - could we also mention that this basically prevents the options below to be loaded from workspace/directory-level settings?

radeksimko (approved):

docs/settings-migration.md
jpogran: Good idea. Amended

radeksimko (approved):

dbanck (approved):
👍

PR #1190, opened 2022-07-26
Add Terraform version info

jpogran:
This commit adds a LanguageStatusItem that shows the discovered Terraform version in the current workspace. This utilizes the LSP request mechanism to automatically show the version when it is discovered for the currently open document.

This requires an open document and does not attempt to query for a folder because LanguageStatusItems only show for open documents. 

Requires https://github.com/hashicorp/terraform-ls/pull/1016

### DIFF
--- a/src/features/terraformVersion.ts
+++ b/src/features/terraformVersion.ts
@@ -1,0 +1,46 @@
+import * as terraform from '../terraform';
+import * as vscode from 'vscode';
+import { ClientCapabilities, ServerCapabilities, StaticFeature } from 'vscode-languageclient';
+import { ExperimentalClientCapabilities } from './types';
+import TelemetryReporter from '@vscode/extension-telemetry';
+import { LanguageClient } from 'vscode-languageclient/node';
+import { getActiveTextEditor } from '../utils/vscode';
+
+export const CLIENT_TERRAFORM_VERSION_CMD_ID = 'client.refreshTerraformVersion';
+
+export class TerraformVersionFeature implements StaticFeature {
+  private disposables: vscode.Disposable[] = [];
+
+  constructor(private client: LanguageClient, private reporter: TelemetryReporter) {}
+
+  public fillClientCapabilities(capabilities: ClientCapabilities & ExperimentalClientCapabilities): void {
+    if (!capabilities['experimental']) {
+      capabilities['experimental'] = {};
+    }
+    capabilities['experimental']['refreshTerraformVersionCommandId'] = CLIENT_TERRAFORM_VERSION_CMD_ID;
+  }
+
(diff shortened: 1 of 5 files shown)


src/terraform.ts
```diff
@@ -1,6 +1,12 @@
+import * as tfStatus from './status/terraform';
 import TelemetryReporter from '@vscode/extension-telemetry';
 import * as vscode from 'vscode';
-import { ExecuteCommandParams, ExecuteCommandRequest, LanguageClient } from 'vscode-languageclient/node';
+import {
+  ExecuteCommandParams,
+  ExecuteCommandRequest,
+  LanguageClient,
+  WorkDoneProgress,
```
dbanck, line 8:
```suggestion
```

src/features/terraformVersion.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+
+      terraform.getTerraformVersion(editor.document.uri, this.client, this.reporter);
+    });
+
+    this.disposables.push(d);
+  }
+
+  public dispose(): void {
+    this.disposables.forEach((d: vscode.Disposable) => d.dispose());
```
dbanck, line 44: Do we need to remove the disposable after disposing it?

src/status/terraform.ts
```diff
@@ -0,0 +1,27 @@
+import * as vscode from 'vscode';
+
+const terraformStatus = vscode.languages.createLanguageStatusItem('terraform.status', [
+  { language: 'terraform' },
+  { language: 'terraform-vars' },
+]);
+terraformStatus.name = 'Terraform';
+terraformStatus.detail = 'Terraform';
+terraformStatus.command = {
+  command: 'terraform.commands',
```
dbanck, line 10: What does this command do? Do we need to register it?

src/terraform.ts
```diff
@@ -44,6 +56,39 @@ interface ModuleProvidersResponse {
...
+export async function getTerraformVersion(
+  moduleUri: vscode.Uri,
+  client: LanguageClient,
+  reporter: TelemetryReporter,
+): Promise<void> {
+  try {
+    const moduleDir = Utils.dirname(moduleUri);
+
+    const response = await terraformVersion(moduleDir.toString(), client, reporter);
+    tfStatus.setTerraformVersion(response.discovered_version);
```
dbanck, line 68: I'm not sure if this is the place where I would expect the modification of the status bar item. Most of the other functions in this file make an API call and return the result.

Maybe this line belongs in the `TerraformVersionFeature`?

radeksimko:
This was implemented as part of https://github.com/hashicorp/vscode-terraform/pull/1325

PR #1192, opened 2022-07-26
Add terraform-ls LanguageStatusItem

jpogran:
This adds a new LanguageStatusItem for terraform-ls. It records the running state of the terraform-ls process and when the server is in an error state, it will show a visible error condition. When terraform-ls is starting, a busy indicator will be shown.

This LanguageStatusItem can be pinned by the user and will show the current terraform-ls version.

This also adds a new command to list terraform-ls specific commands in a quick-pick.


jpogran:
Superseded by https://github.com/hashicorp/vscode-terraform/pull/1547

PR #1200, opened 2022-08-04
Refactor main entry point

jpogran:
## Goals

- Centralize and standardize command registration and enablement
- Centralize and standardize feature registration and enablement
- Extract functionality to dedicated files
- Eliminate cyclic imports
- Dispose disposables properly

## Implementation

- Extract settings.ts

This moves all setting related functions to a new file called settings.ts. This is in preparation for a more formal settings infrastructure.

- Remove enabled function

This prefers direct detection of the setting rather than the function. This is in preparation for a more formal settings management in the future.

- Extract Terraform command registration

This extracts registering Terraform commands to a central class. This cleans up extension.ts from having lots of registrations, enables a central place to keep track of terraform-ls execution related commands, and enables centralized progress reporting in a future commit.

- Extract view providers registration

This commit centralizes the registration of the view providers inside the StaticFeature classes. This simplifies the creation of these elements, centralizes the many moving parts, and ensures all disposable components are disposed properly.

- Extract StaticFeature registrations

This simplifies registering all the custom StaticFeatures by moving the enabling logic to inside each class. This moves the decision logic closest to the point of use instead of a place separate from the class. It also follows the API intention of using the initialize function to determine if the feature is initialized or not.

This also ensures the telemetry handler is disposed as the class is disposed, to ensure the LanguageClient is closed properly.

- Cleanup deactivation

- Extract error handler

This extracts the error handling logic into a dedicated class instead of having the entire set inside the extension.ts file.

- Resolve cyclic dependency on outputChannel

This removes the export on outputChannel and makes it a function parameter to prevent extension.ts exporting it while also importing clientHelpers.ts, which imports from extension.ts.

### DIFF
--- a/src/features/showReferences.ts
+++ b/src/features/showReferences.ts
@@ -40,7 +40,7 @@ export class ShowReferencesFeature implements StaticFeature {
       return;
     }
 
-    const codeLensReferenceCount = config('terraform').get<boolean>('codelens.referenceCount');
+    const codeLensReferenceCount = config('terraform').get<boolean>('codelens.referenceCount', false);
     if (codeLensReferenceCount === false) {
       return;
     }
--- a/src/features/showReferences.ts
+++ b/src/features/showReferences.ts
@@ -7,6 +7,7 @@ import {
   ServerCapabilities,
   StaticFeature,
 } from 'vscode-languageclient';
+import { config } from '../utils/vscode';
 
 import { ExperimentalClientCapabilities } from './types';
 
@@ -36,6 +37,11 @@ export class ShowReferencesFeature implements StaticFeature {
 
   public initialize(capabilities: ServerCapabilities): void {
     if (!capabilities.experimental?.referenceCountCodeLens) {
+      return;
+    }
+
+    const codeLensReferenceCount = config('terraform').get<boolean>('codelens.referenceCount');
+    if (codeLensReferenceCount === false) {
       return;
     }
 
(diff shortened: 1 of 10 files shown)


src/features/showReferences.ts
```diff
@@ -39,6 +40,11 @@ export class ShowReferencesFeature implements StaticFeature {
       return;
     }
 
+    const codeLensReferenceCount = config('terraform').get<boolean>('codelens.referenceCount');
```
dbanck, line 43: This one should be opt-in and not enabled by default, right? So we have to use a default value of `false` here or update the `if` next line to a more "falsy" check

src/handlers/errorHandler.ts
```diff
@@ -0,0 +1,51 @@
+import * as vscode from 'vscode';
+import { CloseAction, ErrorAction, ErrorHandler, Message } from 'vscode-languageclient';
+
+export class ExtensionErrorHandler implements ErrorHandler {
+  constructor(private outputChannel: vscode.OutputChannel) {}
+  error(error: Error, message: Message | undefined, count: number | undefined): ErrorAction {
```
dbanck, line 6:
```suggestion
  error(error: Error, message: Message | undefined, count: number | undefined) {
```
Fun thing with implementing interfaces is, that you have to type the function arguments, but you don't have to type the return type.

src/handlers/errorHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+    vscode.window.showErrorMessage(`Terraform LS connection error: (${count})\n${error.message}\n${message?.jsonrpc}`);
+
+    return ErrorAction.Continue;
+  }
+  closed(): CloseAction {
```
dbanck, line 11:
```suggestion
  closed() {
```

dbanck (approved):
I like the refactoring a lot 👍 

I've found a few minor things. Feel free to merge after addressing those

PR #1203, opened 2022-08-09
fix: avoid sending command ID when codelens is disabled

radeksimko:
The language server will enable reference count code lens when it receives the `experimental.showReferencesCommandId`, which we seem to send always, i.e. we always enable it. Consequently we (correctly) don't register the command, so upon clicking to the code lens, it errors out:

Screenshot 2022-08-09 at 12 48 51

This makes it so that the client only sends that command ID when the feature is actually enabled.


jpogran (approved):
Thanks for figuring this out!

PR #1210, opened 2022-08-22
Publish Web Extension

jpogran:
image


dbanck (approved):
👍 Nice work! Just a few minor questions

package.json
```diff
@@ -687,11 +693,21 @@
...
     "ts-jest": "^27.1.0",
+    "ts-loader": "^9.3.1",
     "ts-node": "^10.7.0",
     "typescript": "^4.6.2",
-    "vsce": "^2.6.6"
+    "vsce": "^2.10.0",
+    "webpack": "^5.74.0",
+    "webpack-cli": "^4.10.0"
+  },
+  "__metadata": {
```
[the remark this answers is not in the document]
jpogran: Not sure, it's been so long. I removed it and reran npm i and it didn't come back.

src/web/test/integration/extension.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  test('Sample test', () => {
+    assert.strictEqual(-1, [1, 2, 3].indexOf(5));
+    assert.strictEqual(-1, [1, 2, 3].indexOf(0));
+  });
+
+  test('Foo', () => {
+    assert.strictEqual(-1, [1, 2, 3].indexOf(5));
+    assert.strictEqual(-1, [1, 2, 3].indexOf(0));
+  });
```
radeksimko, line 19: Is there any benefit in keeping these "sample" tests around?

tsconfig.json
```diff
@@ -2,13 +2,13 @@
...
+    "lib": ["ES2020", "WebWorker"],
     "outDir": "out",
     "rootDir": "src",
     "sourceMap": true,
     "strict": true,
     "skipLibCheck": true
   },
-  "include": ["src"],
-  "exclude": ["node_modules", ".vscode-test"]
+  "include": ["src/**/*.ts", "./node_modules/vscode/vscode.d.ts", "./node_modules/vscode/lib/*"],
```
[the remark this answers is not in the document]
radeksimko: I don't have enough TypeScript knowledge to understand the impact, but I am also curious about the motivation behind this change, as well as the `exclude` below.

.github/workflows/test.yml
```diff
@@ -86,7 +87,9 @@ jobs:
         uses: hashicorp/setup-terraform@v2
         with:
           terraform_wrapper: false
-          terraform_version: '~1.0'
+          terraform_version: '~1'
```
radeksimko, line 90: What are the intentions and implications here?

I'm not very familiar with this constraint syntax `~1`. I know `go-version` recognises `~> 1.0` which means

> _Allows only the rightmost version component to increment._

https://developer.hashicorp.com/terraform/language/expressions/version-constraints#-3

Would that be equivalent to just `~` here, without the `>`? If so - then does that mean we allow upgrades to v2? That would likely be unwanted?

DEVELOPMENT.md
```diff
@@ -31,10 +31,36 @@ After cloning the repo, run `npm install` to install dependencies. There's an in
 
 ## Running the Extension
 
-The extension can be run in a development mode via the launch task called `Launch Client`. This will open a new VS Code window with the extension loaded, and from there you can open any files or folders you want to check against. This extension development window can also be used to run commands or any other feature the extension provides.
+The extension can be run in a development mode via the launch task called `Run Extension`. This will open a new VS Code window with the extension loaded, and from there you can open any files or folders you want to check against. This extension development window can also be used to run commands or any other feature the extension provides.
 
 > New to VS Code development? You can get started [here](https://code.visualstudio.com/api/get-started/your-first-extension)
 
+## Running the Web Extension
```
radeksimko, line 38: Thank you for adding these docs

package.json
```diff
@@ -643,18 +644,20 @@
...
-    "compile": "npm run esbuild",
-    "watch": "npm run download:artifacts && npm run esbuild-watch",
+    "prepare": "npm run download:artifacts",
+    "compile": "webpack",
+    "compile:test": "tsc -p ./",
+    "compile:prod": "webpack --mode production --devtool hidden-source-map",
+    "watch": "webpack --watch",
+    "web": "npm run compile && vscode-test-web --browserType=chromium --extensionDevelopmentPath=. ./testFixture",
+    "web:serve": "npx serve --cors -l 5000",
+    "web:tunnel": "npx localtunnel -p 5000",
```
radeksimko, line 654: I ran into a problem locally on macOS due to the Control Center already listening on port 5000 (which apparently is configured like that by default on a fresh macOS installation):
https://developer.apple.com/forums/thread/682332

Is it worth picking a higher port number, maybe just `5001`, to avoid the conflict?

radeksimko (approved):
Just a few minor questions in-line, but LGTM overall!

src/web/test/integration/extension.test.ts
jpogran: The reason it's a trivial test is that we don't actually have any functionality other than syntax highlighting to 'test'.

The main benefit to do this at all is that we have *something* here to run a unit test, so we can test that we can test the extension under the web host. If we add more functionality at a later date, the functionality is already present and we don't have to consult vscode docs to figure out how to add it, since it was already such a pain to adapt it to our current test harness.

I do think I should probably add this as a comment on the test, since I explained this to Daniel with our 1:1 pairing on this ticket.

radeksimko: I understand there's nothing to test, but I still don't understand under what scenario would the test fail, i.e. what can we catch here? Is it that `activate()` will not throw exceptions?

jpogran: We are testing that we can activate under the web host. We aren't testing any functionality at this poit.

src/test/integration/completion.test.ts
```diff
@@ -13,6 +13,7 @@ suite('completion', () => {
       new vscode.CompletionItem('data', vscode.CompletionItemKind.Class),
       new vscode.CompletionItem('locals', vscode.CompletionItemKind.Class),
       new vscode.CompletionItem('module', vscode.CompletionItemKind.Class),
+      new vscode.CompletionItem('moved', vscode.CompletionItemKind.Class),
```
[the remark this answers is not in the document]
jpogran: This was required several months ago, because the version of terraform in CI did not have the `moved' block we added as completable. I'll check if this is still the case now several months later.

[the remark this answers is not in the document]
radeksimko: I see - that makes sense!

tsconfig.json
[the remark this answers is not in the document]
jpogran: These lines came from the Microsoft [example repo for using webpack](https://github.com/microsoft/vscode-extension-samples/blob/fdd3bb95ce8e38ffe58fc9158797239fdf5017f1/webpack-sample/tsconfig.json#L11-L15), which I was using as as reference in trying to switch from esbuild to webpack as well as add the web extension.

There are other prominent extensions that do the same thing: https://github.com/prettier/prettier-vscode/blob/main/tsconfig.json#L14-L19, but I don't have a strong opinion on either way, and removing it seems to still allow compilation with webpack, and still allows all of the test suites to run so I'm fine with removing.

src/web/test/integration/extension.test.ts
jpogran: @radeksimko are you OK with leaving this as is since getting the testing framework required so much work, or do you want to remove the entire unit test setup and ticket adding it back at a later date?

package.json
jpogran: Changed!

src/test/integration/completion.test.ts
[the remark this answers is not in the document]
jpogran: So, if we leave terraform version in CI as `~1.0` and remove the `moved` completion here, everything passes in CI.

However, locally, if you have 1.1.X or greater, the tests will fail:

```

  code actions
    ✔ supported actions (3233ms)
  completion
    1) simple completion
  document symbols
    ✔ returns symbols (92ms)
  2 passing (3s)
  1 failing
  1) completion
       simple completion:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ 'moved'
- 'output'
      + expected - actual

      -moved
      +output
```

.github/workflows/test.yml
jpogran: See moved completion conversation for continued conversation

src/test/integration/completion.test.ts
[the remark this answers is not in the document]
radeksimko: Originally I wasn't commenting so much on what the constraint should be, but more asking what the syntax means. Admittedly it's a question I can and should answer with docs search myself instead of lazily asking you 😅 

https://github.com/hashicorp/setup-terraform#inputs
https://www.npmjs.com/package/semver#user-content-tilde-ranges-123-12-1

TL;DR I'm assuming `~1.0` is actually equivalent to `~1`,  `1.x` or `>= 1.0.0 < 2.0.0`. Given that it's equivalent I don't understand the reason for changing `~1.0` to `~1` but I expect it's a harmless change, although the failing tests would suggest otherwise?

To answer your question of what version should we be testing against - I'd vote for either any latest _compatible_ version (i.e. changing to `~1.1` when `moved` block was introduced) or latest major version (i.e. keeping it as is). Compatible ~= meaning that all tests keep passing if we run them against any version which satisfies the constraint. 

The downside of latest major version is that our tests are little more fragile, but if we claim compatibility with any `v1.x` version to our users, then maybe this is "by design".

src/web/test/integration/extension.test.ts
radeksimko: I'm fine with it as-is, as long as we add a comment to say that we're merely testing successful activation, to prevent confusion of future readers. 😉

tsconfig.json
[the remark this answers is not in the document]
jpogran: I've removed the lines and things seem to still compile and work, so resolving this

src/web/test/integration/extension.test.ts
jpogran: Documented in a code comment inside the test

PR #1219, opened 2022-08-30
Handle Invalid WSL Uri

jpogran:
This adds logic to handle UNC WSL Paths that will not work in terraform-ls. It prompts the user to switch to the Remote WSL extension or allows the user to ignore the prompt.

Needs https://github.com/hashicorp/terraform-ls/pull/1057 to fully function


dbanck (approved):
LGTM 👍

PR #1230, opened 2022-09-08
handle exception reporting

jpogran:
This removes the duplicate exception telemetry from ExtensionErrorHandler as the LanguageClient start exception handler already sends a exception telemetry event.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -12,7 +12,7 @@ import { GenerateBugReportCommand } from './commands/generateBugReport';
 import { ModuleCallsDataProvider } from './providers/moduleCalls';
 import { ModuleProvidersDataProvider } from './providers/moduleProviders';
 import { ServerPath } from './utils/serverPath';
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -12,7 +12,7 @@ import { GenerateBugReportCommand } from './commands/generateBugReport';
 import { ModuleCallsDataProvider } from './providers/moduleCalls';
 import { ModuleProvidersDataProvider } from './providers/moduleProviders';
 import { ServerPath } from './utils/serverPath';
-import { config, handleLanguageClientStart } from './utils/vscode';
+import { config, handleLanguageClientStart as handleLanguageClientStartError } from './utils/vscode';
```
dbanck, line 15: Nit: What do you think about renaming the function to `handleLanguageClientStartError` and not using an alias here?

jpogran: I wonder why the rename quick action did that?

PR #1238, opened 2022-09-19
Fix enable terraform-ls after disbaling

jpogran:
This commit ensures all commands that control terraform-ls are registered regardles of whether or not terraform-ls is enabled. This allows the when clauses to work appropriately but still have the commands registered.


dbanck (approved):
LGTM :+1:

PR #1284, opened 2022-12-01
Prepare for 2.25.1 release

radeksimko:


radeksimko:
@dbanck I ran `npm version 2.25.1` which I think does basically what you described. PTAL

dbanck (approved):
You're right, running `npm version` works too. It will even create a commit with the new version for you.

PR #1320, opened 2023-01-10
Fix endless webpack build on windows

jpogran:
Ensure webpack watch uses development mode to prevent webpack from detecting changes inside node_modules and endlessly building.


dbanck (approved):
LGTM 👍

PR #1325, opened 2023-01-17
Terraform Version LanguageStatusItem

jpogran:
This uses a StaticFeature to provide a LanguageStatusItem that displays the version of terraform that terraform-ls knows about.

tf_version_statusitem

### DIFF
--- a/src/features/terraformVersion.ts
+++ b/src/features/terraformVersion.ts
@@ -1,0 +1,82 @@
+import * as vscode from 'vscode';
+import * as terraform from '../terraform';
+import { ClientCapabilities, ServerCapabilities, StaticFeature } from 'vscode-languageclient';
+import { getActiveTextEditor } from '../utils/vscode';
+import { ExperimentalClientCapabilities } from './types';
+import { Utils } from 'vscode-uri';
+import TelemetryReporter from '@vscode/extension-telemetry';
+import { LanguageClient } from 'vscode-languageclient/node';
+
+export class TerraformVersionFeature implements StaticFeature {
+  private disposables: vscode.Disposable[] = [];
+
+  private clientTerraformVersionCommandId = 'client.refreshTerraformVersion';
+
+  private terraformStatus = vscode.languages.createLanguageStatusItem('terraform.status', [
+    { language: 'terraform' },
+    { language: 'terraform-vars' },
+  ]);
+
+  constructor(
+    private client: LanguageClient,
+    private reporter: TelemetryReporter,
+    private outputChannel: vscode.OutputChannel,
+  ) {
+    this.terraformStatus.name = 'Terraform';
+    this.terraformStatus.detail = 'Terraform Version';
+    this.disposables.push(this.terraformStatus);
+  }
+
+  public fillClientCapabilities(capabilities: ClientCapabilities & ExperimentalClientCapabilities): void {
+    capabilities.experimental = capabilities.experimental || {};
+    capabilities.experimental.refreshTerraformVersionCommandId = this.clientTerraformVersionCommandId;
+  }
+
+  public async initialize(capabilities: ServerCapabilities): Promise<void> {
+    if (!capabilities.experimental?.refreshTerraformVersion) {
+      this.outputChannel.appendLine("Server doesn't support client.refreshTerraformVersion");
+      return;
+    }
+
+    await this.client.onReady();
+
+    const handler = this.client.onRequest(this.clientTerraformVersionCommandId, async () => {
+      const editor = getActiveTextEditor();
+      if (editor === undefined) {
+        return;
+      }
+
+      const moduleDir = Utils.dirname(editor.document.uri);
+
+      try {
+        const response = await terraform.terraformVersion(moduleDir.toString(), this.client, this.reporter);
+        this.terraformStatus.text = response.discovered_version;
+      } catch (error) {
+        let message = 'Unknown Error';
+        if (error instanceof Error) {
+          message = error.message;
+        } else if (typeof error === 'string') {
+          message = error;
+        }
+
+        /*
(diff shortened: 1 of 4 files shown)


src/terraform.ts
```diff
@@ -42,8 +42,26 @@ interface ModuleProvidersResponse {
     [provider: string]: string;
   };
 }
+
+export interface TerraformInfoResponse {
+  v: number;
+  required_version: string;
+  discovered_version: string;
```
dbanck, line 49: `discovered_version` might be optional. If a user has no Terraform installation, the response looks like this:

```json
{"v": 0, "required_version": "~> 1.2"}
```

src/features/terraformVersion.ts
```diff
@@ -0,0 +1,82 @@
...
+      const editor = getActiveTextEditor();
+      if (editor === undefined) {
+        return;
+      }
+
+      const moduleDir = Utils.dirname(editor.document.uri);
+
+      try {
+        const response = await terraform.terraformVersion(moduleDir.toString(), this.client, this.reporter);
+        this.terraformStatus.text = response.discovered_version;
```
dbanck, line 53: What do we want to display if no version was discovered?

dbanck (commented):
Nice work! Also looks great for dev versions.
CleanShot 2023-01-19 at 17 41 52@2x

I've two questions about the case when no Terraform version can be determined.

src/terraform.ts
jpogran: Good catch

src/features/terraformVersion.ts
jpogran: Not sure here. If I don't set it, its an empty space. I could replace with `N/A` or `Not found` but it looks odd.

You can't hide a LanguageStatusItem that's been created like you can with a StatusBarItem, so you are left with it there after initialization. The alternative is to not create it unless a version comes back, but that does not account for the scenario when one directory has a terraform binary available to query, and another doesn't.

radeksimko (commented):
I'll leave review of the implementation details to Daniel, I have just two mostly UX related questions:

1. Do we know if there's any convention on the location and content of such a "status item"? Reason I'm asking is because the official Go extension puts the Go version on the left side and it appears as one "item" Screenshot 2023-01-20 at 10 53 45
They have two additional items in the status bar on the right  I'm not sure if all these things are relevant for us, now or later, and I'm neither sure if other extensions follow what Go does too, just trying to get a sense of "what's common" in VS Code and what and where does the user will expect to see.
2. Are we intentionally leaving out the discovered Terraform version constraint? I thought in the initial issue that it would be useful to display, but I can see how limited space we're working with, so I'd be fine with leaving it out - just seeking ack that this was intentional rather than accidental. 😃

src/features/terraformVersion.ts
radeksimko: I think the combination of what you're proposing may work best, i.e. not create the `LanguageStatusItem` at all if the version is unknown and display something like `unknown` if it changes to unknown as part of the refresh.

FWIW unlike the constraint, we don't really re-discover the installed version on LS side at the moment, so the scenario with `unknown` basically can not happen, for now. https://github.com/hashicorp/terraform-ls/issues/69

jpogran:
> 1. Do we know if there's any convention on the location and content of such a "status item"? 

Yes, there are now two "status item" items: StatusBarItem and LanguageStatusItem. StatusBarItems can be placed anywhere in the status bar, and originally was the only way to indicate status. You could create as many as you wanted and put whatever text you wanted in it, but were encouraged to be a good extension author and not get carried away. LanguageStatusItem is the newer item, and it only appears next to the Language identifier and there is only "one" of them in the sense there is only one `{}` shown. You can add many LanguageStatusItems to it, and it will stack them together. A user can choose to pin one or more of these in the stack, or choose not to, which solves the clutter that StatusBarItem allows.

The Go extension uses StatusBarItems and chooses to put it to the left for versions and right for updating Go. The placement is a design decision on their part. Use of the left side is discouraged as it's seen as "core functionality", but there is no restriction on placement.

I chose LanguageStatusItem because of the standardized and centralized location for many different pieces of information, with it's built in support for status indication and error notification. You can look at the TypeScript and JSON extensions for examples.

> 2. Are we intentionally leaving out the discovered Terraform version constraint?

Not intentional. While it appears we don't have much "space", we have the ability to add a new LanguageStatusItem for each piece of information, which can show status (error, busy, etc) individually. This is what StatusBarItem can't do: keep multiple sets of information together in a concise and configurable form.

resolved

I can't simulate it in the gif, but imagine the author opening the file with the resolved and required versions pinned. The required version shows up, but the resolved version shows an error because the required version is incorrect. The `{}` will shake and show a warning flag, indicating there is an error resolving the constraint.

radeksimko (approved):
LGTM, just one minor wording suggestion.

dbanck (approved):
LGTM 👍

src/features/terraformVersion.ts
```diff
@@ -0,0 +1,92 @@
...
+      if (editor === undefined) {
+        return;
+      }
+
+      const moduleDir = Utils.dirname(editor.document.uri);
+
+      try {
+        const response = await terraform.terraformVersion(moduleDir.toString(), this.client, this.reporter);
+        this.installedVersion.text = response.discovered_version || 'N/A';
+        this.requiredVersion.text = response.required_version || 'N/A';
```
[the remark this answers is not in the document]
jpogran: It's context dependent in my experience, and here it was intended as "Not Available". I like the suggested alternatives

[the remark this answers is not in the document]
radeksimko: I see, that makes sense then 👍🏻 feel free to dismiss my suggestion, I didn't know `N/A` can also have that meaning.

PR #1327, opened 2023-01-24
Update vscode-languageclient to 8.0.2

jpogran:
The vscode-languageclient library introduced breaking API changes in the 8.X version that require changes to our extension to use:

- Remove use of onReady() to wait for start of the LS and use start() instead
- All features that implement StaticFeature need to use the updated interface

In addition to the above were interface changes to ErrorHandler. In applying these changes, it introduced duplicate error messages when loading the extension with incorrect settings. In effect, turning 1 or 2 messages into 3 or 4, depending on what was misconfigured. These cannot be customized: https://github.com/microsoft/vscode-languageserver-node/issues/1011

This commit solves this by removing use of the ExtensionErrorHandler class and instead using inline methods for both initializationFailedHandler and errorHandler to detect failed extension errors. It avoids the duplicate methods by not using the ErrorAction message property and instead keeping track of whether or not an InitializeError was detected.


src/features/moduleCalls.ts
```diff
@@ -10,6 +17,18 @@ export class ModuleCallsFeature implements StaticFeature {
 
   constructor(private client: BaseLanguageClient, private view: ModuleCallsDataProvider) {}
 
+  fillInitializeParams?: ((params: InitializeParams) => void) | undefined;
+
+  preInitialize?:
+    | ((capabilities: ServerCapabilities<any>, documentSelector: vscode.DocumentSelector | undefined) => void)
+    | undefined;
+
```
dbanck, line 25:
```suggestion
```

Since these two methods are optional, we don't have to implement them. Only `getState` is required. The same applies to the other static features below as well.

dbanck (changes requested):
Nice work! Just a small cleanup suggestion within the static features.

dbanck (approved):
There are a couple of unused imports in the static features. Do you mind removing those before merging?

Everything else LGTM 👍

dbanck (approved):
👍 thanks for improving the error messages

PR #1347, opened 2023-02-22
Add copywrite GHA

jpogran:


.github/workflows/test.yml
```diff
@@ -17,6 +17,15 @@ on:
       - '**.md'
 
 jobs:
+  copywrite:
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
+      - name: Setup Copywrite
+        uses: hashicorp/setup-copywrite@3ace06ad72e6ec679ea8572457b17dbc3960b8ce # v1.0.0
```
radeksimko, line 25:
```suggestion
        uses: hashicorp/setup-copywrite@v1
```
AFAIK actions under the `hashicorp` namespace do not need to be pinned to a hash since we maintain them (HashiCorp trusts HashiCorp in that sense), so we could leave this at v1

radeksimko (approved):

.github/workflows/test.yml
jpogran: 🤷🏼 I was following the copywrite repo directions

radeksimko (approved):

PR #1378, opened 2023-04-11
deps: Bump @vscode/vsce to 2.18.x

radeksimko:
Currently pending https://github.com/microsoft/vscode-vsce/pull/858 and release.

--- 

This is to address https://github.com/advisories/GHSA-776f-qx25-q3cc

AFAICT we should not need to mention this as "vulnerability" from end-user perspective in the Changelog, since the package is only used at release time and the content (XML) is already treated as trusted (vsix).

Sadly Microsoft stopped publishing changelogs since `2.15.0`.

https://github.com/microsoft/vscode-vsce/compare/v2.16.0...v2.18.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "@typescript-eslint/parser": "^5.9.0",
         "@vscode/test-electron": "^2.0.1",
         "@vscode/test-web": "^0.0.29",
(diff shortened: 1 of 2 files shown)


radeksimko:
Closing until https://github.com/microsoft/vscode-vsce/pull/858 is addressed as there's not much we can do at this point.

PR #1408, opened 2023-04-28
Update vscode-languageclient to 8.1.0

dbanck:
Closes https://github.com/hashicorp/vscode-terraform/issues/1387

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MPL-2.0",
       "dependencies": {
         "@vscode/extension-telemetry": "^0.4.9",
-        "vscode-languageclient": "8.0.2",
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
This in principle LGTM, but we should carefully review the pie chart of VS Code versions folks used in the past ~30/14 days and assess the potential impact.

dbanck:
I think you're referring to bumping the telemetry dependency in https://github.com/hashicorp/vscode-terraform/issues/1386?

radeksimko:
> I think you're referring to bumping the telemetry dependency in https://github.com/hashicorp/vscode-terraform/issues/1386? This PR doesn't require any engine bumps


Indeed, I got the two mixed up, sorry!

PR #1411, opened 2023-05-01
Implement Terraform Cloud Authentication Provider

jpogran:
This implements the `vscode.AuthenticationProvider` interface and creates a TFC specific provider. This handles listing existing sessions, creating and removing sessions.

Storing sessions/tokens will be improved by https://github.com/hashicorp/vscode-terraform/issues/1397. Currently it only securely stores the token. Future improvement can be done to prevent looking up user data every time the session is reloaded.

The TFC instance to target is defaulted to `app.terraform.io` with a configuration setting to override. This was added in https://github.com/hashicorp/vscode-terraform/issues/1398. Future work should discover whether detecting the TFC instance from the configuration available in the current working space is desirable.

One of the primary ways a user will trigger authentication will be through the VS Code Command palette commands, so https://github.com/hashicorp/vscode-terraform/issues/1399 is added here to finish out the login workflow.

This currently only supports one logged in user at a time. In the future, this may be extended to support multiple user tokens.

Closes https://github.com/hashicorp/vscode-terraform/issues/1396, https://github.com/hashicorp/vscode-terraform/issues/1398

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MPL-2.0",
       "dependencies": {
         "@vscode/extension-telemetry": "^0.4.9",
+        "axios": "^1.4.0",
         "vscode-languageclient": "8.1.0",
         "vscode-uri": "^3.0.2",
         "which": "^2.0.2"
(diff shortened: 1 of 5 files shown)


dbanck (approved):
Great work! I pushed minor async/await changes as promised

image

dbanck:
Merging this even though the web tests are broken. The failure is caused by a broken insiders build and nothing we can fix on end.

PR #1412, opened 2023-05-03
Onboard to pre-release extensions

dbanck:
Running the `./build/preview` script will now only update the version (and LS version if supplied):

```diff
-  "version": "2.26.0",
+  "version": "2.26.2023050315",
```

We supply the `--pre-release` to `vsce package` and `vsce publish` as outlined [here](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions).


jpogran (approved):

build/preview
```diff
@@ -7,26 +7,24 @@ ROOT_RELATIVE_DIR=$(dirname "${SCRIPT_RELATIVE_DIR}")
...
-
-# set preview info in package.json
-(cat package.json | jq --arg VER $VERSION --arg LANGVER $LANGUAGE_SERVER_VERSION --arg NAME "$DISPLAY_NAME" --arg DESC "$DESCRIPTION" '
+# Update versions in package.json
+(cat package.json | jq --arg VER $VER --arg LANGVER $LANGUAGE_SERVER_VERSION '
 .version=$VER |
-.langServer.version=$LANGVER |
-.preview=true |
-.name="terraform-preview" |
-.icon="assets/icons/terraform_logo_mark_dark_universal.png" |
```
radeksimko, line 26: Shall we also remove this icon now?

radeksimko (approved):
LGTM, my only nit-picky note which is less related to the diff/PR is that shell scripts like `build/preview` typically have an extension like `*.sh` or `*.bash`.

build/preview
dbanck: Good catch, did some cleanup

PR #1414, opened 2023-05-03
Remove Legacy Settings

jpogran:
These settings have been deprecated since 2.24.X and there has been a automatic migration available since then. A user that still has these settings will see them dimmed in the setting viewer but will not experience any errors.

image


dbanck (approved):
LGTM 👍

PR #1419, opened 2023-05-08
Remove default activation events

jpogran:
Beginning with VS Code 1.74.0, languages and commands contributed by the extension do not require a corresponding onLanguage or onCommand activation event declaration for the extension to be activated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,12 +48,9 @@
     "url": "https://github.com/hashicorp/vscode-terraform.git"
   },
   "activationEvents": [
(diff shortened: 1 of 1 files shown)


jpogran:
> This LGTM, but why is this targeting `f-tfc` instead of `main`?

I forgot we upped the minimum in main and not just in the feature branch

dbanck (changes requested):
LGTM, except for the extra TFC commit. Can you drop it from this branch?

PR #1421, opened 2023-05-09
Store all session data in `vscode.SecretStorage`

radeksimko:
Depends on https://github.com/hashicorp/vscode-terraform/issues/1397

(can be repointed to `f-tfc` once that PR is merged)

--- 

## Implementation Notes

I know most of the existing logic was copied from https://github.com/microsoft/vscode-extension-samples/blob/main/authenticationprovider-sample/src/authProvider.ts and so these notes/questions apply to that implementation too.

I may be wrong but according to my understanding we should not need all that event listening logic.

I don't know where exactly the data in `SecretStorage` is actually stored (disk, OS keychain, ...?) but I could imagine that the data there may somehow get removed without the sessions getting removed, which would then leave us in inconsistent state? Aside from such seemingly rare scenario I do not see the use case for `SecretStorage.onDidChange`.

Similarly I don't see how the authentication sessions could ever change without our `createSession()` or `removeSession()` being called, regardless of which window do the events come from. Yes, multiple windows will maintain multiple instances of our extension, but if all the state is maintained in `SecretStorage`, which itself is shared between the windows, and firing `vscode.AuthenticationProviderAuthenticationSessionsChangeEvent` impacts all windows too, then we should be safe, right?

### DIFF
--- a/src/providers/authenticationProvider.ts
+++ b/src/providers/authenticationProvider.ts
@@ -106,7 +106,7 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
   }
 
   get onDidChangeSessions(): vscode.Event<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent> {
-    // TODO: What does this do?
+    // Expose our internal event emitter to the outer world
     return this._onDidChangeSessions.event;
   }
 
@@ -115,6 +115,7 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
   async getSessions(scopes?: string[] | undefined): Promise<readonly vscode.AuthenticationSession[]> {
     try {
       const session = await this.sessionPromise;
+      this.logger.info('Successfully fetched Terraform Cloud session.');
       return session ? [session] : [];
     } catch (error) {
       if (error instanceof Error) {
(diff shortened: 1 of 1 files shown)


src/providers/authenticationProvider.ts
```diff
@@ -156,74 +171,46 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
-          void this.checkForUpdates();
-        }
-      }),
-      // This fires when the user initiates a "silent" auth flow via the Accounts menu.
-      vscode.authentication.onDidChangeSessions((e) => {
-        if (e.provider.id === TerraformCloudAuthenticationProvider.providerID) {
-          this.logger.info('Authentication provider change');
-          void this.checkForUpdates();
-        }
-      }),
```
jpogran, line 190: We're really sure we don't need this? What is the harm in keeping this?

jpogran (approved):
Still not 100% about removing the global session change handler, but otherwise LGTM

src/providers/authenticationProvider.ts
dbanck: I looked at the different authentication examples and found that only the [`authenticationprovider-sample`](https://github.com/microsoft/vscode-extension-samples/tree/0db7faa03cb9fffb9c7d99f934e3c5591207c064/authenticationprovider-sample) has both change handlers in place. Unfortounaly I wasn't able to find a git history or any other information why both are introduced.
[github-authentication](https://github.com/microsoft/vscode/blob/e7e97f8ea8519885e13dac9df6eae7c747d531b1/extensions/github-authentication/src/github.ts#L84) and [microsoft-authentication](https://github.com/microsoft/vscode/blob/e7e97f8ea8519885e13dac9df6eae7c747d531b1/extensions/microsoft-authentication/src/betterSecretStorage.ts#L35) rely only on `secrets.onDidChange`.

Keeping both handlers will fire all events twice whenever there is a change, which seems unnecessary.

jpogran: 👍🏼

PR #1430, opened 2023-05-12
Update VS Code command to open TFC to generate token

jpogran:
image


src/providers/authenticationProvider.ts
```diff
@@ -104,6 +105,40 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
   async createSession(_scopes: readonly string[]): Promise<vscode.AuthenticationSession> {
     this.ensureInitialized();
 
+    vscode.window.showInformationMessage;
+    const choice = await vscode.window.showQuickPick(
+      [
+        {
+          label: 'Enter a existing Terraform Cloud User Token',
```
radeksimko, line 112:
```suggestion
          label: 'Enter an existing Terraform Cloud User Token',
```

radeksimko (approved):
I feel there is quite a lot of `HashiCorp` and `Terraform Cloud` where I'm not sure it's necessary (at points where it should be clear what token are we asking for), but maybe that's fine - we can revisit as we iron out the rest of the UI, which is likely to face similar challenges.

LGTM :shipit:

src/providers/authenticationProvider.ts
```diff
@@ -104,15 +107,65 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+        // Prompt for the UAT.
+        token = await vscode.window.showInputBox({
+          ignoreFocusOut: true,
+          placeHolder: 'User access token',
+          prompt: 'Enter an HashiCorp Terraform User Access Token (UAT).',
+          password: true,
+        });
+        break;
+      case 'Use a stored Terraform Cloud User Token':
+        token = this.getTerraformCLIToken(token);
```
radeksimko, line 162: Shouldn't we have some error handling here? Currently this seems to fail silently when I have only prod token there, or when there's no file found (i.e. no token found).

src/providers/authenticationProvider.ts
```diff
@@ -104,15 +107,65 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+      throw new Error('Must login to continue');
+    }
 
-    // Note: this example doesn't do any validation of the token beyond making sure it's not empty.
+    // TODO: Change to production URL
+    const terraformCloudURL = 'https://app.staging.terraform.io/app/settings/tokens?source=terraform-login';
+
+    let token: string | undefined = '';
+    switch (choice.label) {
+      case 'Open Terraform Cloud to generate a token':
```
radeksimko, line 142:
```suggestion
      case 'Open to generate a User token':
```
nitpick: might want to extract that into a constant to prevent similar mistakes in the future

src/providers/authenticationProvider.ts
```diff
@@ -104,15 +107,65 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+        await vscode.env.openExternal(vscode.Uri.parse(terraformCloudURL));
+        // Prompt for the UAT.
+        token = await vscode.window.showInputBox({
+          ignoreFocusOut: true,
+          placeHolder: 'User access token',
+          prompt: 'Enter an HashiCorp Terraform User Access Token (UAT).',
+          password: true,
+        });
+        break;
+      case 'Enter a existing Terraform Cloud User Token':
```
radeksimko, line 152:
```suggestion
      case 'Existing User Token':
```
nitpick: might want to extract that into a constant to prevent similar mistakes in the future

radeksimko (changes requested):

src/providers/authenticationProvider.ts
jpogran: Working on it 😁

src/providers/authenticationProvider.ts
```diff
@@ -223,6 +277,37 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+    // ~/.terraform.d/credentials.tfrc.json
+    const homeProfile = process.env.HOME ?? '';
+    const credFilePath = path.join(homeProfile, '.terraform.d', 'credentials.tfrc.json');
+    if ((await this.pathExists(credFilePath)) === false) {
+      // TODO show error message
+      return undefined;
+    }
+
+    // read and marshall json file
+    const data = JSON.parse(fs.readFileSync(credFilePath, 'utf8'));
```
dbanck, line 291: Could this be replaced by `vscode.workspace.fs.readFile`?

src/providers/authenticationProvider.ts
```diff
@@ -104,17 +107,68 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
     this.ensureInitialized();
 
-    // Prompt for the UAT.
-    const token = await vscode.window.showInputBox({
-      ignoreFocusOut: true,
-      placeHolder: 'User access token',
-      prompt: 'Enter an HashiCorp Terraform User Access Token (UAT).',
-      password: true,
-    });
+    vscode.window.showInformationMessage;
```
dbanck, line 110:
```suggestion
```

src/providers/authenticationProvider.ts
```diff
@@ -104,17 +107,68 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+    const choice = await vscode.window.showQuickPick(
+      [
+        {
```
dbanck, line 113: Should we hide this choice for non-native (e.g. web) platforms? Or do we skip supporting TFC on the web for the MVP?

jpogran: We're going to have to skip web for now as TFC API does not accept CORS at the moment, last I checked.

PR #1444, opened 2023-05-24
TFC Login command

jpogran:
This adds a new prompt when the login workflow is initiated that asks the user how they want to authenticate: from the terraform cli config file, generate a new token on the web, or enter it manually.

The user can choose to have the extension read the terraform cli configuration file for the token. This means the user can reuse tokens they already configured to use with TFC without having to copy paste or create new tokens.

This also adds a user prompt to open the TFC website to generate a user token. This reuses the terraform cli login page to prompt the user to generate a new token.

If the user selects to enter an existing token, a new prompt is opened where the user can enter the token.

https://github.com/hashicorp/vscode-terraform/assets/272569/39c5103f-3d3f-440e-a61f-9f2c0439b9d1

### DIFF
--- a/src/providers/authenticationProvider.ts
+++ b/src/providers/authenticationProvider.ts
@@ -242,7 +242,7 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
     }
 
     // TODO: Change to production URL
-    const terraformCloudURL = `https://${TerraformCloudHost}/app/settings/tokens?source=terraform-login`;
+    const terraformCloudURL = `https://${TerraformCloudHost}/app/settings/tokens?source=vscode-terraform`;
     let token: string | undefined;
     switch (choice.label) {
       case 'Open to generate a User token':
--- a/src/providers/authenticationProvider.ts
+++ b/src/providers/authenticationProvider.ts
@@ -242,7 +242,7 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
     }
 
     // TODO: Change to production URL
-    const terraformCloudURL = 'https://app.staging.terraform.io/app/settings/tokens?source=terraform-login';
+    const terraformCloudURL = `https://${TerraformCloudHost}/app/settings/tokens?source=terraform-login`;
     let token: string | undefined;
     switch (choice.label) {
       case 'Open to generate a User token':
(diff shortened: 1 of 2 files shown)


src/providers/authenticationProvider.ts
```diff
@@ -215,6 +212,111 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+      vscode.window.showErrorMessage(
+        'Failed to read configuration file. Please login using the Terraform CLI and try again',
+      );
+      return undefined;
+    }
+
+    // find app.terraform.io token
+    try {
+      const data = JSON.parse(text);
+      const cred = data.credentials['app.staging.terraform.io'];
```
dbanck, line 301:
```suggestion
      // TODO: Change to production URL
      const cred = data.credentials['app.staging.terraform.io'];
```

Or do we want to store the hostname (`app.staging.terraform.io`) in a constant at the top of the file and use it here and in line 243?

dbanck (approved):
Nice work! 👍

src/providers/authenticationProvider.ts
```diff
@@ -215,6 +214,111 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+    }
+
+    // find app.terraform.io token
+    try {
+      const data = JSON.parse(text);
+      const cred = data.credentials[TerraformCloudHost];
+      return cred.token;
+    } catch (error) {
+      vscode.window.showErrorMessage(
+        'No token found for app.staging.terraform.io. Please login using the Terraform CLI and try again',
```
radeksimko, line 307:
```suggestion
        `No token found for ${TerraformCloudHost}. Please login using the Terraform CLI and try again`,
```

src/providers/authenticationProvider.ts
```diff
@@ -215,6 +214,111 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
+        placeHolder: 'Choose a method to enter a Terraform Cloud User Token',
+        title: 'HashiCorp Terraform Cloud Authentication',
+      },
+    );
+    if (choice === undefined) {
+      return undefined;
+    }
+
+    // TODO: Change to production URL
+    const terraformCloudURL = 'https://app.staging.terraform.io/app/settings/tokens?source=terraform-login';
```
radeksimko, line 245:
```suggestion
    const terraformCloudURL = `https://${TerraformCloudHost}/app/settings/tokens?source=terraform-login`;
```

radeksimko: I think we should consider using a different `source` here, like `terraform-vscode` or anything other than the existing `terraform-login`. I know it means the modal doesn't come up, but it would make hits from the extension indistinguishable from hits from the CLI if we kept it as is.

radeksimko (approved):

PR #1451, opened 2023-05-26
Implement basic TreeView provider for TFC runs

radeksimko:
Implements parts of https://github.com/hashicorp/vscode-terraform/issues/1395

## UX

Screenshot 2023-05-26 at 14 48 30

### DIFF
--- a/package.json
+++ b/package.json
@@ -461,6 +461,16 @@
         "command": "terraform.cloud.workspaces.refresh",
         "title": "Refresh",
         "icon": "$(refresh)"
+      },
+      {
+        "command": "terraform.cloud.runs.refresh",
+        "title": "Refresh",
+        "icon": "$(refresh)"
+      },
+      {
+        "command": "terraform.cloud.run.view",
+        "title": "View Run",
+        "icon": "$(globe)"
       }
     ],
     "menus": {
@@ -504,6 +514,10 @@
         {
           "command": "terraform.providers.openDocumentation",
           "when": "false"
+        },
+        {
+          "command": "terraform.cloud.run.view",
+          "when": "false"
         }
       ],
       "view/title": [
@@ -526,6 +540,11 @@
           "command": "terraform.cloud.workspaces.refresh",
           "when": "view == terraform.cloud.workspaces",
           "group": "navigation"
+        },
+        {
+          "command": "terraform.cloud.runs.refresh",
+          "when": "view == terraform.cloud.runs",
+          "group": "navigation"
         }
       ],
       "view/item/context": [
@@ -536,6 +555,11 @@
         {
           "command": "terraform.providers.openDocumentation",
           "when": "view == terraform.providers && viewItem == moduleProviderHasDocs"
+        },
+        {
+          "command": "terraform.cloud.run.view",
+          "when": "view == terraform.cloud.runs",
+          "group": "inline"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -468,7 +468,7 @@
         "icon": "$(refresh)"
       },
       {
-        "command": "terraform.cloud.run.view",
+        "command": "terraform.cloud.run.viewInBrowser",
         "title": "View Run",
         "icon": "$(globe)"
       }
@@ -516,7 +516,7 @@
           "when": "false"
         },
         {
-          "command": "terraform.cloud.run.view",
+          "command": "terraform.cloud.run.viewInBrowser",
           "when": "false"
         }
       ],
@@ -557,7 +557,7 @@
           "when": "view == terraform.providers && viewItem == moduleProviderHasDocs"
         },
         {
-          "command": "terraform.cloud.run.view",
+          "command": "terraform.cloud.run.viewInBrowser",
           "when": "view == terraform.cloud.runs",
           "group": "inline"
         }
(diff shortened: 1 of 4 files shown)


src/providers/tfc/runProvider.ts
```diff
@@ -4,15 +4,45 @@
...
+    const baseUrl = 'https://app.staging.terraform.io/app';
+
+    this.ctx.subscriptions.push(
+      vscode.commands.registerCommand('terraform.cloud.run.view', (run: RunTreeItem) => {
+        const orgName = this.ctx.workspaceState.get('terraform.cloud.organization', '');
+        if (orgName === '') {
+          return;
+        }
+
+        const runURL = `${baseUrl}/${orgName}/workspaces/${run.workspaceName}/runs/${run.id}`;
```
jpogran, line 31: We don't need as much info here as you're collecting. Each run has a link attribute with a link to itself https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run#list-runs-in-a-workspace

radeksimko: I also don't like the idea of building out the URL here, but AFAICT that self link is unfortunately just a link to the API, not the "browser" version.

```json
"links": {
    "self": "/api/v2/runs/run-bWSq4YeYpfrW4mx7"
}
```

package.json
```diff
@@ -461,6 +461,16 @@
...
         "title": "Refresh",
         "icon": "$(refresh)"
+      },
+      {
+        "command": "terraform.cloud.runs.refresh",
+        "title": "Refresh",
+        "icon": "$(refresh)"
+      },
+      {
+        "command": "terraform.cloud.run.view",
```
jpogran, line 471: Command names have to be unique across the entirety of VS Code, so to prevent confusion and collision I would like to name this something more descriptive than `view`. We have the word `view` everywhere so it's hard to determine what this command will do just by reading the name.

How about `terraform.cloud.run.viewInBrowser`?

The title can remain the same as from the user POV they see a globe and view and that implies internet.

src/providers/tfc/runProvider.ts
```diff
@@ -21,10 +49,127 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeI
...
+    this.tooltip = new vscode.MarkdownString(`
+### ID: \`${id}\`
+
+${attributes.message}
+
+ - **Status**: ${attributes.status}
+ - **Trigger reason**: ${attributes['trigger-reason']}
+ - **Created at**: ${attributes['created-at']}
+ - **Source**: ${attributes.source}
+ - **Terraform** ${attributes['terraform-version']}
```
jpogran, line 132:
```suggestion
### \`${workspaceName}\`
#### ID: \`${id}\`
___
${attributes.message}
___
| | |
--|--
| **Status**               | ${attributes.status}|
| **Trigger reason** | ${attributes['trigger-reason']}|
| **Created at**        | ${attributes['created-at']}|
| **Source**              | ${attributes.source}|
| **Terraform**         | ${attributes['terraform-version']}|
```

Love what is there already. This suggestion brings it a little closer to what the TFC page looks like while also spacing things automatically

src/providers/tfc/runProvider.ts
```diff
@@ -21,10 +49,127 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeI
...
+      case 'policy_soft_failed':
+        return new vscode.ThemeIcon('warning');
+      case 'applied':
+        return new vscode.ThemeIcon('pass-filled');
+      case 'discarded':
+      case 'canceled':
+      case 'force_canceled':
+        return new vscode.ThemeIcon('discard');
+      case 'errored':
+        return new vscode.ThemeIcon('error');
```
jpogran, line 171: We can style these with color to indicate success or failure, but that can be a followup

jpogran (changes requested):
Minor suggestions to improve display

src/providers/tfc/runProvider.ts
radeksimko: I'm happy to do that but I am really struggling to understand what the best way to do it with the existing VS Code APIs https://code.visualstudio.com/api/references/theme-color

What meanings/IDs should I choose for each occasion? I don't see any appropriate "success" or "fail" ones in the long list. Or did you imply we just declare our own colours? How would I go about that? Implement custom `vscode.ThemeColor` which accepts the colour in hex format or something like that?

How do we know that will render well in all light and dark themes?

I'm leaning more on leaving it for later.

radeksimko:
I really wanted to print human-readable versions of the status, trigger reason and the source but I'm not even sure whether that's what we want to be displaying and I was struggling to find what each key even means (even through browsing the internal atlas codebase).

Similarly, I wanted to print the dates in "time ago" format using https://www.npmjs.com/package/javascript-time-ago but I kinda struggled with that library, specifically with setting up the language.

We could render some additional details in the hover tooltip, like branch name, SHA and author + avatar but all of that data requires an additional request to https://developer.hashicorp.com/terraform/cloud-docs/api-docs/configuration-versions#show-a-configuration-version-s-commit-information which complicates things quite a bit.

I also omitted pagination, in part because we haven't decided what the best way of implementing that would be.

--- 

I'm open to discuss any of the above further.

jpogran (approved):
LGTM as is! We can iterate more after we land org login and workspace view

PR #1455, opened 2023-05-26
Implement workspace filtering by project

dbanck:
Implements parts of #1452

---

This PR adds filtering of workspaces by project.

### API Error
CleanShot 2023-06-02 at 13 08 48@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -471,6 +471,11 @@
         "command": "terraform.cloud.run.viewInBrowser",
         "title": "View Run",
         "icon": "$(globe)"
+      },
+      {
+        "command": "terraform.cloud.workspaces.filterByProject",
+        "title": "Filter by Project",
+        "icon": "$(filter)"
       }
     ],
     "menus": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -461,6 +461,11 @@
         "command": "terraform.cloud.workspaces.refresh",
         "title": "Refresh",
         "icon": "$(refresh)"
+      },
+      {
+        "command": "terraform.cloud.workspaces.selectProject",
+        "title": "Select Project",
```
jpogran, line 467: I think 'Filter by Project' fits more to the concept here, as the expanded concept is 'Filter Workspaces by Project'. The expanded 'Select Project to Filter Workspaces' doesn't exactly roll off the tongue. :)

If we do agree to change the title, I think command should go along as well.

src/providers/tfc/workspaceProvider.ts
```diff
@@ -4,23 +4,35 @@
...
   refresh(): void {
     this.didChangeTreeData.fire();
   }
 
+  async selectProject(): Promise<void> {
+    // TODO! only run this if user is logged in
+    const organization = this.ctx.workspaceState.get('terraform.cloud.organization', '');
+    const projectQuickPick = new ProjectQuickPick(organization);
+    const project = await projectQuickPick.pick();
+    // TODO! use project for filtering
```
jpogran, line 32: The UX of this is great! It just leaves off actually using the selected project :(

jpogran (approved):

PR #1459, opened 2023-05-31
Refresh runs when refreshing workspaces

jpogran:
This commit refreshes the runs view when refreshing the workspace view by adding a call to the runDataProvider.refresh when performing workspaceProvider.refresh.

If a user has selected a workspace, then refreshes the view, the selected workspace tree item is passed to the runDataProvider so that the runs for that workspace are refreshed.

If no workspace is selected, then the runDataProvider is passed undefined, which will clear it's view. The user can then select a workspace to view runs.


dbanck (approved):
LGTM, but I would like to decouple this in the long run and have other providers attach to the `onDidChangeTreeData` of the parent ones.

jpogran:
> LGTM, but I would like to decouple this in the long run and have other providers attach to the `onDidChangeTreeData` of the parent ones.

Agreed. Working now, perfect later.

PR #1472, opened 2023-06-07
Add API based filtering for organizations

dbanck:
This PR adds API-based filtering for organisations, based on work done for filtering projects in https://github.com/hashicorp/vscode-terraform/pull/1455.

As a result of discussion in Slack, it also adds the ability to open the browser to create organisations and refresh the list. This improves the workflow for users starting with an empty TFC account.

## UX

**Empty account**
CleanShot 2023-06-12 at 14 30 22@2x

**Some existing organizations**
CleanShot 2023-06-12 at 14 36 21@2x


src/utils/uiHelpers.ts
```diff
@@ -0,0 +1,64 @@
...
+  readonly name: string;
+
+  readonly title: string;
+  readonly placeholder: string;
+  readonly ignoreFocusOut?: boolean;
+
+  fetchItems(query?: string): Promise<vscode.QuickPickItem[]>;
+}
+
+export class APIQuickPick {
```
jpogran, line 18: Would have expected this to be inside the terraformCloud folder as while it's a generic VSCode UI thing, its only really going to be used by the  tfc stuff

jpogran (approved):
Other than the file placement question, LGTM!

PR #1475, opened 2023-06-08
Polish run tree view

radeksimko:
Closes https://github.com/hashicorp/vscode-terraform/issues/1458

## UX

Screenshot 2023-06-09 at 18 16 58

### DIFF
--- a/src/terraformCloud/index.ts
+++ b/src/terraformCloud/index.ts
@@ -27,7 +27,7 @@ const userAgentHeader = pluginHeader(
 
 function pluginLogger(): ZodiosPlugin {
   return {
-    response: async (api, config, response) => {
+    response: async (_api, _config, response) => {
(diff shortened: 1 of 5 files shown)


src/terraformCloud/index.ts
```diff
@@ -48,6 +48,7 @@ export const apiClient = new Zodios(baseUrl, [
 ]);
 apiClient.use(jsonHeader);
 apiClient.use(userAgentHeader);
+apiClient.use(pluginLogger());
```
radeksimko, line 51: This was kinda forgotten here, but also I found it useful for debugging, I'm not sure what to do with it. Maybe remove the `use()` part and leave the plugin logic in place?

I suppose we don't want to keep logging every request in prod.

jpogran (approved):
❤️

PR #1482, opened 2023-06-12
bump zodios to `v10.9.2` and update incorrect API schemas

radeksimko:
This is to reflect the patch upstream https://github.com/ecyrbe/zodios/pull/471

I ran the schemas through all the data we have available in our shared org and they all comply.

### DIFF
--- a/src/terraformCloud/run.ts
+++ b/src/terraformCloud/run.ts
@@ -60,7 +60,7 @@ export const RUN_SOURCE: { [id: string]: string } = {
 const runSources = Object.keys(RUN_SOURCE);
 
 export const runAttributes = z.object({
-  'created-at': z.date(),
+  'created-at': z.coerce.date(),
   message: z.string(),
   source: z.enum([runSources[0], ...runSources]),
   status: runStatus,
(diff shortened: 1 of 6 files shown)


src/terraformCloud/run.ts
```diff
@@ -121,37 +132,46 @@ export const CONFIGURATION_SOURCE: { [id: string]: string } = {
...
 
+const configurationVersionRelationships = z.object({
+  'ingress-attributes': relationship,
+});
+
 // include=configuration_version (implied from .ingress_attributes too)
 // See https://developer.hashicorp.com/terraform/cloud-docs/api-docs/configuration-versions#show-a-configuration-version
 const configurationVersionAttributes = z.object({
-  source: z.enum([cfgSources[0], ...cfgSources]),
+  source: z.enum([cfgSources[0], ...cfgSources]).nullish(),
```
dbanck, line 146: `[cfgSources[0], ...cfgSources]` We duplicate the first entry to end up with an array that contains `ado` twice? What's the intention here?

```suggestion
  source: z.enum(cfgSources).nullish(),
```

dbanck (approved):
LGTM! Only one small question

src/terraformCloud/run.ts
radeksimko: That decoupling of the first element was intentional but the inclusion of that same element in the 2nd argument was not. This is purely to work around this oddity

Screenshot 2023-06-13 at 16 04 52

radeksimko: I will fix the exclusion bug.

PR #1483, opened 2023-06-12
Bump `@vscode/extension-telemetry` to 0.8.0

dbanck:
This PR updates the telemetry library to the latest version. Due to some breaking changes, I had to make some minor changes.

Error tracking now requires a custom event name. Please have a look at the strings I came up with.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -82,7 +82,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
     initializationFailedHandler: (error: ResponseError<InitializeError> | Error | any) => {
       initializationError = error;
 
-      reporter.sendTelemetryException(error);
+      reporter.sendTelemetryErrorEvent('initializationError', { error });
 
       let msg = 'Failure to start terraform-ls. Please check your configuration settings and reload this window';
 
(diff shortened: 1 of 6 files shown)


src/extension.ts
```diff
@@ -48,7 +48,7 @@ let crashCount = 0;
 
 export async function activate(context: vscode.ExtensionContext): Promise<void> {
   const manifest = context.extension.packageJSON;
-  reporter = new TelemetryReporter(context.extension.id, manifest.version, manifest.appInsightsKey);
+  reporter = new TelemetryReporter(manifest.appInsightsKey);
```
radeksimko, line 51: Where does it get the extension ID & version from now? Is that automagically read at runtime from `package.json` or something?

radeksimko (commented):
Looks reasonable - just one more question aside from the in-line one:

Have you tested this via custom vsix and compared the events sent have the same shape?

src/extension.ts
dbanck: With [0.7.5](https://github.com/microsoft/vscode-extension-telemetry/releases/tag/v0.7.5) they removed the need to provide these parameters. I assume the new VS Code telemetry API handles this internally.

radeksimko: As long as you confirmed the VS Code telemetry API handles that and the events land in AppInsights in the expected shape (i.e. under the expected ID & version), then 👍🏻 LGTM

ansgarm:
superseded by https://github.com/hashicorp/vscode-terraform/pull/1911

PR #1488, opened 2023-06-13
Design feedback

jpogran:
This adds run status icons and human readable text to the workspace hover and adds [] to the projectName in the description

### DIFF
--- a/src/providers/tfc/helpers.ts
+++ b/src/providers/tfc/helpers.ts
@@ -30,17 +30,17 @@ export function GetRunStatusIcon(status: string): vscode.ThemeIcon {
     case 'confirmed':
     case 'post_plan_completed':
     case 'planned_and_finished':
-      return new vscode.ThemeIcon('pass');
+      return new vscode.ThemeIcon('pass', new vscode.ThemeColor('charts.green'));
     case 'policy_soft_failed':
-      return new vscode.ThemeIcon('warning');
+      return new vscode.ThemeIcon('warning', new vscode.ThemeColor('charts.yellow'));
     case 'applied':
-      return new vscode.ThemeIcon('pass-filled');
+      return new vscode.ThemeIcon('pass-filled', new vscode.ThemeColor('charts.green'));
     case 'discarded':
     case 'canceled':
     case 'force_canceled':
-      return new vscode.ThemeIcon('discard');
+      return new vscode.ThemeIcon('discard', new vscode.ThemeColor('charts.red'));
     case 'errored':
-      return new vscode.ThemeIcon('error');
+      return new vscode.ThemeIcon('error', new vscode.ThemeColor('charts.red'));
   }
 
   return new vscode.ThemeIcon('indent');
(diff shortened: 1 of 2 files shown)


src/providers/tfc/helpers.ts
```diff
@@ -46,6 +46,66 @@ export function GetRunStatusIcon(status: string): vscode.ThemeIcon {
...
+    case 'canceled':
+      return 'Canceled';
+    case 'force_canceled':
+      return 'Force canceled';
+    case 'errored':
+      return 'Errored';
+  }
+
+  return '';
+}
```
radeksimko, line 107: I kind of wish we didn't duplicate this (and used [the same approach as we do for other enums](https://github.com/hashicorp/vscode-terraform/blob/59ec335415d577a6c613864517bac18abb10a7c2/src/terraformCloud/run.ts#L40-L60) already) but I understand the MVP nature of this. 🤷🏻‍♂️

src/providers/tfc/workspaceProvider.ts
```diff
@@ -185,13 +185,18 @@ export class WorkspaceTreeItem extends vscode.TreeItem {
       ? `$(source-control) [${this.attributes['vcs-repo-identifier']}](${this.attributes['vcs-repo']['repository-http-url']})`
       : '';
 
+    const statusIcon = GetRunStatusIcon(lastRun?.status ?? '');
+
+    const statusMsg = GetRunStatusMessage(lastRun?.status ?? '');
```
radeksimko, line 190: Given that there can be a workspace without any (last) runs, what's the expected/intended behaviour there?

jpogran: No runs mean no status to report. TFC website does not show a status for workspaces with no runs.

src/providers/tfc/helpers.ts
jpogran: As well as the MVP nature, if we were to put the ThemeIcon and human readable string into an enum, we'd have to had a ThemeIcon for each item.

src/providers/tfc/workspaceProvider.ts
radeksimko: AFAICT the code here would produce empty strings which would below error out because we attempt to access `id` on empty string `$(${statusIcon.id})` wouldn't it?

TL;DR don't we need an extra conditional here somewhere?

dbanck: `GetRunStatusIcon` returns a `new vscode.ThemeIcon('indent')` as fallback, so I think everything should work

radeksimko: True but `GetRunStatusMessage` will return empty string, so below we'd print out `Run Status: \n`, wouldn't we?

radeksimko: Oh, now I see we print out the fallback icon first, then the message - sorry. That's fine then.

radeksimko (approved):

src/providers/tfc/workspaceProvider.ts
dbanck: That's true! We could hide `Run Status:` when the message is empty

dbanck (approved):

PR #1489, opened 2023-06-13
Standardize TFC URL Location

jpogran:
This creates an exported constant for the Terraform Cloud Web URL and replaces all hardcoded references.

### DIFF
--- a/src/providers/tfc/organizationPicker.ts
+++ b/src/providers/tfc/organizationPicker.ts
@@ -16,7 +16,7 @@ export class CreateOrganizationItem implements vscode.QuickPickItem {
     return 'Open the browser to create a new organization';
   }
   async open() {
-    await vscode.env.openExternal(vscode.Uri.parse(`https://${TerraformCloudWebUrl}/organizations`));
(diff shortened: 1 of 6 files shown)


src/providers/tfc/organizationPicker.ts
```diff
@@ -16,7 +16,7 @@ export class CreateOrganizationItem implements vscode.QuickPickItem {
     return 'Open the browser to create a new organization';
   }
   async open() {
-    await vscode.env.openExternal(vscode.Uri.parse(`https://${TerraformCloudHost}/app/organizations`));
+    await vscode.env.openExternal(vscode.Uri.parse(`https://${TerraformCloudWebUrl}/organizations`));
```
radeksimko, line 19:
```suggestion
    await vscode.env.openExternal(vscode.Uri.parse(`${TerraformCloudWebUrl}/organizations`));
```
Otherwise the `https://` part would be duplicated here, wouldn't it?

radeksimko (changes requested):

radeksimko (approved):

dbanck (approved):

PR #1490, opened 2023-06-13
TFC: Improve error handling

radeksimko:
In addition to the UX changes shown below, all errors except for 401 and 404 are being logged now. I expect there will be more user-triggered errors we may want to exclude as we can do nothing to prevent them, but this is best effort for now.

## UX (before)

Screenshot 2023-06-15 at 16 36 03

Screenshot 2023-06-15 at 16 37 16

Screenshot 2023-06-15 at 16 38 22

### API response validation errors (Zodios)

Screenshot 2023-06-15 at 15 15 06

### 401

Screenshot 2023-06-15 at 15 16 18

### 404

Screenshot 2023-06-15 at 15 18 06

Screenshot 2023-06-15 at 16 33 20

### Other API errors

Screenshot 2023-06-15 at 16 28 33

### Other non-API errors

Screenshot 2023-06-15 at 16 22 40

### DIFF
--- a/src/providers/tfc/runProvider.ts
+++ b/src/providers/tfc/runProvider.ts
@@ -23,13 +23,20 @@ import {
 } from '../../terraformCloud/run';
 import { WorkspaceTreeItem } from './workspaceProvider';
 import { GetRunStatusIcon, RelativeTimeFormat } from './helpers';
+import { ZodiosError, isErrorFromAlias } from '@zodios/core';
+import { apiErrorsToString } from '../../terraformCloud/errors';
+import { handleAuthError, handleZodiosError } from './uiHelpers';
 
 export class RunTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeItem>, vscode.Disposable {
   private readonly didChangeTreeData = new vscode.EventEmitter<void | vscode.TreeItem>();
   public readonly onDidChangeTreeData = this.didChangeTreeData.event;
   private activeWorkspace: WorkspaceTreeItem | undefined;
 
-  constructor(private ctx: vscode.ExtensionContext, private reporter: TelemetryReporter) {
+  constructor(
+    private ctx: vscode.ExtensionContext,
+    private reporter: TelemetryReporter,
+    private outputChannel: vscode.OutputChannel,
+  ) {
     this.ctx.subscriptions.push(
       vscode.commands.registerCommand('terraform.cloud.runs.refresh', () => {
         this.reporter.sendTelemetryEvent('tfc-runs-refresh');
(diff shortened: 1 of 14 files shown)


src/providers/tfc/workspaceFilters.ts
```diff
@@ -61,14 +69,29 @@ export class ProjectsAPIResource implements APIResource {
...
       let message = 'Failed to fetch projects';
-      if (error instanceof Error) {
-        message = error.message;
+
+      if (error instanceof ZodiosError) {
+        handleZodiosError(error, message, this.outputChannel, this.reporter);
+        return picks;
+      }
+
+      if (axios.isAxiosError(error)) {
```
dbanck, line 78: Didn't we use `isErrorFromAlias` here to get better types for `error.response`? That should allow us to get rid of the `?` and the `as ErrorResponse` type cast below.

src/providers/tfc/runProvider.ts
```diff
@@ -137,13 +144,44 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeI
...
+
+      if (axios.isAxiosError(error)) {
+        if (error.response?.status === 401) {
+          handleAuthError();
+          return [];
+        }
+
+        if (error.response?.status === 404) {
+          vscode.window.showWarningMessage(
+            `Workspace ${this.activeWorkspace?.attributes.name} (${workspaceId}) not found, please pick another one`,
```
dbanck, line 162:
```suggestion
            `Workspace ${this.activeWorkspace.attributes.name} (${workspaceId}) not found, please pick another one`,
```

In line 105, we're checking if `this.activeWorkspace` is set, so it should never be undefined here

dbanck (changes requested):
Nice work, but I have a few minor suggestions and questions.

Most of the `catch` blocks look similar. Could a generic helper function handle streamline the error handling even further?

radeksimko:
> Most of the catch blocks look similar. Could a generic helper function handle streamline the error handling even further?

Yes, I did try that first but it ended up looking like premature optimisation. Zodios doesn't offer any simple way of processing/detecting errors universally from all endpoints and then we need slightly different outcomes between the two pickers, authenticator and tree views, so there's not a lot that can actually be fully decoupled.

> Didn't we use `isErrorFromAlias` here to get better types for `error.response`? That should allow us to get rid of the `?` and the as `ErrorResponse` type cast below.

Previously yes but I walked that back for similar reasons.

Having to explicitly name the aliases increases the chance that at some point we'll add a new one into the `try` block and not remember to handle it here. I only ended up using it where it was truly necessary and frankly I was trying to find a way of avoiding it altogether for these reasons, I just didn't find any.

Also I was aiming for readability, i.e. less nesting and early returns.

dbanck (approved):
Great work! 👍

PR #1495, opened 2023-06-16
TFC workspaces improvements

dbanck:
This PR improve the quality of life when working with larger projects with many workspaces.

See commits for individual changes


src/providers/tfc/workspaceProvider.ts
```diff
@@ -117,6 +117,8 @@ export class WorkspaceTreeDataProvider implements vscode.TreeDataProvider<Worksp
           include: ['current_run'],
           // Include query parameter only if project filter is set
           ...(this.projectFilter && { 'filter[project][id]': this.projectFilter }),
+          'page[size]': 50,
+          sort: '-current-run.created-at',
```
jpogran, line 121: Not completely sold on this changing the order on refresh, but I do like the experience.

jpogran (approved):
Works on my machine!

PR #1511, opened 2023-06-21
Deploy Prerelease and Stable

jpogran:
This configures the packaging task to rewrite the GitHub URLS to point to the f-tfc branch instead of the default main branch.


.github/workflows/preview.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           npm_config_arch: ${{ matrix.npm_config_arch }}
           ls_target: ${{ matrix.ls_target }}
       - name: Package VSIX
-        run: npm run package -- --pre-release --target=${{ matrix.vsce_target }}
+        run: npm run package -- --githubBranch f-tfc --pre-release --target=${{ matrix.vsce_target }}
```
radeksimko, line 67: Apologies for early comment, I understand this is a draft. 😄 

I'm assuming you're planning to make this an input and also use it above in `actions/checkout` to actually checkout the branch, right? 👀

jpogran: Yep. WIP

.github/workflows/deploy.yml
```diff
@@ -47,34 +55,47 @@ jobs:
     runs-on: 'ubuntu-latest'
     steps:
       - name: Check out repository
+        if: inputs.deploy_type == 'stable'
```
radeksimko, line 58:
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -47,34 +55,47 @@ jobs:
...
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # https://github.com/actions/checkout/releases/tag/v3.5.3
-      - name: Set preview in package.json
+
+      - name: Check out prerelease
+        if: inputs.deploy_type == 'prerelease'
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # https://github.com/actions/checkout/releases/tag/v3.5.3
+        with:
+          ref: 'f-tfc'
```
radeksimko, line 65:
```suggestion
```

The `workflow_dispatch` type of event trigger actually implies the select box for branches and the `checkout` action already seems to reflect that, so we shouldn't need to do anything differently here.

Screenshot 2023-06-23 at 14 41 37

We just need to make sure we read the chosen branch name where appropriate, e.g. from `$GITHUB_REF_NAME` variable.

.github/workflows/deploy.yml
```diff
@@ -47,34 +55,47 @@ jobs:
...
         run: npm ci
         env:
           npm_config_arch: ${{ matrix.npm_config_arch }}
           ls_target: ${{ matrix.ls_target }}
-      - name: Package VSIX
-        run: npm run package -- --pre-release --target=${{ matrix.vsce_target }}
+
+      - name: Package PreRelease VSIX
+        if: inputs.deploy_type == 'prerelease'
+        run: npm run package -- --githubBranch f-tfc --pre-release --target=${{ matrix.vsce_target }}
```
radeksimko, line 88: Instead of hard-coding `f-tfc` we could use `$GITHUB_REF_NAME` here and that also means we don't need to do anything special for previews and enable branch-based deploys for stable as a nice side effect.

jpogran:
Applied feedback!

.github/workflows/deploy.yml
```diff
@@ -48,33 +56,39 @@ jobs:
...
-      - name: Package VSIX
-        run: npm run package -- --pre-release --target=${{ matrix.vsce_target }}
+
+      - name: Package PreRelease VSIX
+        if: inputs.deploy_type == 'prerelease'
+        run: npm run package -- --githubBranch $GITHUB_REF_NAME --pre-release --target=${{ matrix.vsce_target }}
+
+      - name: Package Stable VSIX
+        if: inputs.deploy_type == 'stable'
+        run: npm run package -- --target=${{ matrix.vsce_target }}
```
radeksimko, line 85: I wasn't implying in my earlier comment that I necessarily _want_ us to be cutting stable releases from other branches, but rather that GHA will AFAIK always display the branch dropdown menu in the dialog with `workflow_dispatch` and so it may result in a slightly confusing UX if we don't account for that.

Maybe a minor UX issue, esp. if there's just 3 of us cutting releases and we're all aware of this - so feel free to ignore - but I wanted to clarify the risk.

radeksimko (approved):

dbanck (commented):
Nice work! 👍 

Do we want to keep the publishing as one job or split it into multiple jobs like before? That way they can run in parallel and don't interfere with each other.

jpogran:
Superseded by https://github.com/hashicorp/vscode-terraform/pull/1532

PR #1532, opened 2023-07-07
Streamline extension publishing

jpogran:
One stop shop for publishing prerelease and stable extensions to the Marketplace and stable extensions to OpenVSX


.github/workflows/deploy.yml
```diff
@@ -48,42 +56,75 @@ jobs:
...
+        if: inputs.deploy_type == 'prerelease'
         id: set-preview
         run: |
           ./build/preview.sh
         env:
           LANGUAGE_SERVER_VERSION: ${{ github.event.inputs.langserver }}
+
       - name: Setup Node
-        uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # https://github.com/actions/setup-node/releases/tag/v3.7.0
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # https://github.com/actions/setup-node/releases/tag/v3.6.0
```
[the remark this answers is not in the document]
jpogran: Unintentional. I was working on this before we merged in https://github.com/hashicorp/vscode-terraform/pull/1530, so my local was old

.github/workflows/deploy.yml
```diff
@@ -48,42 +56,72 @@ jobs:
...
-      - name: Check latest published version
-        shell: bash
-        run: |
-          export EXTENSION_ID=`cat package.json | jq -r '.publisher + "." + .name'`
-          npx -- vsce show --json $EXTENSION_ID | jq '.versions[0]'
 
-  publish:
-    name: Publish All
+  publish-pre:
+    name: Publish Prerelease
```
radeksimko, line 94:
```suggestion
  publish-pre-marketplace:
    name: Publish Prerelease to Marketplace
```

.github/workflows/deploy.yml
```diff
@@ -48,42 +56,72 @@ jobs:
...
-      - name: Package VSIX
-        run: npm run package -- --pre-release --target=${{ matrix.vsce_target }}
+
+      - name: Package PreRelease VSIX
+        if: inputs.deploy_type == 'prerelease'
+        run: npm run package -- --githubBranch $GITHUB_REF_NAME --pre-release --target=${{ matrix.vsce_target }}
+
+      - name: Package Stable VSIX
+        if: inputs.deploy_type == 'stable'
+        run: npm run package -- --target=${{ matrix.vsce_target }}
```
radeksimko, line 85: I still don't entirely understand why this needs to be separated and why we couldn't just use the `githubBranch` flag in either case, but I suppose it's not a blocker.

dbanck: Line 81 also contains a required `--pre-release` flag

radeksimko: Oh, I missed that one! Good catch and apologies for the confusion!

radeksimko: I still think we could be passing `--githubBranch $GITHUB_REF_NAME` for stable releases as well but again - it's not a blocker.

dbanck (approved):

PR #1535, opened 2023-07-12
New Terraform icons

jpogran:


dbanck (commented):
It looks great!

Are the old SVGs still being used elsewhere or can we remove them?

radeksimko (approved):

jpogran:
> Are the old SVGs still being used elsewhere or can we remove them?

The existing terraform_cloud.svg can go, sorry Radek :)

The terraform.svg one is being use in the modulecalls view, Im not sure the new sidebar icon works there with the border, I'd rather keep that where it is.: https://github.com/hashicorp/vscode-terraform/blob/817c5809a59739e1d0679ffa7c8f2f22fc02d795/src/providers/moduleCalls.ts#L69

The terraform_logo_mark_light_universal.png is being used as our main extension icon which powers the extension view and marketplace, and can't be removed.

PR #1536, opened 2023-07-13
Improve run view speed

jpogran:
This removes the configuration_version, ingress attributes, plan and apply includes in the main run query.

We include plan and apply when querying runs for a given workspace. We need the this info to populate buttons to view a plan or apply log. We get the plan and apply id in the initial query through the relationship attribute, but use the included data to determine whether to display the button or not (whether the run was planned or applied at all). This info is not in the relationship attribute unfortunately. However we do have a run status, which is technically the same (includes more than just plan/applied state). If we use the run status to display the button, we don't need the include to get essentially the same data. The button press already queries the apply/plan using the stored ID, so nothing is affected there by the include removal.

We include configuration_version.ingress_attributes and created_by for the hover information. Since we use resolve_item for hover, we can move the querying of configuration_version.ingress_attributes and created_by to when the user hovers over an item.


src/terraformCloud/ingressAttribute.ts
```diff
@@ -0,0 +1,58 @@
...
+// function findIngressAttributes(included: IncludedObject[], cfgVersion: IngressAttributes) {
+//   const includedObject = included.find(
+//     (included: IncludedObject) =>
+//       included.type === 'ingress-attributes' &&
+//       included.id === cfgVersion.relationships['ingress-attributes']?.data?.id,
+//   );
+//   if (includedObject) {
+//     return includedObject.attributes as IngressAttributes;
+//   }
+// }
```
dbanck, line 58:
```suggestion
```

Can we remove this?

src/providers/tfc/runProvider.ts
```diff
@@ -9,25 +9,15 @@ import TelemetryReporter from '@vscode/extension-telemetry';
...
-} from '../../terraformCloud/run';
+import { RUN_SOURCE, RunAttributes, TRIGGER_REASON } from '../../terraformCloud/run';
 import { WorkspaceTreeItem } from './workspaceProvider';
 import { GetRunStatusIcon, GetRunStatusMessage, RelativeTimeFormat } from './helpers';
 import { ZodiosError, isErrorFromAlias } from '@zodios/core';
 import { apiErrorsToString } from '../../terraformCloud/errors';
 import { handleAuthError, handleZodiosError } from './uiHelpers';
 import { PlanAttributes } from '../../terraformCloud/plan';
 import { ApplyAttributes } from '../../terraformCloud/apply';
+import { CONFIGURATION_SOURCE, IngressAttributes } from '../../terraformCloud/configurationVersion';
```
dbanck, line 20:
```suggestion
import { CONFIGURATION_SOURCE } from '../../terraformCloud/configurationVersion';
import { IngressAttributes } from '../../terraformCloud/ingressAttribute';
```

src/terraformCloud/configurationVersion.ts
```diff
@@ -0,0 +1,100 @@
...
+// function findConfigurationVersionAttributes(included: IncludedObject[], run: Run): ConfigurationVersion | undefined {
+//   const includedObject = included.find(
+//     (included: IncludedObject) =>
+//       included.type === 'configuration-versions' &&
+//       included.id === run.relationships['configuration-version']?.data?.id,
+//   );
+//   if (includedObject) {
+//     return includedObject as ConfigurationVersion;
+//   }
+// }
```
dbanck, line 100:
```suggestion
```

Can we remove this?

src/providers/tfc/runProvider.ts
```diff
@@ -259,60 +223,10 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeI
...
-  }
-}
-
 export class RunTreeItem extends vscode.TreeItem {
-  public createdBy?: UserAttributes;
-  public configurationVersion?: ConfigurationVersionAttributes;
-  public ingressAttributes?: IngressAttributes;
+  public createdByUserId?: string;
+  public configurationVersionId?: string;
+  public ingressAttributesId?: string;
```
dbanck, line 229: This seems unused. Do we need it?

dbanck (changes requested):
I have added some questions and cleanup suggestions.

I noticed that some runs now lack a plan or apply log button.
For example, the first two runs of the `prebuilt-providers` workspace:

Before
CleanShot 2023-07-17 at 17 33 09@2x

After
CleanShot 2023-07-17 at 17 33 33@2x

PR #1540, opened 2023-07-18
Language Status indicator

jpogran:
This sets a busy status when finding out the terraform version, which approximates a loading status for the extension because the request for status happens after all the other requests happen "in the background". So, we get a loading bar for the extension in the lower right hand corner.

https://github.com/hashicorp/vscode-terraform/assets/272569/e3472dd4-97d1-43c1-9124-d1bd173ab7c0


src/features/languageStatus.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.reporter.sendTelemetryEvent('startClient');
+    this.outputChannel.appendLine('Started client');
+
+    const initializeResult = this.client.initializeResult;
+    if (initializeResult === undefined) {
+      return;
+    }
+
+    const multiFoldersSupported = initializeResult.capabilities.workspace?.workspaceFolders?.supported;
+    this.outputChannel.appendLine(`Multi-folder support: ${multiFoldersSupported}`);
```
radeksimko, line 44:
```suggestion
```

This is unrelated to your PR, but since you're touching it here ... 

I'm not entirely sure this log message is still useful 🤔 It has been a long time since we introduced multi-folder support in LS and ~99% of users use the bundled LS, hence it's extremely unlikely this would be a differentiating factor when debugging a problem. We can still see this from the LSP request log anyway.

src/status/language.ts
```diff
@@ -0,0 +1,50 @@
...
+  lsStatus.busy = true;
+}
+
+export function setLanguageServerBusy() {
+  lsStatus.busy = true;
+}
+
+export function setLanguageServerStopped() {
+  // this makes the statusItem a different color in the bar
+  // and triggers an alert, so the user 'sees' that the ls is stopped
```
radeksimko, line 37:
```suggestion
  // and triggers an alert, so the user 'sees' that the LS is stopped
```

src/status/installedversion.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
radeksimko, line 1: Nitpick - shouldn't this file be called `installedVersion.ts` (with capital `V`)?

radeksimko (approved):

PR #1548, opened 2023-07-25
Explicitly set diagnosticCollectionName

jpogran:
In general VS Code is supposed to allow multiple extensions to provide diagnositics for a file without collision. There are [reported cases](https://github.com/sourcery-ai/sourcery-vscode/issues/121) where multiple extensions that claim support for the same language overwrite diagnositics. To avoid this, we can set diagnosticCollectionName explicitly and create our own collection.

While we haven't specifically seen this happen for us, it may have happened and we are not the only ones to set it explicitly.

Reference: https://github.com/microsoft/vscode-languageserver-node/blob/f2ff7d55464a1f58f978cb6635bd8865f050553c/client/src/common/client.ts#L1080-L1084

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -75,6 +75,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
         vscode.workspace.createFileSystemWatcher('**/*.tfvars'),
       ],
     },
+    diagnosticCollectionName: 'HashiCorpTerraform',
     outputChannel: outputChannel,
     revealOutputChannelOn: RevealOutputChannelOn.Never,
     initializationOptions: initializationOptions,


jpogran:
Repointed to main from experimental https://github.com/hashicorp/vscode-terraform/pull/1541

src/extension.ts
```diff
@@ -75,6 +75,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         vscode.workspace.createFileSystemWatcher('**/*.tfvars'),
       ],
     },
+    diagnosticCollectionName: 'HashiCorpTerraform',
```
radeksimko, line 78: [It looks like](https://github.com/search?q=diagnosticCollectionName&type=code) there isn't any conventions for the name. Some extensions choose to reflect the fact that diagnostics come from the LS and some don't, some use spaces, some camelCase. So ... **TL;DR LGTM**

On a separate note, I would have hoped that VS Code separates out diagnostics produced by different extensions "out of the box" but I suppose I have to adjust my expectations. 😄

radeksimko (approved):

PR #1550, opened 2023-07-31
Do not prompt for TFC authentication

jpogran:
This removes the badge on the User Account icon prompting to login to TFC on extension activation. They will no longer be prompted to login unless they click on the Terraform Cloud sidebar. A user can already additionally choose to hide the Terraform Clound sidebar icon and not see any mention of TFC.

Previously the extension would check if there was an existing login session, and if none was found would add a badge to the User Account icon indicating the user should login. This provided two opportunities for the user to see how to login, but would prompt users who did not want to use TFC anyway. By disabling this we lose the second way, but gain control of when the login prompt is displayed.


dbanck (approved):
👍 

great branch name 😂

PR #1551, opened 2023-07-31
Standardize text casing in TFC views

jpogran:
- Choose Sentence case for most phrases
- Always properly capitalize HashiCorp, Terraform, Terraform Cloud, and other HashiCorp products
- Prefer using just 'Terraform Cloud' instead of 'HashiCorp Terraform Cloud'
- Do not capitalize TFC concepts like workspace or organization unless its the start of a phrase
- Do not put periods at the end of phrases in buttons, welcome views, messages
- Do not use colons or semi colons before buttons in welcome views

Exceptions:

- VS Code always capitalizes view titles

### DIFF
--- a/src/providers/authenticationProvider.ts
+++ b/src/providers/authenticationProvider.ts
@@ -147,7 +147,7 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
     try {
       const session = await this.sessionPromise;
       if (session) {
-        this.logger.info('Successfully fetched Terraform Cloud session.');
+        this.logger.info('Successfully fetched Terraform Cloud session');
         await vscode.commands.executeCommand('setContext', 'terraform.cloud.signed-in', true);
         return [session];
       } else {
@@ -171,8 +171,8 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
     // Prompt for the UAT.
     const token = await this.promptForToken();
     if (!token) {
-      this.logger.error('User did not provide a UAT');
-      throw new Error('UAT is required');
+      this.logger.error('User did not provide a token');
+      throw new Error('Token is required');
     }
 
     try {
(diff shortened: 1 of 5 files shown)


src/providers/authenticationProvider.ts
```diff
@@ -280,28 +280,28 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
     const terraformCloudURL = `${TerraformCloudWebUrl}/settings/tokens?source=vscode-terraform`;
     let token: string | undefined;
     switch (choice.label) {
-      case 'Open to generate a User token':
+      case 'Generate a user token':
```
radeksimko, line 283: As far as I remember, this choice cannot be accompanied with an icon to indicate it would open a browser (new window), so I think the more verbose version of the label is still helpful here. I'd perhaps even make it more verbose - e.g. `Open browser to generate a token` or `Open ${TerraformCloudWebUrl} to generate a token` (or just a hostname) or something similar.

radeksimko (approved):
LGTM, just two nitpicks.

src/providers/authenticationProvider.ts
```diff
@@ -280,28 +280,28 @@ export class TerraformCloudAuthenticationProvider implements vscode.Authenticati
...
         break;
-      case 'Existing User Token':
+      case 'Existing user token':
         this.reporter.sendTelemetryEvent('tfc-login', { method: 'existing' });
         // Prompt for the UAT.
         token = await vscode.window.showInputBox({
           ignoreFocusOut: true,
           placeHolder: 'User access token',
-          prompt: 'Enter an HashiCorp Terraform User Access Token (UAT).',
+          prompt: 'Enter an Terraform Cloud user access token',
```
xiehan, line 300: Minor: `s/an/a/g`

xiehan (approved):
One small nitpick but other than that, LGTM

src/providers/authenticationProvider.ts
jpogran: The `detail` part right under this line says `Open the Terraform Cloud website to generate a new token`, which I thought was explanation enough to shorten the label alongside the guidelines in the ticket

I could add a globe, but then the others look like something is missing or not rendered:

image

I could add other icons, but then we'd have to discuss which to use:

image

radeksimko: I completely missed the fact there's the other more detailed explanation right below! Apologies.

PR #1554, opened 2023-08-07
Introduce a new setting for enhanced validation

jpogran:
Adds `terraform.validation.enableEnhancedValidation` setting to the initializationOptions sent to terraform-ls.

I chose to use `terraform.validation` as we expect to add future settings for this feature.

This relies on functionality to be added to terraform-ls to parse the new setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -284,7 +284,7 @@
             "default": false,
             "description": "Display reference counts above top level blocks and attributes."
           },
-          "terraform.validation.earlyValidation": {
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
👍🏻 Working as expected, assuming https://github.com/hashicorp/terraform-ls/pull/1436 is merged

radeksimko:
Regarding documentation I expect we can address that in a separate PR, either reusing something from https://github.com/hashicorp/terraform-ls/pull/1433 or linking to it.

PR #1563, opened 2023-09-11
deps: Bump js-releases to 1.7.0

radeksimko:
This also updates a transitive dependency `openpgp` to `5.10.1`, which should in turn address https://github.com/advisories/GHSA-ch3c-v47x-4pgp

While this was flagged as security alert my understanding is that it doesn't impact us (or our users) in any way, because `js-releases` does not use `readCleartextMessage` nor `verifyCleartextMessage` as per https://github.com/hashicorp/js-releases/blob/6b64ff2a17be14820d7cdef5fad8d45645e83a80/src/index.ts#L215-L231

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -764,14 +764,14 @@
       }
     },
     "node_modules/@hashicorp/js-releases": {
-      "version": "1.6.1",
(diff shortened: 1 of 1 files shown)


dbanck (approved):
And it shouldn't affect users, because we use js-releases at build time to bundle the language server and don't ship it with the final extension.

dbanck:
The test failure is strange, though

PR #1564, opened 2023-09-12
Bump Terraform version in CI

dbanck:
This PR bumps the Terraform version used in the tests from `1.1` to `1.5`. This is more inline with our development environments.


radeksimko (approved):
LGTM

PR #1565, opened 2023-09-12
Add sleep to code action test

jpogran:
The test for code actions can take slightlty longer to activate terraform-ls than the method open() accounts for. This adds a sleep to ensure that it's activated before we attempt to ask for supported code actions.

### DIFF
--- a/src/test/integration/codeAction.test.ts
+++ b/src/test/integration/codeAction.test.ts
@@ -29,6 +29,9 @@ suite('code actions', () => {
       new vscode.CodeAction('Format Document', vscode.CodeActionKind.Source.append('formatAll').append('terraform')),
     ];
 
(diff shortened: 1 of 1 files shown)


src/test/integration/codeAction.test.ts
```diff
@@ -29,6 +29,9 @@ suite('code actions', () => {
       new vscode.CodeAction('Format Document', vscode.CodeActionKind.Source.append('formatAll').append('terraform')),
     ];
 
+    // wait till the LS is ready to acccept a code action request
+    await sleep(1000);
```
dbanck, line 33:
```suggestion
    await new Promise(r => setTimeout(r, 1000));
```
What do you think about inlining the sleep? Since it is not used anywhere else

dbanck (approved):

PR #1570, opened 2023-10-03
Readme: Document enhanced validation

radeksimko:
Implies merging of https://github.com/hashicorp/vscode-terraform/pull/1554

--- 

Rendered version at https://github.com/hashicorp/vscode-terraform/blob/974bfb840f35b1ad9b51fd586c00087b6da8ef75/README.md#validation

I'm open to suggestions for changes as this is just a first pass.


README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
 
 Completing the snippet allows you to tab complete through each attribute and block.
 
-### Syntax Validation
+### Validation
```
jpogran, line 56: FYI changing headers like this breaks the generated URL link, so if we (or anyone else) linked to this markdown block it won't work. Not the end of the world, as it'll load the page and just stop at the top, but thought I should mention it.

README.md
```diff
@@ -21,7 +21,7 @@ Read the [Troubleshooting Guide](#troubleshooting) for answers to common questio
 ## Features
 
 - [IntelliSense](#intellisense-and-autocomplete) Edit your code with auto-completion of providers, resource names, data sources, attributes and more
-- [Syntax validation](#syntax-validation) Diagnostics using `terraform validate` provide inline error checking
+- [Validation](#validation) provides inline diagnostics for invalid configuration as you type
```
jpogran, line 24:
```suggestion
- [Validation](#validation) Provides inline diagnostics for invalid configuration as you type
```

All of the other points are capitalized

README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
 
 Completing the snippet allows you to tab complete through each attribute and block.
 
-### Syntax Validation
+### Validation
+
+While editing, the configuration is validated and invalid code is highlighted.
```
jpogran, line 58:
```suggestion
While editing, the configuration is validated and invalid code is highlighted. 
Terraform configuration files are validated when on opened and on change, and invalid code is marked with diagnostics.
```

Active versus passive tone is preferred in general. I switched higlighted to marked as we technically do not highlight words, but use the built in diagnostics framework of underlines and problems pane.

README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
 
 Completing the snippet allows you to tab complete through each attribute and block.
 
-### Syntax Validation
+### Validation
+
+While editing, the configuration is validated and invalid code is highlighted. 
+
+HCL syntax is checked for e.g. missing `}`, `"` or other "control characters" being out of place.
```
jpogran, line 60:
```suggestion
HCL syntax is checked for e.g. missing control characters like `}`, `"` or others in the wrong place.
```

README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
...
+
+![](docs/validation-rule-hcl.png)
+
+Enhanced validation of selected Terraform language constructs in both `*.tf` and `*.tfvars` files based on detected Terraform version and provider versions is also available from `v2.28.0`. This can highlight deprecations, missing required attributes or blocks, references to undeclared variables and more, [as documented](https://github.com/hashicorp/terraform-ls/blob/main/docs/validation.md#enhanced-validation).
```
jpogran, line 64:
```suggestion
Enhanced validation of selected Terraform language constructs in both `*.tf` and `*.tfvars` files based on detected Terraform version and provider versions is also provided. This can highlight deprecations, missing required attributes or blocks, references to undeclared variables and more [as documented in](https://github.com/hashicorp/terraform-ls/blob/main/docs/validation.md#enhanced-validation).
```

Do we need to state version here? We don't start this section by gating version.

README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
...
+
+![](docs/validation-rule-missing-attribute.png)
+
+![](docs/validation-rule-invalid-ref.png)
+
+The enhanced validation is on by default but can be turned off using the following settings option:
```
jpogran, line 70:
```suggestion
The enhanced validation feature is enabled by default but can be disabled using the following setting:
```

README.md
```diff
@@ -53,9 +53,31 @@ Combining `editor.suggest.preview` with the [pre-fill required fields](#code-com
...
+
+```json
+"terraform.validation.enableEnhancedValidation": false
+```
+
+The extension also provides validation through [`terraform validate`](https://www.terraform.io/cli/commands/validate). This can be triggered via command palette. Unlike the other validation methods, this one implies Terraform CLI being installed and prior successful run of `terraform init` (i.e. local installation of all providers and modules). It is the slowest method, but the most thorough - i.e. it will catch the most mistakes.
```
jpogran, line 76:
```suggestion
The extension also provides validation through [`terraform validate`](https://www.terraform.io/cli/commands/validate). This can be triggered via command palette. Unlike the other validation methods, this one requires the Terraform CLI installed and a previous successful run of `terraform init` (i.e. local installation of all providers and modules) to function correctly. It is the slowest method, but the most thorough - i.e. it will catch the most mistakes.
```

jpogran (changes requested):

README.md
radeksimko: Good point - I think I put too much emphasis on the word "syntax" here when it _can_ include pretty much all that we do (not just HCL syntax).

I will change it back.

I am hoping that at some point in the future this long list is split into individual files per feature and we publish the whole thing as a website - at which point I can imagine link stability becoming more important.

README.md
radeksimko: I wouldn't add the `in` near the end though, because the link isn't standalone.

jpogran (approved):

PR #1587, opened 2023-10-11
Add TFC workspace pagination

dbanck:
This PR adds a simple pagination to the workspaces view. We load the next page and append it to the tree view when a user clicks "load more".

Hitting refresh or using the project filter will reset the cache.

## UX
I'd like to improve the UX here, but it's not possible to programatically update the scroll position or deselect an item. Also, the current solution is cumbersome for 500+ workspaces, where a user would have to hit "load more" at least 10 times to get to workspaces further down the list. A filter (e.g. by name) might improve this in the future.

### DIFF
--- a/src/providers/tfc/workspaceProvider.ts
+++ b/src/providers/tfc/workspaceProvider.ts
@@ -35,14 +35,14 @@ export class WorkspaceTreeDataProvider implements vscode.TreeDataProvider<vscode
     this.ctx.subscriptions.push(
       vscode.commands.registerCommand('terraform.cloud.workspaces.refresh', (workspaceItem: WorkspaceTreeItem) => {
         this.reporter.sendTelemetryEvent('tfc-workspaces-refresh');
-        this.cache = [];
+        this.reset();
         this.refresh();
         this.runDataProvider.refresh(workspaceItem);
       }),
       vscode.commands.registerCommand('terraform.cloud.workspaces.resetProjectFilter', () => {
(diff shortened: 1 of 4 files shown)


src/providers/tfc/workspaceProvider.ts
```diff
@@ -86,28 +96,40 @@ export class WorkspaceTreeDataProvider implements vscode.TreeDataProvider<Worksp
...
       return [element];
     }
 
+    return this.buildChildren();
+  }
+
+  private async buildChildren() {
     try {
-      return this.getWorkspaces();
+      this.cache = [...this.cache, ...(await this.getWorkspaces())];
```
jpogran, line 119: If I switch between organizations, I get a `Element with id '<long_id> is already registered`. I think we need to have the refresh call in the organization switcher somehow indicate we need to clear the cache?

jpogran (changes requested):

jpogran:
> I'd like to improve the UX here, but it's not possible to programatically update the scroll position or deselect an item 

I've seen this behavior in other extensions that attempt to implement a load more experience, so this is expected if frustrating.

> Also, the current solution is cumbersome for 500+ workspaces, where a user would have to hit "load more" at least 10 times to get to workspaces further down the list. A filter (e.g. by name) might improve this in the future.

The Gitlens extension has a 'Load All' button on the 'Load More' item, which would address the multiple waits for more. It would not handle paging through the massive list after clicking, but that is a problem endemic to treeviews not just us.

PR #1589, opened 2023-10-11
TFC organization icon

jpogran:
- Add icon and 'TFC' prefix to statusBarItem text
- Add organization name to Workspace view title
- Add button to Workspace view to select organization

### DIFF
--- a/src/features/terraformCloud.ts
+++ b/src/features/terraformCloud.ts
@@ -74,6 +74,8 @@ export class TerraformCloudFeature implements vscode.Disposable {
       showCollapseAll: true,
       treeDataProvider: workspaceDataProvider,
     });
+    const organization = this.context.workspaceState.get('terraform.cloud.organization', '');
+    workspaceView.title = organization !== '' ? `Workspaces - [${organization}]` : 'Workspaces';
 
     this.context.subscriptions.push(runView, runDataProvider, workspaceDataProvider, workspaceView);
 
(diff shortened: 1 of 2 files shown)


src/features/terraformCloud.ts
```diff
@@ -74,6 +74,8 @@ export class TerraformCloudFeature implements vscode.Disposable {
       showCollapseAll: true,
       treeDataProvider: workspaceDataProvider,
     });
+    const organization = this.context.workspaceState.get('terraform.cloud.organization', '');
+    workspaceView.title = organization !== '' ? `Workspaces - [${organization}]` : 'Workspaces';
```
dbanck, line 78:
```suggestion
    workspaceView.title = organization !== '' ? `Workspaces - (${organization})` : 'Workspaces';
```

Further down, we use `( )` for the title

dbanck (commented):
Looks good! Just a small nit

src/features/terraformCloud.ts
jpogran: Yes, we do, but we use `[ ] ` for denoting Projects right underneath this title, so I thought matching that was more important visually.

I was hesitant about using any bracket type thing at all in the statusbar because its not common to do that, so I'd be inclined to remove the `( )` there before changing the title.

dbanck: I don't have any strong opinions here, as long as we keep things consistent.

Right now we use `[ ]` on first load
CleanShot 2023-10-13 at 14 19 47@2x

and `( )` after changing the org
CleanShot 2023-10-13 at 14 19 59@2x

---

VS Code uses `( )`for workspaces, but I think it would also work without any brackets
CleanShot 2023-10-13 at 14 21 09@2x

dbanck (approved):
👍

PR #1590, opened 2023-10-12
TFC: Implement run panel for viewing plan

radeksimko:


package.json
```diff
@@ -652,6 +652,13 @@
           "name": "Runs",
           "contextualTitle": "Terraform Cloud runs"
         }
+      ],
+      "terraform-cloud-panel": [
+        {
+          "id": "terraform.cloud.run.plan",
+          "name": "Plan",
+          "contextualTitle": "Terraform Plan"
```
jpogran, line 660:
```suggestion
```

A `contextualTitle` is shown whenever a view isn't inside a treeview, which this will never be by default, so this view will always use the contextualTitle. This results in `TERRAFORM PLAN: PLAN` as the effective title, which in addition to being redundant is confusing as we're asking TFC.

I suggest removing this, as VS Code by default uses the viewContainer/panel title and view name, which will result in `TERRAFORM CLOUD: PLAN`

src/providers/tfc/runProvider.ts
```diff
@@ -44,6 +43,13 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<TFCRunTreeIt
         this.reporter.sendTelemetryEvent('tfc-runs-viewInBrowser');
         vscode.env.openExternal(run.websiteUri);
       }),
+      vscode.commands.registerCommand('terraform.cloud.run.viewPlan', async (run: PlanTreeItem) => {
+        if (!run.id) {
+          await vscode.window.showErrorMessage(`No plan found for ${run.id}`);
+          return;
+        }
+        this.planDataProvider.refresh(run.id);
```
jpogran, line 51:
```suggestion
        await vscode.commands.executeCommand('terraform.cloud.run.plan.focus');
        this.planDataProvider.refresh(run.id);
```

If the user hasn't already opened this panel, then clicking this button results in no visual change. The plan is queried and data successfully retrieved, but no data is displayed because the panel isn't visible. There isn't a direct method on a TreeView to change visibility, but you can use an `executeCommand` method to set focus to our panel. This works because all views have a set of commands automatically registered for them by VS Code to control things like visiblity.

src/providers/tfc/runProvider.ts
```diff
@@ -44,6 +43,13 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<TFCRunTreeIt
         this.reporter.sendTelemetryEvent('tfc-runs-viewInBrowser');
         vscode.env.openExternal(run.websiteUri);
       }),
+      vscode.commands.registerCommand('terraform.cloud.run.viewPlan', async (run: PlanTreeItem) => {
```
jpogran, line 46: PlanTreeItem already stores the log URL, because by this point the user has expanded the Run to see the plan and/or apply status and we've queried for the plan info. We can either pass the entire PlanTreeItem object to the refresh method for PlanDataProvider, or pass the log url string instead of querying a second time inside PlanDataProvider

src/providers/tfc/runProvider.ts
radeksimko: Thanks, good catch and great suggestion! 😅 

Are these "magic" command ID "suffixes" (like `.focus`) documented anywhere, so I know where to look next time?

src/providers/tfc/runProvider.ts
radeksimko: Great catch, I'll fix that.

src/providers/tfc/runProvider.ts
jpogran: They are technically documented. If you open an editor with the terraform extension activated and run Open Default Keyboard Shortcuts, you get a list of all keyboard shortcuts.

image

At the bottom is a commented out list of all commands that are available in this session to bind. Search for our extension identifier, and you will see the ones we contribute along with the ones generated for our views:

image

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+          }
+          if (logLine.type === 'outputs' && logLine.outputs) {
+            // JSON.parse() won't do the conversion from [JSON] object
+            // to the Map so we have to do it here.
+            Object.entries(logLine.outputs).forEach(function (value: [string, OutputChange]) {
+              if (!planLog.outputs) {
+                planLog.outputs = new Map();
+              }
+              planLog.outputs.set(value[0], value[1]);
+            });
```
radeksimko, line 157: FWIW This is my understanding of the situation - if that's wrong or if there is any more elegant way of dealing with this I'm all ears!

src/features/terraformCloud.ts
```diff
@@ -77,7 +85,7 @@ export class TerraformCloudFeature implements vscode.Disposable {
     const organization = this.context.workspaceState.get('terraform.cloud.organization', '');
     workspaceView.title = organization !== '' ? `Workspaces - (${organization})` : 'Workspaces';
 
-    this.context.subscriptions.push(runView, runDataProvider, workspaceDataProvider, workspaceView);
+    this.context.subscriptions.push(runView, planView, runDataProvider, workspaceDataProvider, workspaceView);
```
dbanck, line 88:
```suggestion
    this.context.subscriptions.push(runView, planDataProvider, planView, runDataProvider, workspaceDataProvider, workspaceView);
```

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+    const planLog = await this.getPlanFromUrl(plan);
+
+    const items: vscode.TreeItem[] = [];
+    if (planLog && planLog.plannedChanges) {
+      items.push(new PlannedChangesItem(planLog.plannedChanges, planLog.changeSummary));
+    }
+    if (planLog && planLog.driftChanges) {
+      items.push(new DriftChangesItem(planLog.driftChanges, planLog.driftSummary));
+    }
+    //  && planLog.outputs.size > 0
```
dbanck, line 84: Do we need to add this check?

src/providers/tfc/planProvider.ts
dbanck: Usually `JSON.parse` will create an object here and you can use it very similar to a map. Why do you prefer a map?

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+              if (!planLog.outputs) {
+                planLog.outputs = new Map();
+              }
+              planLog.outputs.set(value[0], value[1]);
+            });
+
+            continue;
+          }
+          if (logLine.type === 'diagnostic' && logLine.diagnostic) {
+            if (!planLog.diagnostics || !planLog.diagnosticSummary) {
```
dbanck, line 162:
```suggestion
            if (!planLog.diagnostics && !planLog.diagnosticSummary) {
```
Shouldn't this be an and, so we don't accidentally overwrite data?

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+  constructor(public change: Change) {
+    let label = change.resource.addr;
+    if (change.previous_resource) {
+      label = `${change.previous_resource.addr} → ${change.resource.addr}`;
+    }
+
+    super(label, vscode.TreeItemCollapsibleState.None);
+    this.id = change.action + '/' + change.resource.addr;
+    const icon = GetChangeActionIcon(change.action);
+    this.iconPath = icon;
```
dbanck, line 290:
```suggestion
    this.iconPath = GetChangeActionIcon(change.action);;
```

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+    this.description = change.action;
+
+    const tooltip: vscode.MarkdownString = new vscode.MarkdownString();
```
dbanck, line 293:
```suggestion
    const tooltip = new vscode.MarkdownString();
```
The type will be inferred from the return value.

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+    }
+    super(label, vscode.TreeItemCollapsibleState.Expanded);
+  }
+
+  getChildren(): vscode.TreeItem[] {
+    const items: vscode.TreeItem[] = [];
+    this.plannedChanges.forEach((change) => {
+      items.push(new PlannedChangeItem(change));
+    });
+    return items;
```
dbanck, line 276:
```suggestion
    return this.plannedChanges.map((change) => new PlannedChangeItem(change));
```
`map()` is a good way to transform list items.

You could apply this to the other `getChildren` functions, too.

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+  constructor(public change: Change) {
+    let label = change.resource.addr;
+    if (change.previous_resource) {
+      label = `${change.previous_resource.addr} → ${change.resource.addr}`;
+    }
+
+    super(label, vscode.TreeItemCollapsibleState.None);
+    this.id = 'drift/' + change.action + '/' + change.resource.addr;
+    const icon = GetChangeActionIcon(change.action);
+    this.iconPath = icon;
```
dbanck, line 345:
```suggestion
    this.iconPath = GetChangeActionIcon(change.action);
```

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+    const message = GetDriftChangeActionMessage(change.action);
+    this.description = message;
```
dbanck, line 347:
```suggestion
    this.description = GetDriftChangeActionMessage(change.action);
```

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+
+    const tooltip: vscode.MarkdownString = new vscode.MarkdownString();
+    tooltip.appendMarkdown(`\`${change.resource.addr}\` _${message}_`);
+    this.tooltip = tooltip;
```
dbanck, line 351:
```suggestion
    this.tooltip = new vscode.MarkdownString(`\`${change.resource.addr}\` _${this.description}_`);
```
You can provide an initial value to `MarkdownString()`, so we don't have to call append on an empty string.

src/providers/tfc/planProvider.ts
```diff
@@ -0,0 +1,424 @@
...
+}
+
+class DiagnosticItem extends vscode.TreeItem {
+  constructor(diagnostic: Diagnostic) {
+    super(diagnostic.summary, vscode.TreeItemCollapsibleState.None);
+    this.description = diagnostic.severity;
+    const icon = GetDiagnosticSeverityIcon(diagnostic.severity);
+    this.iconPath = icon;
+
+    const tooltip: vscode.MarkdownString = new vscode.MarkdownString();
```
dbanck, line 413:
```suggestion
    const tooltip = new vscode.MarkdownString();
```

dbanck (commented):
Works great!

I added a couple of questions and minor refactoring suggestions.

src/providers/tfc/planProvider.ts
radeksimko: There may be a little bit of a hesitancy due to me coming to this from Go, where it's difficult to loop over objects, for mostly good reasons like meaning of the data type and the way its stored in memory etc. I do keep forgetting though that things which are wrong in one language could be business as usual in another. 😅 

Even if looping over object is fine in JS I still think there is still benefit to the IntelliSense gained through this explicit map-based type and possibly validation of the object structure as it comes from the API, is there not?

src/providers/tfc/planProvider.ts
radeksimko: I'm not too sure about this one since the message is also used in the tooltip below. It would probably work the same way if we just referred to `this.description` there but semantically it seems different in that we don't intend to be repeating description in the tooltip but rather display the human-readable version of the action.

src/providers/tfc/planProvider.ts
dbanck: Don't get me wrong, I think using maps can bring great benefits over using objets in JS.

But I'm not sure I'm following here. `JSON.parse` returns an object and you loop over an object (`Object.entries(logLine.outputs).forEach`). If `logLine.outputs` were a map, as the types (and IntelliSense) suggest, you could use `logLine.outputs.forEach` to iterate over the contents. But this will throw an error at runtime.

By assigning a type to logLine (`const logLine: LogLine`), _we are telling_ TypeScript that this result is of type `LogLine`, but we don't validate that claim. (This is why we use zod for the other TFC API responses, to make sure they really do conform to a schema). But using zod for the log response may be overkill.
The correct type for `outputs` in `LogLine` would be an object:
```js
export interface LogLine {
  ...
  outputs?: {
    [key: string]: {
      action: ChangeAction;
      sensitive: boolean;
    };
  };
}
```

dbanck (approved):
Thank you for taking the time to implement my suggestions.

LGTM 👍

src/providers/tfc/runProvider.ts
```diff
@@ -44,6 +47,15 @@ export class RunTreeDataProvider implements vscode.TreeDataProvider<TFCRunTreeIt
         this.reporter.sendTelemetryEvent('tfc-runs-viewInBrowser');
         vscode.env.openExternal(run.websiteUri);
       }),
+      vscode.commands.registerCommand('terraform.cloud.run.viewPlan', async (plan: PlanTreeItem) => {
+        await vscode.commands.executeCommand('setContext', 'terraform.cloud.run.viewingPlan', true);
```
jpogran, line 51: Shouldn't this be after the check if no plan found?

radeksimko: Yeah, can be - I didn't realise that the condition produces popup error message which is entirely independent of the tree view. Will fix that.

PR #1620, opened 2023-11-06
Update TFC documentation

jpogran:
This updates the README for the recent TFC updates:

- Removes referencs to preview or prerelease
- Records gifs using default VS Code theme, not light
- Update login gif with correct sidebar icons
- Runs view: Shows collapsed Runs, which when expanded show plan/apply details
- Plan/Apply view: Updated to use the collapsible state to expose the plan and apply items, then clicking the button to download logs


radeksimko (approved):
I'm not sure why but I see the GIFs in markdown rendered with some square-based texture. The texture is also visible on the pictures alone but not as distracting - only when zoomed in.

I'm not sure what to do, it may be just GitHub's own rendering but I am curious if you're seeing it as well?

Screenshot 2023-11-07 at 10 48 27

jpogran:
> I'm not sure why but I see the GIFs in markdown rendered with some square-based texture. The texture is also visible on the pictures alone but not as distracting - only when zoomed in.
> 

I'm going to merge as is, since it displays accurately enough in normal zoom and create a followup ticket to re-record with different software.

PR #1625, opened 2023-11-09
Update Node to v18

jpogran:
VS Code ships with Electron 25 As of v1.82, which moves the supported Node version to v18. This doesn't change our functionality, but keeps us in line with VS Code and with minimum supported versions of Node in GitHub Actions.


package.json
```diff
@@ -740,7 +740,7 @@
     "package": "vsce package",
     "test": "npm run compile:test && node ./out/test/runTest.js",
     "test:unit": "jest",
-    "test:unit:web": "npm run compile && vscode-test-web --headless --browserType=chromium --extensionDevelopmentPath=. --extensionTestsPath=./out/web/test/integration/index.js",
+    "test:unit:web": "npm run compile && npx playwright install chromium && vscode-test-web --headless --browser=chromium --extensionDevelopmentPath=. --extensionTestsPath=./out/web/test/integration/index.js",
```
dbanck, line 743:
```suggestion
    "test:unit:web": "npm run compile && vscode-test-web --headless --browser=chromium --extensionDevelopmentPath=. --extensionTestsPath=./out/web/test/integration/index.js",
```

What do you think about adding `@playwright/browser-chromium` as dev dependency instead? This allows for better dependency caching and fixes the `npm run web` command, too.

jpogran: Kinda prefer being explicit, and the install command is idempotent, but can't argue with fire and forget automatic dependency installation

dbanck (approved):
:shipit:

PR #1642, opened 2023-11-27
Document TFC account log out

jpogran:


README.md
```diff
@@ -141,6 +141,12 @@ If a Run has been Planned or Applied, you can view the raw log for each by expan
 
 ![](docs/tfc/plan_apply_view.gif)
 
+To sign out or log out of your Terraform Cloud session, click the Accounts icon next to the Settings icon in the Activity Bar and select "Sign Out":
```
radeksimko, line 144: What is the difference between "sign out" and "log out"? I've been using the two terms as synonymous 😄

radeksimko (approved):

README.md
jpogran: People use it interchangeably so I included both so that if you search for either of them you'll land at the right place.

radeksimko: Ah, yeah, good call - Ctrl+F "SEO" 😅

PR #1643, opened 2023-11-28
Configure command enablement

jpogran:
- Hide Terraform Cloud commands
- Use proper titles for Terraform commands

Before:

image

After:

With TFC disabled:

image

With TFC enabled and not logged in:

image

With TFC enabled and logged in:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -421,32 +421,32 @@
     "commands": [
       {
         "command": "terraform.generateBugReport",
-        "title": "Terraform: Generate Bug Report"
+        "title": "HashiCorp Terraform: Generate Bug Report"
       },
       {
         "command": "terraform.enableLanguageServer",
-        "title": "Terraform: Enable Language Server"
+        "title": "HashiCorp Terraform: Enable Language Server"
       },
       {
         "command": "terraform.disableLanguageServer",
-        "title": "Terraform: Disable Language Server"
+        "title": "HashiCorp Terraform: Disable Language Server"
       },
       {
         "command": "terraform.init",
-        "title": "Terraform: init"
+        "title": "HashiCorp Terraform: init"
       },
       {
         "command": "terraform.initCurrent",
-        "title": "Terraform: init current folder",
+        "title": "HashiCorp Terraform: init current folder",
         "icon": "$(cloud-download)"
       },
       {
         "command": "terraform.validate",
-        "title": "Terraform: validate"
+        "title": "HashiCorp Terraform: validate"
       },
       {
         "command": "terraform.plan",
-        "title": "Terraform: plan"
+        "title": "HashiCorp Terraform: plan"
       },
       {
         "command": "terraform.modules.refreshList",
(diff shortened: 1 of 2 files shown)


jpogran:
Needs https://github.com/hashicorp/vscode-terraform/pull/1644 for tests to be fixed

src/features/terraformCloud.ts
```diff
@@ -120,13 +120,16 @@ export class TerraformCloudFeature implements vscode.Disposable {
...
-    workspaceView.onDidChangeVisibility((event) => {
+    workspaceView.onDidChangeVisibility(async (event) => {
       if (event.visible) {
         // the view is visible so show the status bar
         this.statusBar.show();
+        await vscode.commands.executeCommand('setContext', 'terraform.cloud.views.visible', true);
       } else {
         // hide statusbar because user isn't looking at our views
         this.statusBar.hide();
+        //terraform.cloud.enabled
```
radeksimko, line 131: What does this comment refer to exactly?

package.json
```diff
@@ -534,15 +544,15 @@
...
         },
         {
           "command": "terraform.validate",
-          "when": "config.terraform.languageServer.enable == true"
+          "when": "config.terraform.languageServer.enable == true && editorLangId == terraform"
         },
         {
           "command": "terraform.plan",
-          "when": "config.terraform.languageServer.enable == true"
+          "when": "config.terraform.languageServer.enable == true && editorLangId == terraform"
```
radeksimko, line 555: AFAICT the intention here was to hide the command when other, primarily non-Terraform languages are being edited.

Was it also intentional to hide it when `terraform-vars`, i.e. `*.tfvars` files are being edited? If not then I think the condition needs updating.

package.json
```diff
@@ -472,50 +468,64 @@
         "title": "Open Documentation",
         "icon": "$(book)"
       },
+      {
+        "command": "terraform.cloud.login",
+        "title": "HashiCorp Terraform Cloud: Login",
+        "enablement": "terraform.cloud.signed-in === false"
```
radeksimko, line 474: Wouldn't users who end up with expired token still want to re-login? Or do we log them out automatically in that case, so this is still relevant? Or do we prompt them from the point we detect expired token automatically and we can just rely on that?

Probably not a big deal, just a question that came to mind!

radeksimko (approved):
I have to say I don't love the `HashiCorp` prefix as it makes everything quite verbose in places but that's more a reflection on the whole Marketplace and VS Code situation in general - as in I don't love it in the extension name either. It is certainly not a reflection on your PR here! 😅 

It dampens the impact of those newly introduced commands, so LGTM aside from the minor questions in-line.

package.json
jpogran: When there is an authentication error like a token expiry, we call `handleAuthArror()`, which directly calls the vscode authentication API to create a new session. That logs out the user and pops the modal dialog to log back in. At this point they're not signed in and the other commands disappear

radeksimko: 👍🏻 That's good enough, thanks for the explanation.

PR #1647, opened 2023-12-05
TFC: Implement apply panel for structured output

radeksimko:
Closes https://github.com/hashicorp/vscode-terraform/issues/1637

Follow-up:

 - https://github.com/hashicorp/vscode-terraform/issues/1650
 - https://github.com/hashicorp/vscode-terraform/issues/1651

--- 

## UX Examples

Single panel

Both panels

https://github.com/hashicorp/vscode-terraform/assets/287584/dba715b2-1e95-4f40-b22b-cec4cdc76f57

## Implementation Notes

While it may look simple and not as helpful without all the deeply structured data in the form of attributes etc. I'm afraid this is as much as we can squeeze out of the _structured logs_ without downloading the whole plan file and rendering all the things with the full knowledge of types etc. This would be a whole different story in terms of complexity though.

I found a few bugs which affected the existing plan log, so while implementing apply log correctly I also fixed them there
 - we previously didn't ignore pending plans/applies, which have no log to offer
 - we were also trying to catch a specific axios exception related to `listRuns` when that methods is never actually called in the `try` part, it was just an earlier copy-paste error

### DIFF
--- a/package.json
+++ b/package.json
@@ -508,6 +508,11 @@
         "icon": "$(output)"
       },
       {
+        "command": "terraform.cloud.run.viewApply",
+        "title": "View apply output",
+        "icon": "$(output)"
+      },
+      {
         "command": "terraform.cloud.organization.picker",
         "title": "HashiCorp Terraform Cloud: Pick Organization",
         "icon": "$(account)"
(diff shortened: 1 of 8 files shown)


jpogran (approved):
LGTM as long as you link the followup tickets in the description so we know where to pickup

PR #1653, opened 2023-12-13
Remove `win32-ia32` build target

dbanck:
Based on https://github.com/microsoft/vscode-vsce/pull/908 and https://github.com/microsoft/vscode/pull/195559, it appears that extensions should no longer be published for the now removed `win32-ia32` target.


radeksimko (approved):
Interesting find!

Truly poor to see upstream merging such impactful PRs with _No description provided._

I think we should mention this in our Changelog (that we no longer provide builds for Windows 32bit and maybe link to that upstream PR (in spite of it having little context).

PR #1656, opened 2023-12-20
Rename TokenTypeCapsule to TokenTypeComplex

jpogran:
This renames the existing token type TokenTypeCapsule to TokenTypeComplex to aid readability and reflect existing naming conventions.

This token represents `list` in `type = list(string)` and similar, i.e. the name of the complex type. The term "capsule" has slightly different meaning in cty and HCL and we should not use it in this context.

Relies on https://github.com/hashicorp/terraform-ls/pull/1529

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
         "description": "Traversal step"
       },
       {
-        "id": "hcl-typeCapsule",
+        "id": "hcl-typeComplex",
         "superType": "keyword",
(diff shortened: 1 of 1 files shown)


dbanck (approved):

dbanck:
https://github.com/hashicorp/vscode-terraform/pull/1659 should fix the test failures

PR #1657, opened 2023-12-27
Update MSW to 2.0

kettanaito:
Hi! I know you are using MSW so I decided to update your stack to the latest 2.0 version of the library. That version introduces a bunch of improvements and bug fixes, as well as brings MSW closer to the platform since it now relies on the standard Fetch API primitives. 

Let me know in case I missed anything!

### DIFF
--- a/src/test/integration/mocks/server.ts
+++ b/src/test/integration/mocks/server.ts
@@ -4,7 +4,7 @@
  */
 
 import { ZodiosPathsByMethod, ZodiosResponseByPath } from '@zodios/core/lib/zodios.types';
-import { ResponseResolver, rest, RestContext, RestRequest } from 'msw';
+import { ResponseResolver, http, HttpResponse } from 'msw';
 import { setupServer } from 'msw/node';
 import { TerraformCloudAPIUrl, apiClient } from '../../../terraformCloud';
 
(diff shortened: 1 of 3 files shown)


src/test/integration/mocks/server.ts
```diff
@@ -4,34 +4,31 @@
...
+import { ResponseResolver, http, HttpResponse } from 'msw';
 import { setupServer } from 'msw/node';
 import { TerraformCloudAPIUrl, apiClient } from '../../../terraformCloud';
 
 type Api = typeof apiClient.api;
 
 export function mockGet<Path extends ZodiosPathsByMethod<Api, 'get'>>(
   path: Path,
-  resolver: ResponseResolver<RestRequest, RestContext, Awaited<ZodiosResponseByPath<Api, 'get', Path>>>,
+  resolver: ResponseResolver<{}, never, Awaited<ZodiosResponseByPath<Api, 'get', Path>>>,
```
kettanaito, line 15: The `ResponseResolver` type generic has changed:

- The first argument is now an `Extras` object. In your context, we can leave it empty.
- The second argument is the request body type. Since you are intercepting a GET request here, its body type is always `never`.
- The third argument is the response body type. I left your code that extracts the body type from the API client as-is.

dbanck:
Thank you for raising this PR! ⭐ 

I've opened https://github.com/hashicorp/vscode-terraform/pull/1659 to address the test failures on `main`, which should resolve most of the failed tests.

I'll be talking to the team about [how to handle 1.75.1](https://github.com/hashicorp/vscode-terraform/actions/runs/7337927130/job/20088660305?pr=1657#step:6:34), which is currently our lowest supported engine version and therefore part of our test matrix. 1.75.1 still uses Node 16 under the hood and fails due to lack of fetch support. Only [VS Code 1.82](https://code.visualstudio.com/updates/v1_82#_electron-25-update) and higher will ship with an updated Electron version and use Node 18.

kettanaito:
Got it, @dbanck. Thanks for the update! Let me know if you need any help with these changes once you upgrade the Node version. I confirm this runs great on Node.js 18.

jpogran:
This is rebased and ready to merge, once https://github.com/hashicorp/vscode-terraform/pull/1718 is merged

jpogran (approved):

jpogran:
Thank you for the contribution @kettanaito !

kettanaito:
Yohoo! 🚀 Let me know if you have any questions about MSW, I would be happy to help. 

Also, if you find it great to work with, please consider [Becoming a Sponsor](https://github.com/sponsors/mswjs). That would be amazing! Thanks.

PR #1659, opened 2024-01-02
Fix test failures on main

dbanck:
We removed calling `terraform init` for integration tests in https://github.com/hashicorp/vscode-terraform/pull/1645 and also removed the Terraform installation altogether. It turns out that the formatting code action still requires Terraform. 

This PR adds the installation of Terraform 1.6 to fix the testing issues on Windows and MacOS (the Ubuntu runner ships with Terraform installed).

VS Code automatically updates the workspace settings when the project is opened, so I've included that small change here as well.


radeksimko (approved):
LGTM, thanks for the investigation and fix!

PR #1673, opened 2024-01-19
Allow customization of tfcloud host

Convez:
Issue: https://github.com/hashicorp/vscode-terraform/issues/1671
* Add palette command for TFC logout vscode authentication provider (appears only after successful login):
image
* Add a provider for TFC Apis
* Allow TFC host selection during login. Hosts already in TFC credentials file are displayed. Options for insertion of new hosts added:
image
image

N.B. Logging to a new hostname does not update the TFC credentials file (Maybe to be implemented in the future?)


jpogran (changes requested):
Hey @Convez , thank you so much for this work! I appreciate the time and effort put in here. 

This is a great start! As it stands right now this does too much in one PR for us to merge as-is, so I would like to split up the work here to get some parts merged and we can collaborate on the design of other parts with you.

The stored credential file location handling fix is something I’d definitely like to get merged in, so can you please make that section a separate PR?

While I like the new UI around selecting TFC endpoints, I’d like to collaborate on the implementation. We see as many people generating user tokens via the TFC website as using the stored credentials file (from `terraform login`), so I want to spend some more time working on how that login workflow looks and works for the different use cases.

Changing most of the way the client authenticates is more than I’d like to change in this PR. I would like to work on a way that doesn’t alter the authentication parts while also accommodating the new selection methods.

Thanks again for the work put into this PR submission!

Convez:
Hello @jpogran I agree there may be a lot going on in this PR.

I can create a new PR with the least possible changes to handle the credential file.

However what I would like to keep in the new PR is the fact I moved the API handling to its own class (apiProvider.tf). 
I did that because the alternative would have been to have global non-const variables in the index.tf file that is deleted in this PR.

I'll open anyway a new PR as soon as I can with the last possible changes (without the apiProvider.tf)

Convez:
Closing this PR, since I opened a simplified version of this

PR #1676, opened 2024-01-22
Improve moduleProviders and moduleCalls welcome views

jpogran:
This commit improves the welcome views for moduleProviders and moduleCalls by adding context variables to the views. This allows us to show different messages depending on the state of the open file inside the tree views.

### DIFF
--- a/package.json
+++ b/package.json
@@ -762,7 +762,7 @@
       },
       {
         "view": "terraform.providers",
-        "contents": "There are no installed providers found for the current open file.\n[Learn more about providers](https://www.terraform.io/docs/language/modules/develop/index.html)",
+        "contents": "There are no installed providers found for the current open file.\n[Learn more about providers](https://www.terraform.io/docs/language/providers/index.html)",
         "when": "terraform.providers.noProviders === true"
       },
       {
--- a/package.json
+++ b/package.json
@@ -752,7 +752,7 @@
     "viewsWelcome": [
       {
         "view": "terraform.providers",
-        "contents": "This view requires the language server to be enabled. Please enable it in your Settings:\n[Open Settings]()",
+        "contents": "This view requires the language server to be enabled:\n[Open Settings](command:terraform.openSettingsJson)",
         "when": "config.languageServer.enable === false"
       },
       {
@@ -777,7 +777,7 @@
       },
       {
         "view": "terraform.modules",
-        "contents": "This view requires the language server to be enabled. Please enable it in your Settings:\n[Open Settings]()",
+        "contents": "This view requires the language server to be enabled:\n[Open Settings](command:terraform.openSettingsJson)",
         "when": "config.languageServer.enable === false"
       },
       {
--- a/package.json
+++ b/package.json
@@ -753,27 +753,27 @@
       {
         "view": "terraform.providers",
         "contents": "This view requires the language server to be enabled:\n[Open Settings](command:terraform.openSettingsJson)",
-        "when": "config.languageServer.enable === false"
+        "when": "config.terraform.languageServer.enable === false"
       },
       {
         "view": "terraform.providers",
         "contents": "There are no open Terraform files. Please open a Terraform configuration file to see installed providers.",
-        "when": "config.languageServer.enable && terraform.providers.documentOpened === false"
+        "when": "config.terraform.languageServer.enable && terraform.providers.documentOpened === false"
       },
       {
         "view": "terraform.providers",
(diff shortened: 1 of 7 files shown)


src/providers/moduleProviders.ts
```diff
@@ -38,15 +38,9 @@ export class ModuleProvidersDataProvider implements vscode.TreeDataProvider<Modu
     ctx.subscriptions.push(
       vscode.commands.registerCommand('terraform.providers.refreshList', () => this.refresh()),
       vscode.window.onDidChangeActiveTextEditor(async (event: vscode.TextEditor | undefined) => {
```
dbanck, line 40:
```suggestion
      vscode.window.onDidChangeActiveTextEditor(async () => {
```

src/providers/moduleCalls.ts
```diff
@@ -76,15 +76,9 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
         }
       }),
       vscode.window.onDidChangeActiveTextEditor(async (event: vscode.TextEditor | undefined) => {
```
dbanck, line 78:
```suggestion
      vscode.window.onDidChangeActiveTextEditor(async () => {
```

dbanck (commented):
Great work! This is a nice improvement over the current state. 

I noticed that when there is no LS running (disabled or on the web) that the message defaults to "There are no open Terraform files [...]". Do we want to improve the message for this case as well?

src/providers/moduleCalls.ts
```diff
@@ -109,26 +103,48 @@ export class ModuleCallsDataProvider implements vscode.TreeDataProvider<ModuleCa
...
+        await vscode.commands.executeCommand('setContext', 'terraform.modules.noResponse', true);
         return [];
       }
+    } catch {
+      // error from terraform-ls
+      await vscode.commands.executeCommand('setContext', 'terraform.modules.noResponse', true);
+      return [];
+    }
 
+    try {
```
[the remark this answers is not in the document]
jpogran: I do not have an example. Previous code was a generic catch over the entire operation, and I was moving towards targeted handling.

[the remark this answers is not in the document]
dbanck: I think we don't need this catch then (same for providers)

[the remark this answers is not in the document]
jpogran: Discussed with Daniel and agreed that from the client's POV, being defensive is helpful for cases we cannot predict.

jpogran:
> I noticed that when there is no LS running (disabled or on the web) that the message defaults to "There are no open Terraform files [...]". Do we want to improve the message for this case as well?

I added another clause with a new command that will open the settings window (either UI or JSON) with the relevant setting highlighted and ready to edit when this occurs.

src/commands/terraformls.ts
```diff
@@ -38,6 +38,12 @@ export class TerraformLSCommands implements vscode.Disposable {
 
         await config('terraform').update('languageServer.enable', false, scope);
       }),
+      vscode.commands.registerCommand('terraform.openSettingsJson', async () => {
+        // this opens the default settings window (either UI or json)
+        await vscode.commands.executeCommand('workbench.action.openSettings', {
```
dbanck, line 43: This opens the settings for me, but jumping to the entry doesn't seem to work.

dbanck: Passing the string as second argument pre-fills the settings search, but I'm not sure if that's the solution you're looking for?
```ts
await vscode.commands.executeCommand('workbench.action.openSettings', 'terraform.languageServer.enable');
```

dbanck (commented):
Thank you for going the extra mile and adding a button! I like the UX.

package.json
```diff
@@ -752,11 +752,53 @@
...
+      },
+      {
+        "view": "terraform.providers",
+        "contents": "The active editor cannot provide information about installed providers.\n[Learn more about providers](https://www.terraform.io/docs/language/providers/index.html)",
+        "when": "config.terraform.languageServer.enable && terraform.providers.noResponse === true"
+      },
+      {
+        "view": "terraform.modules",
+        "contents": "This view requires the language server to be enabled:\n[Open Settings](command:terraform.openSettingsJson)",
+        "when": "config.languageServer.enable === false"
```
dbanck, line 781: There are still 5 `terraform.` prefixes missing ;)
```suggestion
        "when": "config.terraform.languageServer.enable === false"
```

jpogran: One thing at a time! :P

dbanck (approved):
Nice!

PR #1722, opened 2024-03-22
Fix webdriver.io tsconfig.json imports

jpogran:
This adds the types to the package.json for the webdriver.io tests that were missed when moving the dependencies from the main package.json. This resolves errors in tsconfig.json when compiling.


dbanck (approved):
> This adds the types to the package.json for the webdriver.io tests that were missed when moving the dependencies from the main package.json.

I'm not sure what you mean here, but I have no objection to adding these packages now.

PR #1732, opened 2024-04-03
Update vscode-languageclient to 9.0.1

jpogran:
This updates the vscode-languageclient package to 9.0.1. This is a major update that includes a number of changes, including a new dependency on vscode-jsonrpc 8.2.0 and vscode-languageserver-protocol 3.17.5

NOTE: This requires bumping the minimum supported VS Code version to 1.86

### DIFF
--- a/src/features/languageStatus.ts
+++ b/src/features/languageStatus.ts
@@ -19,6 +19,9 @@ export class LanguageStatusFeature implements StaticFeature {
     private outputChannel: vscode.OutputChannel,
   ) {}
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  clear(): void {}
+
   getState(): FeatureState {
     return {
       kind: 'static',
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
       fail-fast: false
       matrix:
         vscode:
-          - '1.82.3'
+          - '1.86.2'
           - 'insiders'
           - 'stable'
         os:
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ The Terraform VS Code extension bundles the [Terraform Language Server](https://
 
 The extension does require the following to be installed before use:
 
-- VS Code v1.82 or greater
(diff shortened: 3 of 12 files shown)


ansgarm (approved):

src/features/languageStatus.ts
```diff
@@ -19,6 +27,15 @@ export class LanguageStatusFeature implements StaticFeature {
     private outputChannel: vscode.OutputChannel,
   ) {}
 
+  fillInitializeParams?: ((params: InitializeParams) => void) | undefined;
+
+  preInitialize?:
+    | ((capabilities: ServerCapabilities<any>, documentSelector: DocumentSelector | undefined) => void)
+    | undefined;
```
dbanck, line 34: Do we need those two optional method definitions or can they (and the additional imports) be removed?

dbanck: This applies to our other static features as well

dbanck (commented):
Nice! Do you expect any issues with this major bump? Should we include it in the 2.30.0 release?

src/features/languageStatus.ts
jpogran: I implemented the interface figuring it was no harm since I had to add `clear` anyway. Removed.

jpogran:
> Nice! Do you expect any issues with this major bump? Should we include it in the 2.30.0 release?

The [9.0.1](https://www.npmjs.com/package/vscode-languageclient/v/9.0.1) version was released 6 months ago, so any issues should have been raised by now.

We do have to bump the engine from from [1.82.3](https://github.com/hashicorp/vscode-terraform/blob/main/package.json#L18) to [1.86.X](https://github.com/microsoft/vscode-languageserver-node/blob/54dd4abae40251a18fbe263285e6e97f27457e62/client/package.json#L7-L9). This may require waiting until the next release as we have several things pending right now I don't want to lose the version for.

dbanck (changes requested):
Nice! I think the only thing missing is bumping the types (`@types/vscode`) to 1.86.0 as well

dbanck (approved):
:shipit:

PR #1739, opened 2024-04-11
Add end to end extension to language server testing

jpogran:
## Organizes and adds the following tests:

- Completions
  - Root document completion
  - Completion for a local module sources (prefix ./)
  - Completion for inputs of a local module
  - Completion of the variable name inside the tfvars file
- Go to Definition
  - Go-to-definition on a local module source
  - Go-to-definition on a module input
  - Go-to-definition on a var.something reference
  - Go-to-definition on the variable name inside tfvars
- References
  - Find module references
  - Find references on the variable block
- Symbols
  - Basic symbol return
- Hover
  - core schema provides docs
- Code actions
  - formats the document

## Migrate to vscode-cli

Removes the original test runner infrastructure and replaces it with vscode-cli, which is the new recommend testing approach. This helps simplify executing VS Code API tests and helps ensure terraform-ls is up and initialized before running tests.

## Notes

We now have two test folders: `src/test` and `test/specs`, along with two fixtures folders: `test/fixtures` and `testFixtures`. This does not address combining them or moving them because the wdio tsconfig/npm packaging conflicts with the vscode ones.


.vscode-test.mjs
```diff
@@ -0,0 +1,30 @@
...
+  launchArgs: ['--disable-extensions', '--disable-workspace-trust'],
+  files: 'out/test/**/*.test.js',
+  mocha: {
+    ui: 'tdd',
+    color: true,
+    timeout: 100000,
+  },
+});
+
+// how to do this?
```
dbanck, line 22: I think one way to achieve this, is introducing a setup file and requiring it in the mocha options above.

_src/test/setup.ts_
```ts
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */

import { apiClient, tokenPluginId } from '../terraformCloud';
import { server } from './integration/mocks/server';

export async function mochaGlobalSetup() {
  apiClient.eject(tokenPluginId);

  server.listen();
}

export async function mochaGlobalTeardown() {
  server.close();
}
```

This makes use of mochas global setup fixtures.

```js
  mocha: {
    ui: 'tdd',
    color: true,
    timeout: 100000,
    require: './out/test/setup.js'
  }
```

src/test/helper.ts
```diff
@@ -5,38 +5,124 @@
...
+  });
 }
 
-export const testFolderPath = path.resolve(__dirname, '..', '..', 'testFixture');
+export async function testHover(docUri: vscode.Uri, position: vscode.Position, expectedCompletionList: vscode.Hover[]) {
+  const actualhover = (await vscode.commands.executeCommand(
+    'vscode.executeHoverProvider',
+    docUri,
+    position,
+  )) as vscode.Hover[];
```
dbanck, line 49:
```suggestion
  const actualhover = await vscode.commands.executeCommand<vscode.Hover[]>(
    'vscode.executeHoverProvider',
    docUri,
    position,
  );
```

We can pass the type argument to `executeCommand` here, like you did for the other helpers.

src/test/helper.ts
```diff
@@ -5,38 +5,124 @@
...
+    assert.deepStrictEqual(actualItem.range.start, expectedItem.range.start);
+    assert.deepStrictEqual(actualItem.range.end, expectedItem.range.end);
+  });
+}
+
+export async function testSymbols(docUri: vscode.Uri, symbolNames: string[]) {
+  const symbols = (await vscode.commands.executeCommand(
+    'vscode.executeDocumentSymbolProvider',
+    docUri,
+  )) as vscode.SymbolInformation[];
```
dbanck, line 110:
```suggestion
  const symbols = await vscode.commands.executeCommand<vscode.SymbolInformation[]>(
    'vscode.executeDocumentSymbolProvider',
    docUri,
  );
```

We can pass the type argument to `executeCommand` here, like you did for the other helpers.

src/test/integration/completion.test.ts
```diff
@@ -4,64 +4,145 @@
...
+        new vscode.CompletionItem('"cloudposse/label/null"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/eks/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/iam/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/kms/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/lambda/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/rds/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/s3-bucket/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/security-group/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-aws-modules/vpc/aws"', vscode.CompletionItemKind.Text),
+        new vscode.CompletionItem('"terraform-google-modules/project-factory/google"', vscode.CompletionItemKind.Text),
```
dbanck, line 107: I like that you've included the results from the module registry here, but I'm wondering if this might introduce some flakiness to this test. I think the results and order are based on the popularity of the modules in the registry. If another module becomes more popular, we will probably have to update this test.

To complete only local modules, we could move the completion position right after the `./` prefix (18, 14). This should give us two results.

src/test/integration/hover.test.ts
```diff
@@ -0,0 +1,50 @@
...
+          new vscode.Range(new vscode.Position(14, 12), new vscode.Position(14, 20)),
+        ),
+      ]);
+    });
+
+    test('returns docs for google provider', async () => {
+      await testHover(docUri, new vscode.Position(0, 1), [
+        new vscode.Hover(
+          new vscode.MarkdownString(
+            '`google` hashicorp/google 5.24.0\n\nProvider Name\n\n[`google` on registry.terraform.io](https://registry.terraform.io/providers/hashicorp/google/5.24.0/docs?utm_content=documentHover\u0026utm_medium=Visual+Studio+Code+-+Insiders\u0026utm_source=terraform-ls)',
```
dbanck, line 43: The provider version is likely to change with the next LS release that bundles a newer version. Can we use this with a wildcard for the version instead? Or, if that is too involved, hover over something else provider related?

dbanck (commented):
Really good work ❤️ ! I think these integration tests will help us a lot in the future!

I added some suggestions on how to reduce the potential flakiness and how to bring back the mocha helpers.

src/test/helper.ts
jpogran: Yeah, after the first few it was tedious adjusting the samples and I forgot to circle back. Thanks!

src/test/integration/hover.test.ts
jpogran: Discussed and resolved to add testing hover on providers that are not bundled to some time in the future in order to not hold up the refactor

dbanck (approved):
Awesome! :shipit:

PR #1741, opened 2024-04-12
Rename TFC to HCP Terraform

jpogran:
This renames all user facing information containing the words "Terraform Cloud" to "HCP Terraform".

It does not rename/refactor any code using the old references. The authentication provider ID and sesison key remain the same to preserve existing logins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Read the [Troubleshooting Guide](#troubleshooting) for answers to common questio
 - [Code Navigation](#code-navigation) Navigate through your codebase with Go to Definition and Symbol support
 - [Code Formatting](#code-formatting) Format your code with `terraform fmt` automatically
 - [Code Snippets](#code-snippets) Shortcuts for common snippets like `for_each` and `variable`
-- [Terraform Cloud Integration](#terraform-cloud-integration) View Terraform Cloud Workspaces and Run details inside VS Code
+- [HCP Terraform Integration](#hcp-terraform-integration) View HCP Terraform Workspaces and Run details inside VS Code
 - [Terraform Module Explorer](#terraform-module-and-provider-explorer) View all modules and providers referenced in the currently open document.
 - [Terraform commands](#terraform-commands) Directly execute commands like `terraform init` or `terraform plan` from the VS Code Command Palette.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -115,13 +115,13 @@ The extension provides several snippets to accelerate adding Terraform code to y
 - `vare` - Empty variable
 - `varm` - Map Variable
 
-### Terraform Cloud Integration
```
jpogran, line 118: This will change the markdown links, is that ok?

src/providers/authenticationProvider.ts
```diff
@@ -91,7 +91,9 @@ class TerraformCloudSessionHandler {
   }
 }
 export class TerraformCloudAuthenticationProvider implements vscode.AuthenticationProvider, vscode.Disposable {
-  static providerLabel = 'HashiCorp Terraform Cloud';
+  static providerLabel = 'HashiCorp Cloud Platform Terraform';
+  // These are IDs and session keys that are used to identify the provider and the session in VS Code secret storage
+  // we cannot change these in the rebrand without the user losing the previous session
   static providerID = 'HashiCorpTerraformCloud';
   private sessionKey = 'HashiCorpTerraformCloudSession';
```
jpogran, line 98: As far as I can tell, changing these will result in the user having to login again.

src/terraformCloud/configurationVersion.ts
```diff
@@ -12,6 +12,8 @@ export const CONFIGURATION_SOURCE: { [id: string]: string } = {
   gitlab: 'GitLab',
   github: 'GitHub',
   terraform: 'Terraform',
+  // TODO: check that the map is rebranded
+  // 'terraform+cloud': 'HCP Terraform',
```
jpogran, line 16: TODO: Check when/if HCP updates this key

src/providers/authenticationProvider.ts
ansgarm: Seems fine to me, as this is not user facing :)

src/terraformCloud/configurationVersion.ts
jpogran: Confirmed this will not change with the rename.

dbanck (approved):
:shipit:

PR #1747, opened 2024-04-23
feat: add tests for `terraform init` based completions

ansgarm:
This PR adds multiple test cases for certain completions that are using schemas that Terraform downloaded during `init`.
It also refactors the test setup to allow for multiple directories containing fixtures. See the first two commits for that.

### Todos
- [x] Fix zero tests running on windows (maybe `path.resolve` creates something not expected by the test framework)
- [ ] Fix tests failing in CI


dbanck:
Superseded by https://github.com/hashicorp/vscode-terraform/pull/1749

PR #1757, opened 2024-05-20
Adopt changie for changelog generation

jpogran:
This adds a config file that supports adding new changelog entires using the changie tool. 

It also splits the current CHANGLOG entries into separate MD files that can be processed by changie to generate the final CHANGELOG along with new changes.


ansgarm (approved):
Looks good! I wonder if we can codify how our beta releases are tagged as well and ensure that changie doesn't increment the data based patch number. But that's not blocking this PR 👍

PR #1772, opened 2024-06-07
Update launch config to use temp profile

jpogran:
This change updates the launch configuration to use a temporary profile
when launching the extension host. This is useful for debugging and developing the extension in an isolated environment from user settings.


dbanck (approved):
Nice! I feel like that should be the default. Why do I need all my installed extensions in the extension host?

PR #1773, opened 2024-06-07
Declare support for Terraform Stacks

jpogran:
This begins the process of adding support for tfstack and tfdeploy files by declaring the file extensions and language identifiers in the package.json file. This will allow the extension to recognize these files and provide syntax highlighting and other language features. Future work will add icons and other features to fill out support.

Of particular note is the aliases used for tfstack.hcl and tfdeploy.hcl files. These aliases are used to provide a more user-friendly name for the language in the VS Code status bar and other places where the language name is displayed. Getting these aliases right is not only important from a branding perspective but also from a user experience perspective. They have to be readable and understandable in a small amount of space.

### DIFF
--- a/build/downloader.ts
+++ b/build/downloader.ts
@@ -108,38 +108,44 @@ async function downloadLanguageServer(platform: string, architecture: string, ex
   });
 }
 
+async function downloadFile(url: string, installPath: string) {
+  if (process.env.downloader_log === 'true') {
+    console.log(`Downloading: ${url}`);
+  }
+
+  const buffer = await fileFromUrl(url);
+  fs.writeFileSync(installPath, buffer);
+  if (process.env.downloader_log === 'true') {
+    console.log(`Download completed: ${installPath}`);
+  }
+}
+
 async function downloadSyntax(info: ExtensionInfo) {
   const release = `v${info.syntaxVersion}`;
-
-  const productName = info.name.replace('-preview', '');
-  const fileName = `${productName}.tmGrammar.json`;
 
   const cwd = path.resolve(__dirname);
   const buildDir = path.basename(cwd);
   const repoDir = cwd.replace(buildDir, '');
   const installPath = path.join(repoDir, 'syntaxes');
 
-  const fpath = path.join(installPath, fileName);
-  if (fs.existsSync(fpath)) {
+  if (fs.existsSync(installPath)) {
     if (process.env.downloader_log === 'true') {
-      console.log(`Syntax path exists at ${fpath}. Exiting`);
+      console.log(`Syntax path exists at ${installPath}. Exiting`);
     }
     return;
   }
 
   fs.mkdirSync(installPath);
 
-  const url = `https://github.com/hashicorp/syntax/releases/download/${release}/${fileName}`;
-  if (process.env.downloader_log === 'true') {
-    console.log(`Downloading: ${url}`);
-  }
-  // const content = await got({ url }).text();
-  // fs.writeFileSync(fpath, content);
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -92,6 +112,16 @@
...
+      },
+      {
+        "language": "terraform-stacks",
+        "scopeName": "source.hcl.terraform",
+        "path": "./syntaxes/terraform.tmGrammar.json"
+      },
+      {
+        "language": "terraform-deployment",
+        "scopeName": "source.hcl.terraform",
+        "path": "./syntaxes/terraform.tmGrammar.json"
```
dbanck, line 124:
```suggestion
        "scopeName": "source.hcl",
        "path": "./syntaxes/hcl.tmGrammar.json"
```

build/downloader.ts
```diff
@@ -108,38 +108,44 @@ async function downloadLanguageServer(platform: string, architecture: string, ex
...
-  const fileName = `${productName}.tmGrammar.json`;
-
   const cwd = path.resolve(__dirname);
   const buildDir = path.basename(cwd);
   const repoDir = cwd.replace(buildDir, '');
   const installPath = path.join(repoDir, 'syntaxes');
 
-  const fpath = path.join(installPath, fileName);
-  if (fs.existsSync(fpath)) {
+  if (fs.existsSync(installPath)) {
```
dbanck, line 131: I think if we only check for the existence of the directory, it's easy to run into a situation where one grammar file exists and the other doesn't. `npm run download:artifacts` will then exit silently. Can we improve this behavior?

dbanck (changes requested):

build/downloader.ts
jpogran: It does have logging if you enable the env variable, but its a valid point. This script was never intended to be permanent, and it attempted to prevent downloading each time, but thats not as important as repeatable and reliable.

Changed method to remove directory if present and redownload again.

package.json
```diff
@@ -75,6 +75,26 @@
         ],
         "configuration": "./language-configuration.json"
       },
+      {
+        "id": "terraform-stacks",
```
ansgarm, line 79: Why is the language identifier for *.tfstack.hcl files terraform-stacks (with a plural s) and the one for *.tfdeploy.hcl just terraform-deployment (singular)?

I don't have any strong reason to ask other than that this might be last chance for us to fix an accidental inconsistency 😄

jpogran: Instead of attempting to reconcile the different pluralizations of the product name vs the language name for each of these files for what is essentially an internal facing identifier, I removed the plural and changed deployment to deploy.

The aliases are the public facing text, and can be changed as the product cements how it's referenced

ansgarm: Sounds good! One thing I just asked myself: The files use `*.tfdeploy.hcl` and `*.tfstack.hcl` – why don't we use `tfdeploy` and `tfstack` as language identifiers 🤔 Would be very close to the suffix and a bit shorter than `terraform-deploy` or `terraform-stack`

dbanck: So far we've established `terraform` for `*.tf` files and `terraform-vars` for `*.tfvars` files. These language IDs are used in all editors that have LSP support. I like James' suggestion and think `terraform-deploy` and `terraform-stack` are reasonable choices to continue the established pattern.
It is also easier to decude the product name/context from the `terraform-*` rather than having to know what `tf` stands for.

ansgarm: Makes sense! Thanks for the explanation!

ansgarm (approved):

dbanck (approved):

PR #1777, opened 2024-06-11
Remove preview script from deploy workflow

jpogran:
This commit removes the preview script from the deploy workflow. The preview script is no longer needed as we use changie for incrementing the version number and will be committing the version number with prerelease as well as stable releases.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -54,14 +54,6 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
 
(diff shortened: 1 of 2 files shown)


ansgarm (approved):
We should document somewhere that `build/preview.sh` should be run manually to create a PR that targets the `pre-release` branch.

dbanck (approved):
I noticed that we will update the version twice, first with `npm version`, then with the changie replacement option. The second time will be a noop, so it shouldn't be a big deal.

As you mentioned in Slack, we need to explore if this PR means that we will have `pre-release` as a long-running branch forever? Or we if can still run the pre-release workflow against `main` and it will update the changelog for the release, but then discard the changes since we're not committing anything?

PR #1779, opened 2024-06-18
Enable LSP for Terraform Stacks and Deployments

jpogran:
This change adds document selectors for the Terraform Stacks and Terraform Deployments languages. This will enable the Language Server Protocol (LSP) for these languages.

Needs https://github.com/hashicorp/terraform-ls/pull/1732

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -41,6 +41,8 @@ const brand = `HashiCorp Terraform`;
 const documentSelector: DocumentSelector = [
   { scheme: 'file', language: 'terraform' },
   { scheme: 'file', language: 'terraform-vars' },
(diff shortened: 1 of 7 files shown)


src/extension.ts
```diff
@@ -41,6 +41,8 @@ const brand = `HashiCorp Terraform`;
 const documentSelector: DocumentSelector = [
   { scheme: 'file', language: 'terraform' },
   { scheme: 'file', language: 'terraform-vars' },
+  { scheme: 'file', language: 'terraform-stacks' },
```
dbanck, line 44:
```suggestion
  { scheme: 'file', language: 'terraform-stack' },
```
🙂

PR #1780, opened 2024-06-19
stacks integration tests initial

ansgarm:
This PR adds a `stacks` integration test suite with fixtures based on [this example](https://github.com/hashicorp-guides/lambda-multi-account-stack) and two first tests that check whether the two new languages are registered correctly.
As they already pass, this is a first PR we can already merge on which I will build upon to add more tests (for completions, etc.)


src/test/integration/stacks/workspace/api-gateway/main.tf
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
```
jpogran, line 3: We have an ignore for files inside workspace folders for copywrite, why do we want them here?

ansgarm: They are in here because I copied the files as is from the stacks example for easier diffing in the future

jpogran: roger roger

jpogran (approved):

PR #1807, opened 2024-07-24
Use esbuild

jpogran:
This PR replaces webpack with esbuild for building or bundling the extension in both development or deployment workflows for both the desktop and web versions of this extension. This improves the speed of building and bundling the extension, reduces the size of the output, and simplifies the build process.

This required changing the build scripts, the npm tasks, adding a new esbuild.js file, and updating the package.json file to include the necessary dependencies. The result is a much faster build process that is more reliable and easier to maintain, without changing the core npm script commands.

In addition to esbuild, this updates the versions of eslint, prettier, and eslint rulesets. This required some updates to existing files to conform to the rulesets.

This is a follow-up to #1761 which configured webpack to use esbuild instead of a typescript loader. While that PR did improve the speed of the build, it still used webpack which is just not as fast as esbuild.


ansgarm (approved):
Test drove it on my machine, worked for me from F5 and from VSIX 👍

PR #1812, opened 2024-07-31
Static schema support for Terraform Test and Mock files

dbanck:
* Requires https://github.com/hashicorp/terraform-ls/pull/1782

This PR adds Terraform Test & Mock file support. It adds the language configuration for test and mock files and registers the document selectors, so test files are picked up by the language server.


jpogran (approved):
🚢

PR #1830, opened 2024-08-27
Remove snippets file in favor of terraform-ls

jpogran:
This commit removes the static snippets file in favor of using the Terraform Language Server to provide snippets completions.

The file was a legacy feature from when there was no language server to provide more specific snippets. These snippets used the existing VS Code api to provide a minimal completion experience, but there was no facility to control when or where these snippets appeared. This frequently caused confusion for users as the snippets in the file were not always relevant to the current context and could not be disabled.

The Terraform Language Server provides a more powerful and context-aware completion experience that can suggest completions for all Terraform language constructs. This is combined with the `editor.suggest.preview` and `terraform.experimentalFeatures.prefillRequiredFields` settings to provide a more accurate and helpful completion experience.


ansgarm (approved):
Nice!

radeksimko:
@jpogran I think this closed https://github.com/hashicorp/vscode-terraform/issues/1130 , right?

PR #1847, opened 2024-09-23
Add terraform-test and terraform-mock to status bar items

ansgarm:
Add terraform-test and terraform-mock to status bar items


jpogran (approved):
🚀

PR #1884, opened 2024-11-21
code coverage

jpogran:
- **integration test coverage**
- **VS Code integration coverage**


jpogran:
This really needs #1873 before this can continue. It's working in the same space and will have to accomodate changes from that PR

PR #1885, opened 2024-11-25
Add build files to .vscodeignore

jpogran:
This commit adds build files to the .vscodeignore file to prevent them from being included in the extension.


ansgarm (approved):
Nice catch!

PR #2034, opened 2025-06-02
feat: TF-25532: Recognize tfcomponent.hcl extension

anubhav-goel:
Update the editor extension to recognize this new extension (tfcomponent.hcl) so that syntax highligting as well as other editor features for stacks are available when users use the .tfcomponent.hcl extension.


jpogran-hashi (changes requested):
Before this can merge, we need a release of terraform-ls which then means increasing the version in package.json

anubhav-goel:
> Before this can merge, we need a release of terraform-ls which then means increasing the version in package.json

Yes sure, have added as a pending task in PR description.

jpogran-hashi (approved):

vignesh-jayabalan (approved):

PR #2265, opened 2026-05-13
ci: declare permissions on check-changelog and test

arpitjain099:
Declares an explicit `permissions:` block on the two CI workflows still relying on the default `GITHUB_TOKEN` scope:

- **check-changelog.yml** — `contents: read` + `pull-requests: read`. Uses `dorny/paths-filter` (reads PR file list) and `actions/github-script` to fail the run if no changelog entry was added. No writes.
- **test.yml** — `contents: read`. Read-only lint/format/types/tests on PR + push to main/pre-release.

YAML validated locally.

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -10,6 +10,11 @@ on:
 
 name: Check if changelog entry exists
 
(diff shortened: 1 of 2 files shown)


arpitjain099:
Closing this for now. Thanks for the review consideration; may revisit in the future if needed.

